PR #1868, opened 2014-12-15
Switch to the DiscoverRunner test runner

maurizi:
The test runner we currently use was deprecated in Django 1.6

This PR is mostly renaming files.
With the new test runner, the `DiscoverRunner`, all tests are discovered based on the file name.
All files matching with `test*.py` are considered test files.

I've also changed our UI tests to match `uitest*.py`, and changed our fabric file to use that test pattern for UI tests


maurizi:
I did some investigation, and I think the drop in coverage is because our coverage reporting becamse *more* accurate with the Discover Runner.

For some (unknown) reason, when I run the unit tests with coverage using the old runner, many of our test files do not have coverage reported (`coverage report` shows output like `NoSource: No source for code: '/usr/local/otm/app/opentreemap/treemap/tests/auth.py` for *most* of the files in `treemap/tests/`).

The changes I made now accurately report all of our source code, which means that the percentage of code covered has dropped.

PR #1869, opened 2014-12-15
Run the unit tests on Circle CI

maurizi:
I got significantly faster results from CircleCI than we're currently seeing from Travis CI, which I think is mostly from the caching of dependencies.

Average results for running the full test suite (without UI tests) seem to be around 5-7 minutes.  For comparison, Travis CI builds seem to take around 8-12 minutes.

Adding in UI tests makes the build time go up to about ~~12 minutes.~~ We can run the UI tests in parallel with the normal unit tests, which takes around 9:30

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,25 +23,24 @@ before_install:
 install:
   - pip install -r travis-requirements.txt
   - npm install
-  - npm install -g grunt-cli@0.1.9
   - npm install -g testem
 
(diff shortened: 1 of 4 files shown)


RickMohr:
Consider removing `flake8` and `npm run check` from Circle CI since they're being run on Travis.

RickMohr:
+1

PR #1879, opened 2014-12-31
Internationalize importer app

steventlamb:


opentreemap/importer/errors.py
```diff
@@ -23,51 +26,55 @@
...
-BOOL_ERROR = (44, 'Not formatted as a boolean', True)
-STRING_TOO_LONG = (45, 'Strings must be less than 255 characters', True)
-INVALID_DATE = (46, 'Invalid date (must by YYYY-MM-DD', True)
+FLOAT_ERROR = (40, trans('Not formatted as a number'), True)
+POS_FLOAT_ERROR = (41, trans('Not formatted as a positive number'), True)
+INT_ERROR = (42, trans('Not formatted as an integer'), True)
+POS_INT_ERROR = (43, trans('Not formatted as a positive integer'), True)
+BOOL_ERROR = (44, trans('Not formatted as a boolean'), True)
+STRING_TOO_LONG = (45, trans('Strings must be less than 255 characters'), True)
+INVALID_DATE = (46, trans('Invalid date (must by YYYY-MM-DD)'), True)
```
steventlamb: @RickMohr @maurizi Don't want to sweep this under the rug - I noticed that this was missing a closing `)` and no other message is structured that way, so I rolled this fix in.

maurizi: Noted, and good catch

opentreemap/importer/templates/importer/partials/species_resolve_popover.html
```diff
@@ -1,11 +1,13 @@
...
 {% load l10n %}
 
-{% block title %}Species Field Error{% endblock title %}
+{% block title %}{% trans "Species Field Error" %}{% endblock title %}
 
 {% block content %}
+{# these are already translated on the server side: #}
 <p>{{ field.msg }}</p>
-<p>Did you mean: "{{ field.custom_resolver.suggestion.display_name }}"?</p>
+<p>{% trans "Did you mean:" %} "{{ field.custom_resolver.suggestion.display_name }}"?</p>
```
maurizi: This still is a good candidate for `blocktrans`, since the `?` should translated as well. (So we can include `¿` for spanish).  Can you change this to something like:
```html
{% with display_name=field.custom_resolver.suggestion.display_name %}
<p>{% blocktrans %}Did you mean: "{{ display_name }}"?{% endblocktrans %}</p>
{% endwith %}
```

PR #1886, opened 2015-01-08
Don't store importer errors in the database

RickMohr:
* Create table of error messages by error code
* `append_error` saves only error code, not error message
* Add `errors_by_array_with_messages` for situations where messages are needed
* Fix a few tests
* No data migration needed because error fields are accessed by name, not position


steventlamb:
(pulled down the branch and did some grepping)

With this change, `errors_as_array` is used almost entirely inside models/base.py. There is one exception:
https://github.com/RickMohr/OTM2/blob/bf06f682b7a8071c8fe53c41a719430cd86917ff/opentreemap/importer/views.py#L534

It was probably always a good idea for this mutation logic to be factored into a model method, but it makes even more sense now. Consider factoring this out to a model method like `GenericImportRow.strip_errors_by_code` and making `errors_as_array` a "private" method, ie `_errors_as_array`. The more encapsulation of error processing, the easier it is to reason about.

opentreemap/importer/errors.py
```diff
@@ -3,13 +3,18 @@
...
-# TODO: stop storing error text in the database in order to
-# make lazy translation possible.
-# See https://github.com/OpenTreeMap/OTM2/issues/1880
-# from django.utils.translation import ugettext_lazy as trans
-trans = lambda x: x
+from django.utils.translation import ugettext_lazy as trans
+
+_messages_by_code = {}
+
+def e(code, message, fatal):
```
steventlamb: consider `e` -> `_e`

opentreemap/importer/errors.py
```diff
@@ -3,13 +3,18 @@
...
+    _messages_by_code[code] = message
+    return (code, message, fatal)
+
+
+def message(code):
```
steventlamb: consider renaming this to `get_message`, because `error['msg']` and `errors.message` look deceptively similar despite their semantic difference.

steventlamb:
On initial review, this looks like a nice, clean way to handle this. I'll +1 after I run the branch and look at some db rows.

opentreemap/importer/errors.py
RickMohr: I did try that, and thought it hurt readability. So I'm leaving it even though `_e` is more correct.

RickMohr:
@steventlamb, other suggestions addressed.

steventlamb:
+1 I tested that this is reading and writing error messages correctly. This is as expected for new records. For existing records, I took a tree import row with errors on three fields that was already in my database, it had `msg` attributes for these errors. I corrected one error and inspected the DB and the `msg` attributes were gone. Excellent.

Issue #1678, opened 2014-09-30
Time zone in reported edits is one hour too early


RickMohr:
I edited a plot (using https://www.opentreemap.org/phillytreemap/features/457475/ which appears to be a bogus planting site) and went to the Edits page. I saw my edit, but the time was one hour too early.

RickMohr:
See also OpenTreeMap/otm-addons#999

PR #1899, opened 2015-01-13
Show import upload date rather than PK

### LINKED ISSUE #1891: Importer: Import event counts are not unique to each map
I got up to Tree Import number 15 on one test map. When I switched to a different test map, it listed my first import as tree import number 16 even though I hadn't done any imports for that map. It seems like the import event counts aren't unique to each map. I think they need to be or it's confusing for the user.

jwalgran:
The import event id is an internal primary key shared across all instances. It is confusing for the first import created by the map owner to be labeled 17, and the next labeled 21.

The upload date is not only less confusing, but provides additional, useful information.

Fixes #1891 

screen shot 2015-01-13 at 12 11 52 pm

### DIFF
--- a/opentreemap/importer/templates/importer/partials/import_table.html
+++ b/opentreemap/importer/templates/importer/partials/import_table.html
@@ -2,7 +2,7 @@
 <table class="table table-hover">
     <thead>
         <tr>
-            <th>{% trans "Import Event" %}</th>
+            <th>{% trans "Uploaded Date" %}</th>
             <th>{% trans "Status" %}</th>
             <th>{% trans "Rows" %}</th>
             <th>{% trans "View" %}</th>
@@ -11,7 +11,7 @@
     <tbody>
     {% for ie in things %}
         <tr>
-            <td>{{ ie }}</td>
+            <td>{{ ie.created }}</td>
             <td>{{ ie.status_summary }}</td>
             <td>
                 {% if ie.is_finished %}
@@ -29,4 +29,3 @@
     {% endfor %}
     </tbody>
 </table>
-


opentreemap/importer/templates/importer/partials/import_table.html
```diff
@@ -11,7 +11,7 @@
     <tbody>
     {% for ie in things %}
         <tr>
-            <td>{{ ie }}</td>
+            <td>{{ ie.created }}</td>
```
maurizi: Shouldn't this be formatted using the instance's date format?

Like so: `{{ ie.create|date:request.instance.date_format }}`

.. Though to be more accurate, we should probably add an instance-level `datetime_format` [here](https://github.com/OpenTreeMap/OTM2/blob/master/opentreemap/treemap/instance.py#L234-L238) and use that, since `date_format` does not include time.

maurizi:
+1

RickMohr:
Consider a humanized date (e.g. "3 hours ago") with actual date/time as tooltip. Pretty easy with Django.

RickMohr:
Should we worry about converting to the timezone of the user's browser? (If not, all the more reason to use the humanized date)

jwalgran:
The dates are being localized. In the screenshot, those times correspond to my mountain time clock, but they are in the DB as UTC.

screen shot 2015-01-13 at 1 06 37 pm

I am going to open an issue to add a time format setting to the instance, but I am not going to tackle it now.

jwalgran:
Humanize is ok, but in the case where you upload a file over and over (which will probably be a relatively common occurrence) you get a stack of indistinguishable rows.

screen shot 2015-01-13 at 1 18 04 pm

I would prefer to leave the full time. @dboyer do you have a preference?

dboyer:
I prefer the actual time rather than humanizing it for the reason Justin described.

RickMohr:
According to https://docs.djangoproject.com/en/1.6/ref/contrib/humanize/
`naturaltime` should tell you e.g. "5 minutes ago" or "27 seconds ago"

jwalgran:
``naturaltime`` does work as advertised, but the granularity decreases over time. In the example above 11:25 and 12:05 both show up as "2 hours ago"

RickMohr:
I see, and agree with your conclusion.

dboyer:
I don't know why, but the importer is showing my dates as an hour behind the date showing on my computer (2:30 rather than 3:30).

jwalgran:
@dboyer Do you only see the problem on 1 specific computer? I think we are using some automatic localization of times that may be getting a wrong location from your browser.

dboyer:
I've only noticed it on my home office computer. I'll check on another computer when I'm in the office tomorrow. Either way, I don't think this is something that needs to be fixed before we push to production.

dboyer:
This also happens on my computer at the office. I don't think it's a huge deal but something to keep in mind since we've seen this issue before - #1678

PR #1904, opened 2015-01-13
Add links to importer guids and templates

### LINKED ISSUE #1795: The bulk importer page needs to educate users about how to format their CSVs
We have rather stringent requirements on CSV column header names, but they are not shown or mentioned anywhere in the page.  Perhaps we should add a section on the main "Bulk Uploader" page?

maurizi:
Fixes #1795


jwalgran:
+1
links work

PR #1908, opened 2015-01-14
Only check the current instance when validating opentreemap id number

### LINKED ISSUE #1894: Importer: Adding tree with OpenTreeMap Plot ID number causes no tree on map
I tried to upload a fresh set of trees and put in an OpenTreeMap Plot ID number of 12345. I assumed since there were no trees in the system, the uploader would indicate this field was an error as there was no tree with the ID 12345 with which to match this new data.

Instead, I got no error and the uploader said it successfully uploaded the tree to my tree map. I can't find that tree on my map, however. It's not at the location it should be and I couldn't find it searching by diameter either. 

The map was on dev and is called phillyuploader.

maurizi:
Fixes #1894


steventlamb:
+1

PR #1911, opened 2015-01-14
Importer error updates

### LINKED ISSUE #1896: Importer: Trees can be edited from successfully added tab
If trees were on the warnings tab, they are still added to the tree map when the user clicks "Add to Tree Map." At that point, they move to the successfully added tab but the Point X and Point Y columns are highlighted yellow and can still be edited. If edited, the tree moves back to the Ready to Add tab.

I think it's confusing for Point X and Point Y to be editable for only some trees on this tab. No other fields can be edited. It seems like trees moved to successfully added should no longer be editable. Thoughts?

RickMohr:
Fixes #1896

### DIFF
--- a/opentreemap/importer/errors.py
+++ b/opentreemap/importer/errors.py
@@ -17,6 +17,9 @@ def get_message(code):
     return _messages_by_code[code]
 
 
+def is_itree_error_code(code):
+    return 60 <= code <= 69
+
 ######################################
 # FILE LEVEL ERRORS
 ######################################
--- a/opentreemap/importer/templates/importer/partials/resolve_popover.html
+++ b/opentreemap/importer/templates/importer/partials/resolve_popover.html
@@ -21,3 +21,9 @@ value="{{ field.value }}"
 data-url="{% url 'importer:update_row' instance_url_name=request.instance.url_name import_type=panel.import_type row_id=field.row_id %}"
 data-field-name="{{ field.name }}"
 {% endblock extra_save_attributes %}
+
+{% block help_text %}
+{% if field.help_text %}
+<p class="help_text">{{ field.help_text }}</p>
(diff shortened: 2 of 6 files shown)


opentreemap/importer/errors.py
```diff
@@ -65,6 +65,7 @@ def get_message(code):
 
 INVALID_UDF_VALUE = e(50, trans('Invalid value for custom field'), True)
 
+FIRST_ITREE_ERROR_CODE = 60
```
steventlamb: how about a function at the top of this module?
```
def is_itree_error_code(code): return 60 < code < 69
```

opentreemap/importer/templates/importer/partials/resolve_popover.html
```diff
@@ -21,3 +21,9 @@
 data-url="{% url 'importer:update_row' instance_url_name=request.instance.url_name import_type=panel.import_type row_id=field.row_id %}"
 data-field-name="{{ field.name }}"
 {% endblock extra_save_attributes %}
+
+{% block suggestion %}
+{% if field.suggestion %}
+<p class="suggestion">{{ field.suggestion }}</p>
```
steventlamb: I have no problem with this, but I'm raising it as a question. I feel like we don't have a strong convention here, but I feel like we often use the pattern of leaving empty dom elements for empty values, taking advantage of silent failing of the template syntax, and configuring the CSS to avoid any spacing when the element is empty.

In this case, that would just mean removing the `if` clause entirely.

Does that sound like the more common approach? Is that a pattern worth considering a convention, and using here instead?

steventlamb:
+1

RickMohr:
@steventlamb, a force push hid some of your comments but I believe I addressed them all.

PR #1913, opened 2015-01-14
Remove gender from species model

### LINKED ISSUE #1870: Importer: remove "Gender" Species field
...per Deb

jwalgran:
The other_part_of_name field is a better place to keep information related to the gender of a species, since it participates in importer species matching.

I confirmed that the production database does not contain any non-blank values in the "gender" column, so a simple delete migration is sufficient.

I left a gender value in an otm1_migrator unit test because OTM1 installations will have a gender column, and it may have values.

Fixes #1870


jwalgran:
Going to close this and merge it onto release.

PR #1914, opened 2015-01-14
Limit* importer uploads to csv and update text

### LINKED ISSUE #1898: Importer: File structure error is unclear
I tried to upload an excel file instead of a CSV. The error message I got was:

There were errors in the file structure: An exception was raised while uploading the file: line contains NULL byte

Anyway this could be adjust to say: Incorrect file type. Please upload a CSV

steventlamb:
fixes #1898 on github

* this has decent but not complete support on popular browsers. It's
supposed to be included in the version of firefox that ships in April,
and it's already supported in Chrome and IE11.


maurizi:
+1

PR #1915, opened 2015-01-15
Fix "Tree Present" handling

RickMohr:
Meaning of "Tree Present" values:
* False: don't create a tree even if tree data is present
* True: do create a tree even if no tree data is present
* blank value or column absent: create a tree only if tree data is present


maurizi:
+1

PR #1929, opened 2015-01-21
Export imports

maurizi:
Mostly this is refactoring already existing views into asynchronous celery tasks


maurizi:
@RickMohr updated, can you take another look?

opentreemap/importer/tasks.py
```diff
@@ -170,3 +172,33 @@ def _get_waiting_row_count(ie):
     return ie.rows()\
              .filter(status=GenericImportRow.WAITING)\
              .count()
+
+
```
RickMohr: I think you can simplify this now, merging `species_export_builder` and `_get_export_builder` into `get_all_species_export` and removing the `all_fields` filter.

opentreemap/importer/models/trees.py
```diff
@@ -44,18 +44,39 @@ def get_udf_column_name(self, udf_def):
...
         plot_udfs = udf_column_names('Plot')
         tree_udfs = udf_column_names('Tree')
 
-        legal_fields = fields.trees.ALL | plot_udfs | tree_udfs
+        return fields.trees.ALL + plot_udfs + tree_udfs
+
+    def legal_and_required_fields(self):
+        legal_fields = set(self.ordered_legal_fields())
 
         return (legal_fields, {fields.trees.POINT_X, fields.trees.POINT_Y})
```
RickMohr: I think these three functions are now unused and should be deleted.

maurizi: Good catch! :hand: :baseball:

RickMohr:
Otherwise, ship it! +1

PR #1930, opened 2015-01-26
Importer: prevent importing duplicate trees (fixes #1903)

### LINKED ISSUE #1903: Importer: prevent importing duplicate trees
If you click "Add to Tree Map" for an import and later click "Add to Tree Map" again for that same import, we don't add a new tree on top of an existing one.

However, if you make two separate imports of the same file, the second import will show "nearby tree" warnings and if you click "Add to Tree Map" then we will add new trees on top of the existing ones.

We should have an **error** "there is already a tree at that location" in addition to our **warning** "there are nearby trees".

RickMohr:


maurizi:
It looks like this had some test failures?

opentreemap/importer/errors.py
```diff
@@ -83,7 +83,9 @@ def is_itree_error_code(code):
 MERGE_REQUIRED = e(71, trans('This row must be merged'), False)
 
 NEARBY_TREES = e(
-    1050, trans('There are already trees very close to this one'), False)
+    1050, trans('There are already trees within ten feet of this one'), False)
```
maurizi: I'm a little leery of making this change.

It may be a little jarring for maps using metric units to see imperial units in their error messages when we require metric units in their upload.

RickMohr: I hear you, but giving no information at all seems worse.

maurizi:
+1

PR #1933, opened 2015-01-29
Add social media sharing buttons

RickMohr:
* Facebook, Twitter, and Google+
* Show on plot detail and resource detail pages for public sites
* Add meta-tags so posts will get a sensible title/picture/description
* Include total ecobenefits in description if possible


opentreemap/treemap/lib/map_feature.py
```diff
@@ -303,8 +309,41 @@ def context_dict_for_map_feature(request, feature):
...
     return context
+
+
+def _add_share_context(context, request, photos):
+    if len(photos) > 0:
+        photo_url = photos[0].thumbnail.url
+    elif context.get('has_tree', False):
+        photo_url = settings.STATIC_URL + "img/tree.png"
+    else:
+        photo_url = settings.STATIC_URL + "img/otmLogo126.png"
```
jwalgran: It may be helpful to drop a comment in here explaining why we are showing the otm logo when there a tree is present and our "generic tree" image when a tree is not present.

opentreemap/treemap/lib/map_feature.py
```diff
@@ -303,8 +309,41 @@ def context_dict_for_map_feature(request, feature):
...
+    photo_url = request.build_absolute_uri(photo_url)
+
+    title = trans("%s on OpenTreeMap site %s") % (context['title'],
```
jwalgran: "Some title on OpenTreeMap site PhillyTreeMap" is a little wordy compared to "Some title on PhillyTreeMap." Is getting the OTM branding in there a specific part of the feature?

opentreemap/treemap/lib/map_feature.py
RickMohr: It's actually a generic tree when a tree is present and the OTM logo otherwise. Maybe 
```python
elif context.get('has_tree', False) == True:
```
would be clearer

opentreemap/treemap/lib/map_feature.py
jwalgran: On a stand-alone install this wording is would be a little more out of place.

RickMohr: I was worried about instances whose name didn't give enough context. But the real ones people have made do have enough context (TreeMapLA, yegTreeMap, GowanusTrees). Then I take your point about stand-alone instances. Will remove OpenTreeMap.

opentreemap/treemap/templates/treemap/map_feature_detail.html
```diff
@@ -5,6 +5,13 @@
 {% load l10n %}
 
 {% block head_extra %}
+{# Facebook & Google+ scrape these properties for their share dialogs #}
+<meta property="og:url" content="{{ share.url }}" />
+<meta property="og:title" content="{{ share.title }}" />
+<meta property="og:description" content="{{ share.description }}" />
+<meta property="og:image" content="{{ share.image }}" />
+<meta property="og:type" content="article" />
+<meta property="og:site_name" content="OpenTreeMap" />
```
jwalgran: Perhaps ``instance.name`` for the ``og:site_name``?

opentreemap/treemap/lib/map_feature.py
RickMohr: Actually
```python
elif context.get('has_tree'):
```
is best since missing values return `None`

jwalgran: Right. Sorry about that. Looks good.

opentreemap/treemap/templates/treemap/map_feature_detail.html
RickMohr: Yes that's better, changed.

jwalgran:
+1
I ran this locally and tested that all three options bring up filled in share sheets.

maurizi:
@RickMohr Any reason this hasn't been merged yet?

PR #1953, opened 2015-02-19
Allow readonly access to all pages for superadmins

steventlamb:


jwalgran:
Because we are talking about management details the number of boolean operators and parens in that if block make me a little uneasy from a future refactor and maintenance standpoint. How do you feel about splitting out a helper method or two to simplify the ``if`` statement?

jwalgran:
I anticipate depending on this feature for customer support, so adding some test coverage to protect against regressions is a good idea.

steventlamb:
I will add test coverage. Certainly.

I tried to add a `def is_super_admin_safe(user)` fn, replacing all occurrence of the whole `hasattr(user, 'is_super_admin)` deal, and I ran into cryptic circular import problems. I can try a little harder, it seems like a good idea. It's not readable in its current form. I wanted to get the idea up for review ASAP before trying too hard.

steventlamb:
@jwalgran pushed a few commits to fix some issues with testing. I think this is ready to merge.

jwalgran:
+1

PR #1955, opened 2015-02-23
Update user audit query to be more efficient

jwalgran:
Trying to view user profiles other than your own with a large database was triggering a 500 response. The root cause was the fact that the ORM was generating a 2.2MB SQL statement, causing Postgres to choke with a stack error.

The audit query code in lib/user.py builds up the final QuerySet by gluing together several small parts. I discovered that the same instance was being fed into the filter multiple times, causing the ORM to generate a lot of duplicate clauses, bloating the query. Adding a ``distinct`` took care of that.

The fast majority of users will only edit a single instance, and our audit table is optimized for filtering by instance ID, so I added an additional query to fetch the ids of a users edited instances and use that to limit the scope of the query, rather than limiting it to the larger list of visible instances.

### DIFF
--- a/opentreemap/treemap/lib/user.py
+++ b/opentreemap/treemap/lib/user.py
@@ -12,6 +12,13 @@ from treemap.models import Plot, Tree, Instance, MapFeature, InstanceUser, User
 from treemap.util import get_filterable_audit_models
 
 
+def _instance_ids_edited_by(user):
+    return Audit.objects.filter(user=user)\
+                        .values_list('instance_id', flat=True)\
+                        .exclude(instance_id=None)\
+                        .distinct()
+
+
 def get_audits(logged_in_user, instance, query_vars, user, models,
                model_id, page=0, page_size=20, exclude_pending=True,
                should_count=False):
@@ -26,8 +33,11 @@ def get_audits(logged_in_user, instance, query_vars, user, models,
     # If we didn't specify an instance we only want to
     # show audits where the user has permission
     else:
-        instances = Instance.objects.filter(
-            user_accessible_instance_filter(logged_in_user))
+        instances = Instance.objects\
+                            .filter(pk__in=_instance_ids_edited_by(user))\
+                            .filter(user_accessible_instance_filter(
+                                logged_in_user))\
+                            .distinct()
 
     if not instances.exists():
         # Force no results


opentreemap/treemap/lib/user.py
```diff
@@ -12,6 +12,14 @@
 from treemap.util import get_filterable_audit_models
 
 
+def _instance_ids_edited_by(user):
+    edited_instance_ids = Audit.objects.filter(user=user)\
+                                       .values_list('instance_id', flat=True)\
+                                       .distinct()
+    # In some cases a None is returned in the values list, so filter it out
+    return [x for x in edited_instance_ids if x]
```
steventlamb: I think it's generally more idiomatic and would guess in this case more performant if this stays lazy as a queryset instead of forcing to a list. Why not `.exclude(instance_id=None)`? That way, callers can execute a single query.

opentreemap/treemap/lib/user.py
```diff
@@ -98,6 +109,11 @@ def get_audits(logged_in_user, instance, query_vars, user, models,
     if exclude_pending:
         audits = audits.exclude(requires_auth=True, ref__isnull=True)
 
+    # Uncomment to dump the large, complex query for debugging
```
steventlamb: We don't have a precedent for committing this type of code comment. Over the lifetime of this project we'll profile many queries, and often times once they've been profiled once, they are done. For these two reasons, I concerned that leaving this comment is not a sustainable idea. It seems likely that these comments will contribute noise and will not be well-placed in the long-term.

Keeping this technique on the brain is important though. Consider writing this up in documentation instead, or factoring it out as a helper that takes a queryset, confirms `settings.DEBUG == True`, and does this.

steventlamb:
+1. Pulled down, tested against my dev data with some of my bigger users that have 10k+ edits.

opentreemap/treemap/lib/user.py
jwalgran: I am going to just remove it. It is easy enough to write this in by hand in the rare cases when you want to dig through a big SQL statement.

maurizi:
+1

opentreemap/treemap/lib/user.py
jwalgran: done

PR #1956, opened 2015-02-24
Fix favorite list image URL construction

jwalgran:
My initial implementation of getting thumbnails for the favorites list had two flaws:

1. I was manually constructing a photo URL, rather than using the
``url`` property of the photo instance. This broke when running in AWS
where photos are in S3 rather than the media directory.

2. When the map feature is a plot, the photo to be displayed is attached
to the current tree, not the plot itself.


maurizi:
+1

steventlamb:
+1

PR #1957, opened 2015-02-24
Speed up instance edits page

jwalgran:
There are two optimizations in this change set that reduce the query time of the instance edits page on my dev vm by 5x

```
Before: 10,700.12 ms (153 queries)
After:   2,741.48 ms (152 queries)
```
When the edits page was created as part of 369749a, an additional sort by id was added to the queryset. This technically produces more correct output, since the audits will be displayed in the exact order they were written to the table, but it causes a 5x slowdown in query execution time.

On my development wm with  5 million row audit table the filtered count query was taking 2 seconds to execute. I was able to remove the call to ``.count()`` entirely because a slice operator is applied to the queryset  earlier in the function, which realizes the quersets and allows us to use len()


maurizi:
+1

RickMohr:
+1 Great sleuthing and good comments.

PR #1967, opened 2015-03-05
Fix issues with invalid UDF names.

### LINKED ISSUE #1765: Allow more characters in UDF field names and add disallowed characters check to UserDefinedFieldDefinition.validate, shared with form_extras template tag
Naming a UDF "50+ cm tally" causes the following exception when trying to show the tree detail page

```
 File "/usr/local/otm/app/opentreemap/treemap/templatetags/form_extras.py",
  line 300, in render 'to follow "from" %s' % identifier)
TemplateSyntaxError: expected a string with the format
"object_name.property" to follow "from" tree.udf:50+ cm tally
```

Which is caused by this regex

https://github.com/OpenTreeMap/OTM2/blob/6d4e7af6658b2930c0f65daea8bdad092ff40ed5/opentreemap/treemap/templatetags/form_extras.py#L30

maurizi:
Adds validation that prevents UDFs from being created if their name contains
any problematic characters.  I tried every special character on a US keyboard
and only found issues with '%', '_', and '.'

'%' and '__' both cause issues for the Django ORM, preventing us from querying
for those UDFs by name.
'.' causes issues when attempting to render the Plot Detail

Also loosens restrictions on UDF names used by the form_extras template tags,
to allow any characters.

Fixes #1765

### DIFF
--- a/opentreemap/treemap/templatetags/form_extras.py
+++ b/opentreemap/treemap/templatetags/form_extras.py
@@ -26,8 +26,12 @@ register = template.Library()
 # Used to check that the identifier follows the format model.field or
 # model.udf:field name, can't be done in the grammar as it can't be checked
 # until looked up in the context
+#
+# We are being a little looser with our restrictions on UDF names here than
+# we are in the UDFD validation methods.  This regex is a sanity check, so it
+# is not essential that we are 100% accurate here
 _identifier_regex = re.compile(
-    r"^[a-zA-Z_.\-]+(?:udf\:[\w '\._-]+|[a-zA-Z0-9_\-]+)$")
+    r"^[a-zA-Z_.\-]+(?:udf\:.+|[a-zA-Z0-9_\-]+)$")
 
 FIELD_MAPPINGS = {
     'IntegerField': 'int',
--- a/opentreemap/treemap/tests/test_audit.py
+++ b/opentreemap/treemap/tests/test_audit.py
@@ -662,23 +662,23 @@ class PendingInsertTest(OTMTestCase):
             datatype=json.dumps({'type': 'choice',
                                  'choices': ['1', '2', '3']}),
             iscollection=False,
-            name='times_climbed')
+            name='times climbed')
 
         set_write_permissions(self.instance, self.commander_user,
-                              'Plot', ['udf:times_climbed'])
+                              'Plot', ['udf:times climbed'])
 
         FieldPermission.objects.create(
             model_name='Plot',
-            field_name='udf:times_climbed',
+            field_name='udf:times climbed',
             permission_level=FieldPermission.WRITE_WITH_AUDIT,
             role=self.pending_user.get_instance_user(self.instance).role,
             instance=self.instance)
 
         initial_plot = Plot(geom=self.p1, instance=self.instance)
-        initial_plot.udfs['times_climbed'] = '2'
+        initial_plot.udfs['times climbed'] = '2'
         initial_plot.save_with_user(self.pending_user)
 
-        udf_audit = Audit.objects.get(model='Plot', field='udf:times_climbed',
+        udf_audit = Audit.objects.get(model='Plot', field='udf:times climbed',
                                       model_id=initial_plot.pk)
         approve_or_reject_audit_and_apply(udf_audit, self.commander_user,
                                           approved=True)
@@ -704,7 +704,7 @@ class PendingInsertTest(OTMTestCase):
         self.assertEqual(new_plot.pk, initial_plot.pk)
         self.assertEqual(new_plot.readonly, False)
         self.assertEqual(new_plot.geom, self.p1)
-        self.assertEqual(new_plot.udfs['times_climbed'], '2')
+        self.assertEqual(new_plot.udfs['times climbed'], '2')
 
     def test_lots_of_trees_and_plots(self):
         """
--- a/opentreemap/treemap/tests/test_udfs.py
+++ b/opentreemap/treemap/tests/test_udfs.py
@@ -696,6 +696,31 @@ class UDFDefTest(OTMTestCase):
               'default': 'anything'}],
             iscollection=True)
 
+    def test_invalid_names(self):
+        with self.assertRaises(ValidationError):
+            UserDefinedFieldDefinition.objects.create(
+                instance=self.instance,
+                model_type='Plot',
+                datatype=json.dumps({'type': 'string'}),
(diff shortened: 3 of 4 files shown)


RickMohr:
Did you query the production data to see if there are existing UDFs containing the characters we are invalidating?

maurizi:
Yes, there aren't any UDFs with those problem characters on OTM02.

opentreemap/treemap/templatetags/form_extras.py
```diff
@@ -27,7 +27,7 @@
 # model.udf:field name, can't be done in the grammar as it can't be checked
 # until looked up in the context
 _identifier_regex = re.compile(
-    r"^[a-zA-Z_.\-]+(?:udf\:[\w '\._-]+|[a-zA-Z0-9_\-]+)$")
+    r"^[a-zA-Z_.\-]+(?:udf\:.+|[a-zA-Z0-9_\-]+)$")
```
jwalgran: It seems like we sync this with ``_UDF_NAME_REGEX``

```python
r"^[a-zA-Z_.\-]+(?:udf\:[^_%.]+|[a-zA-Z0-9_\-]+)$")
```

vs.

```python
r"^[a-zA-Z_.\-]+(?:udf\:.+|[a-zA-Z0-9_\-]+)$")
```

maurizi: I purposefully didn't do that.  We may need to add more restrictions to the UDF name regex in the future, and I thought it would quickly get out of sync from this regex.  This only really "sanity checking", so there isn't a need to be as restrictive here as we are when creating UDFs.

jwalgran: If synchronization is the only thing we are worried about, we can make a single source of truth

```
UDF_NAME_PATTERN = r'[^_%.]+'
_UDF_NAME_REGEX = re.compile(UDF_NAME_PATTERN)
```

```
_identifier_regex = re.compile(
    r"^[a-zA-Z_.\-]+(?:udf\:" + UDF_NAME_PATTERN + r"|[a-zA-Z0-9_\-]+)$")
```

If you don't want to do that, I would like to embed your concern in a comment above the "looser" regex.

jwalgran:
When I try adding a UDF named "75% dead" I get an exception:

```
IndexError at /jwtrialtest2014-09-30/management/udfs/
tuple index out of range

Traceback:
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  114.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  22.                 return view_func(request, *args, **kwargs)
File "/usr/local/otm/app/opentreemap/treemap/decorators.py" in wrapper
  45.             return view_fn(request, instance, *args, **kwargs)
File "/usr/local/otm/app/opentreemap/treemap/decorators.py" in f
  69.                 return view_fn(request, instance, *args, **kwargs)
File "/usr/local/otm/app/opentreemap/treemap/decorators.py" in wrapped
  119.                 return view_fn(request, instance, *args, **kwargs)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django_tinsel/decorators.py" in routed
  33.             return req_method(request, *args2, **kwargs2)
File "/usr/local/otm/app/opentreemap/treemap/decorators.py" in run_and_catch_validations
  144.             return req(*args, **kwargs)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django_tinsel/decorators.py" in wrapper
  79.                 params = callable_or_dict(request, *args, **wrapper_kwargs)
File "/usr/local/otm/app/opentreemap/cloud_management/udf.py" in udf_create
  76.     return udf_context(instance, udf)
File "/usr/local/otm/app/opentreemap/cloud_management/udf.py" in udf_context
  98.                                 .exclude(**{udf_def.canonical_name: ""})\
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/models/query.py" in count
  291.         return self.query.get_count(using=self.db)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/models/sql/query.py" in get_count
  390.         number = obj.get_aggregation(using=using)[None]
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/contrib/gis/db/models/sql/query.py" in get_aggregation
  87.         return super(GeoQuery, self).get_aggregation(using)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/models/sql/query.py" in get_aggregation
  356.         result = query.get_compiler(using).execute_sql(SINGLE)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py" in execute_sql
  781.         cursor.execute(sql, params)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/backends/util.py" in execute
  69.             return super(CursorDebugWrapper, self).execute(sql, params)
File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/db/backends/util.py" in execute
  53.                 return self.cursor.execute(sql, params)

**Comment by maurizi** [comment]
@jwalgran @RickMohr Can you take another look?

**Comment by jwalgran** [comment]
Testing this now

**Comment by jwalgran** [comment]
I was able to add a field with the name "81% dead" Regex validation is not working.

```python
 if not _UDF_NAME_REGEX.match(self.name):
            raise ValidationError(
                {'name': [trans("A field name may not contain percent (%), "
                                "period (.) or underscore (_) characters.")]})
```

It is looking for the match response to be falsey, which it isn't

```
(Pdb) self.name
u'81% dead'
(Pdb) _UDF_NAME_REGEX.pattern
u'[^_%.]+'
(Pdb) _UDF_NAME_REGEX.match(self.name)
<_sre.SRE_Match object at 0x7f813c592098>
(Pdb) _UDF_NAME_REGEX.match(self.name).group(0)
u'81'

maurizi:
@jwalgran Oops, thats what I get for testing with single character UDF names.  I updated the regex to require matching on the entire name (`^[^_%.]+$`)

jwalgran:
I would like to provide some guidance in the template, telling the customer about the disallowed characters before they try to use them. Could be in a separate issue/pull.

jwalgran:
+1, prevents those nasty chars

maurizi:
@jwalgran I think that's a good idea, but I'd like to do it as a separate PR/issue

PR #1969, opened 2015-03-11
Add updated_at to MapFeature

### LINKED ISSUE #1972: Add updated_at column to MapFeature and backfill from audits

jwalgran:
We have decided to keep a denormalized copy of the time each MapFeature was last updated to facilitate simple querying. This commit adds a new updated_at field and functionality to keep it maintained.

Managing updates to this new field when only a child object changes (Tree, Photo, etc.) is tricky. I opted to target the easily findable save methods and add a call to a new function that updates the parent MapFeature row directly, rather than depend on save_with_user being called on the MapFeature instance when none of the properties of that object have changed.

Adding updated_at created an interesting field permission wrinkle. I did not want to have to create explicit permission objects to control this "special" field. Adding a tree photo will require permission to edit updated_at, even if the user has no other writable permissions on that object, so I opted to use a class-level property to flag a field as being world writable.

I implemented the backfilling of updated_at values with raw SQL queries for efficiency. Running the function with 500k map features and 5.6m audit records takes < 3 minutes. I also find the nested joins in the SQL queries easier to understand than complex ORM calls.

Fixes #1972

### DIFF
--- a/opentreemap/treemap/lib/map_feature.py
+++ b/opentreemap/treemap/lib/map_feature.py
@@ -14,10 +14,12 @@ from django.db.models import Q
 
 from treemap.audit import Audit
 from treemap.ecobackend import ECOBENEFIT_ERRORS
+from treemap.lib import execute_sql
 from treemap.models import Tree, MapFeature, User, Favorite
 
 from treemap.lib import format_benefits
 from treemap.lib.photo import context_dict_for_photo
+from treemap.util import leaf_subclasses
 
 
 def _map_feature_audits(user, instance, feature, filters=None,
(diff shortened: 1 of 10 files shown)


opentreemap/treemap/lib/map_feature.py
```diff
@@ -358,3 +359,64 @@ def _add_share_context(context, request, photos):
...
+    # NOTE: This MUST be run first. Additional update queries compare
+    # dates against the updated_at values set by this statement.
+    cursor.execute("""
+UPDATE treemap_mapfeature
+SET updated_at = a.updated_at
+FROM (
+  SELECT model_id as id, MAX(created) AS updated_at
+  FROM treemap_audit
+  WHERE treemap_audit.model IN ('Plot', 'RainBarrelLA', 'RainGardenLA',
+                                'TurfConcreteGardenLA')
```
maurizi: For a bit of future-proofing, consider replacing this hardcoded list of model names with a dynamically generated one by calling `[Model.map_feature_type for Model in leaf_subclasses(MapFeature)]`

opentreemap/treemap/models.py
```diff
@@ -953,6 +978,7 @@ def is_attr_set(attr):
         elif is_attr_set('tree'):
             self.map_feature = self.tree.plot
 
+        self.map_feature.update_updated_at()
```
maurizi: This is unnecessary, since the same call is made in `MapFeaturePhoto`, which is the superclass of `TreePhoto`.  Consider removing it (and possibly adding a comment to explain why it is unnecessary)

maurizi:
I think we need to add calls to `MapFeature.update_updated_at()` in the `delete_with_user` methods of `Tree` and `MapFeaturePhoto`

jwalgran:
We certainly do.

maurizi:
I like the general approach of this PR.

I think raw SQL was a good idea, and adding the `joint_writable_set` property certainly simplifies all of our other permission checking logic (permissions was a hassle I was trying to think of how to overcome for the advanced search).

I'm going to hold off on a +1 until some tests for the raw SQL are written though.

opentreemap/treemap/models.py
jwalgran: Removed

maurizi:
@jwalgran Is this ready for another look?  Or do you still need to write more tests?

jwalgran:
I will be writing some tests for the ``set_map_feature_updated_at`` function and will assign the PR when they are done.

opentreemap/treemap/tests/test_updated_at.py
```diff
@@ -0,0 +1,97 @@
...
+        treephoto_audit = self.max_audit_for_model_type(['MapFeaturePhoto',
+                                                         'TreePhoto'])
+        plot_audit = self.max_audit_for_model_type('Plot')
+        # Backdate the audits so photo it is definitely the newsest
+        plot_audit.created = treephoto_audit.created - timedelta(days=2)
+        plot_audit.save()
+        tree_audit.created = treephoto_audit.created - timedelta(days=1)
+        tree_audit.save()
+
+        # import pdb; pdb.set_trace()
```
maurizi: You should remove this commented out pdb trace

maurizi:
+1 after removing the commented out `pdb.set_trace()`

jwalgran:
Merging because tests pass locally. See #1977 for CircleCI fix.

Issue #1910, opened 2015-01-14
List the map's i-Tree regions on the basic information tab


dboyer:
This information is useful if the user wants to upload new species via the bulk uploader.

steventlamb:
Outside the scope of this repo.

PR #1971, opened 2015-03-11
Add `itree_regions` method to instance

steventlamb:
addresses part of #1910

For presentation purposes


jwalgran:
+1 (did not pull and test)

PR #1973, opened 2015-03-13
Add a plugin to filter out instances from the public instance list

maurizi:


RickMohr:
+1 but note failing tests

maurizi:
@RickMohr The test failures seem to have been an intermittent failure.

PR #1976, opened 2015-03-16
Gate importer on tree limit instead of feature

steventlamb:


RickMohr:
For limit checking we need to also count the existing trees.

steventlamb:
@RickMohr updated to integrate your suggestions

PR #1988, opened 2015-03-31
Prevent submitting empty comments.

### LINKED ISSUE #1596: Attempting to post an empty comment redirects to an unstyled page
1. Log in
  1. Go to https://www.opentreemap.org/edmonton/features/130854/
  1. Leave the comment box empty and click "Post"

screen shot 2014-09-10 at 2 26 20 pm

maurizi:
Attempting to post an empty comment would lead to incorrect behaviour
like redirecting to an unstyled page.

This commit fixes the issue by disabling the submit button when the
textbox is empty, and enabling when text has been entered.

Fixes #1596


RickMohr:
+1

PR #1989, opened 2015-04-06
Mark export jobs as failed on async exception

### LINKED ISSUE #1370: CSV export errors are never reported to the user
The client continuously polls for the job status and the server keeps returning "PENDING" even though the celery job failed.

steventlamb:
fixes #1370 on github

This changeset wraps the logic of async export jobs to mark the job as
failed in the case of an uncaught exception. Jobs will be marked failed
and then the exception will reraise, for logging or further handling in
the future.

If celery is not running, jobs will still poll forever, and once celery
is started, they will be picked up and finished.

### DIFF
--- a/opentreemap/exporter/models.py
+++ b/opentreemap/exporter/models.py
@@ -12,12 +12,14 @@ from treemap.instance import Instance
 
 
 class ExportJob(models.Model):
+    UNCAUGHT_EXCEPTION_ERROR = -1
     PENDING = 0
     EMPTY_QUERYSET_ERROR = 1
     MODEL_PERMISSION_ERROR = 2
     COMPLETE = 3
 
     STATUS_STRINGS = {
+        UNCAUGHT_EXCEPTION_ERROR: 'UNCAUGHT_EXCEPTION_ERROR',
         PENDING: 'PENDING',
         EMPTY_QUERYSET_ERROR: 'EMPTY_QUERYSET_ERROR',
         MODEL_PERMISSION_ERROR: 'MODEL_PERMISSION_ERROR',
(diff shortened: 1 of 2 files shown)


opentreemap/exporter/models.py
```diff
@@ -58,3 +61,6 @@ def get_url_if_ready(self):
     def complete_with(self, filename, file_obj):
         self.outfile.save(filename, file_obj)
         self.status = self.COMPLETE
+
+    def fail(self):
+        self.status = self.UNCAUGHT_EXCEPTION_ERROR
```
maurizi: I don't think there is case where we'll use `fail` and *not* immediately save the `ExportJob`.  As such, consider refactoring this method like so:
```python
def record_failure(self):
    self.status = self.UNCAUGHT_EXCEPTION_ERROR
    self.save()
```
And omitting the calls to `job.save()` below.

steventlamb: Fair point. I thought about this, but decided to leave it explicit (and repetitive) because this is how the similar method `complete_with` works. Also, I think django model methods that call `save()` and don't have `save` in the name are a little bit dangerous.

opentreemap/exporter/tasks.py
```diff
@@ -161,12 +174,19 @@ def simple_async_csv(job_pk, qs):
...
     job = ExportJob.objects.get(pk=job_pk)
 
-    csv_obj = TemporaryFile()
-
-    writer = csv.DictWriter(csv_obj, fields)
-    writer.writeheader()
-    for row in csv_rows:
-        writer.writerow(sanitize_unicode_record(row))
+    # TODO: roll this into the @job_transaction pattern
+    # OR decide to leave separate and remove this TODO.
```
maurizi: I don't see any reason why you can't just use `@job_transaction` on this task too.  It doesn't seem any functionally different from whats being done here.

steventlamb: `@job_transaction` and celery subtasking and threading are both fighting for control of the argument order. It would require a change to the caller and/or some refactoring here.

maurizi: Makes sense

steventlamb:
@maurizi updated to kill TODO and DRY out logic

maurizi:
+1

PR #1992, opened 2015-04-06
Allow map admins to disable exports

### LINKED ISSUE #1359: Give map owners option to disable exports

jwalgran:
This commit adds a boolean attribute to the instance model for controlling whether or not non-administrators can download data as CSV.

The flag is checked by a helper method both at the template level, to hide the export button, and at the task level to prevent a template bug or manually crafted request from "leaking" the data when the admins have disallowed it.

The new helper extends the existing check of whether the export feature is enabled at all and, as a result, applies that check in more places than it was being applied previously.

Fixes #1359

### DIFF
--- a/opentreemap/exporter/__init__.py
+++ b/opentreemap/exporter/__init__.py
@@ -1,0 +1,7 @@
+from django.utils.translation import ugettext_lazy as trans
+
+
+EXPORTS_NOT_ENABLED_CONTEXT = {
+    'start_status': 'ERROR',
+    'message': trans('Data exports are not enabled.')
+}
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -248,6 +248,8 @@ class Instance(models.Model):
     mobile_api_fields = _make_config_property('mobile_api_fields',
                                               DEFAULT_MOBILE_API_FIELDS)
 
+    non_admins_can_export = models.BooleanField(default=True)
+
     @property
     def advanced_search_fields(self):
         # TODO pull from the config once users have a way to set search fields
--- a/opentreemap/exporter/decorators.py
+++ b/opentreemap/exporter/decorators.py
@@ -6,6 +6,8 @@ from __future__ import division
 from functools import wraps
 from celery import chain
 
+from exporter import EXPORTS_NOT_ENABLED_CONTEXT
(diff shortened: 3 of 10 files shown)


opentreemap/treemap/instance.py
```diff
@@ -449,6 +451,22 @@ def is_accessible_by(self, user):
...
             return False
 
+    def export_enabled_for(self, user):
+        if self.feature_enabled('exports'):
+            if self.non_admins_can_export:
+                return True
+            else:
+                if user.is_authenticated():
+                    iuser = user.get_instance_user(self)
+                    return iuser and iuser.admin
```
steventlamb: if `iuser` is falsy but not `False`, this will sometimes fail to return a `Boolean`. Since the rest of this function clearly returns `True` or `False`, consider making this line do it too, for consistency. I understand the callers will just be doing truthy/falsy checks so this is not an issue now, but it can't hurt in case somewhere down the line a different kind of check is added.

opentreemap/treemap/instance.py
```diff
@@ -449,6 +451,22 @@ def is_accessible_by(self, user):
         except ObjectDoesNotExist:
             return False
 
+    def export_enabled_for(self, user):
```
steventlamb: I think it would make a little more sense to have `def export_enabled_for(instance, user)` in the `exporter/` app, for encapsulation. If still necessary, this method could import and thinly wrap it. This technique would have consistency with how the `EXPORTS_NOT_ENABLED_CONTEXT` dict is used.

steventlamb:
+1, tested, works.

opentreemap/treemap/instance.py
jwalgran: Good idea. I like doing nice things for future us.

opentreemap/treemap/instance.py
jwalgran: I like the idea of moving it to ``exporter``

opentreemap/exporter/__init__.py
```diff
@@ -0,0 +1,7 @@
+from django.utils.translation import ugettext as trans
```
maurizi: I think this needs to be `ugettext_lazy`

maurizi:
+1

PR #1994, opened 2015-04-08
Make importer event statuses unique

jwalgran:
The CANCELED status code was overlapping with the LOADING status code.

### DIFF
--- a/opentreemap/importer/models/base.py
+++ b/opentreemap/importer/models/base.py
@@ -28,7 +28,7 @@ class GenericImportEvent(models.Model):
     CREATING = 4
     FINISHED_CREATING = 5
     FAILED_FILE_VERIFICATION = 6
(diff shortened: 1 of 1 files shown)


opentreemap/importer/models/base.py
```diff
@@ -28,7 +28,7 @@ class Meta:
     CREATING = 4
     FINISHED_CREATING = 5
     FAILED_FILE_VERIFICATION = 6
-    CANCELED = 7
+    CANCELED = 8
```
maurizi: Consider moving `LOADING` to be above cancelled, so all of the statuses are in numerical order, which should hopefully prevent this from happening again in the future

maurizi:
+1

jwalgran:
The constants are arranged in workflow order, so I am going to leave them as is.

PR #1995, opened 2015-04-09
Add states to import validation

### LINKED ISSUE #1979: Import file with 260k+ rows stuck on Verifying
On the dev server

screen shot 2015-03-18 at 12 22 03 pm

~~I tried the import twice, and this same error appeared twice in the celery log~~

This was almost definitely caused by the fact that I restarted celeryd

```
[2015-03-18 14:19:54,011: ERROR/MainProcess] Task importer.tasks.run_import_event_validation[b6dfe41d-b38e-4881-b1ae-699a5b34fbcc] raised unexpected: WorkerLostError('Worker exited prematurely: signal 15 (SIGTERM).',)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/billiard/pool.py", line 1169, in mark_as_worker_lost
    human_status(exitcode)),
WorkerLostError: Worker exited prematurely: signal 15 (SIGTERM).
```

jwalgran:
Import jobs were appearing to be "hung" in the validation step of a tree import for two reasons:

1. The process of chunking up a large import file into tasks takes some time. Rows are not being processed during this setup phase, so the counter appeared to be stuck at zero.

2. An error during the verification setup would cause the task to fail silently without updating the import event status, which remained stuck at "Verifying."

I added two additional import event states to cover these two cases, PREPARING_VERIFICATION and VERIFICATION_ERROR.

Fixes #1979

I tested the PREPARING_VERIFICATION state by importing a 100,000 tree CSV and watching the management page change status

screen shot 2015-04-08 at 4 45 52 pm

I tested the VERIFICATION_ERROR status by inserting a ``raise`` into the try block and then importing a tree file.

screen shot 2015-04-08 at 9 42 47 pm

### DIFF
--- a/opentreemap/importer/tasks.py
+++ b/opentreemap/importer/tasks.py
@@ -87,18 +87,30 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
             pass
         return
 
-    ie.status = GenericImportEvent.VERIFIYING
+    ie.status = GenericImportEvent.PREPARING_VERIFICATION
     ie.save()
 
-    row_set = ie.rows()
-    validation_tasks = (_validate_rows.subtask(row_set[i:(i+BLOCK_SIZE)])
-                        for i in xrange(0, ie.row_count, BLOCK_SIZE))
+    try:
+        row_set = ie.rows()
+        validation_tasks = (_validate_rows.subtask(row_set[i:(i+BLOCK_SIZE)])
+                            for i in xrange(0, ie.row_count, BLOCK_SIZE))
 
-    final_task = _finalize_validation.si(import_type, import_event_id)
-    res = chord(validation_tasks, final_task).delay()
+        final_task = _finalize_validation.si(import_type, import_event_id)
+        res = chord(validation_tasks, final_task).delay()
 
-    ie.task_id = res.id
-    ie.save()
+        ie.status = GenericImportEvent.VERIFIYING
+        ie.task_id = res.id
+        ie.save()
+    except Exception as e:
+        ie.status = GenericImportEvent.VERIFICATION_ERROR
+        ie.save()
+        try:
+            ie.append_error(errors.GENERIC_ERROR, data=[str(e)])
+            ie.save()
+            ie.row_set().delete()
+        except Exception:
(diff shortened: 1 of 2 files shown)


steventlamb:
looking at this now

opentreemap/importer/tasks.py
```diff
@@ -87,18 +87,30 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
...
-                        for i in xrange(0, ie.row_count, BLOCK_SIZE))
+    try:
+        row_set = ie.rows()
+        validation_tasks = (_validate_rows.subtask(row_set[i:(i+BLOCK_SIZE)])
+                            for i in xrange(0, ie.row_count, BLOCK_SIZE))
 
-    final_task = _finalize_validation.si(import_type, import_event_id)
-    res = chord(validation_tasks, final_task).delay()
+        final_task = _finalize_validation.si(import_type, import_event_id)
+        raise Exception('Testing failure')
```
steventlamb: remove this?

opentreemap/importer/tasks.py
```diff
@@ -87,18 +87,30 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
...
+        res = chord(validation_tasks, final_task).delay()
 
-    ie.task_id = res.id
-    ie.save()
+        ie.status = GenericImportEvent.VERIFIYING
+        ie.task_id = res.id
+        ie.save()
+    except Exception as e:
+        try:
```
steventlamb: this inner try block is pretty ambiguous. It makes sense that the outer try is a generic catch-all because it dumps to a generic error, but can this one be more specific?

I can't tell which statement you were expecting to catch on. The only one that looks like it might fail is `ie.append_error(errors.GENERIC_ERROR, data=[str(e)])`. If so, consider isolating that in the try block, and/or catching something more descriptive and/or adding comments.

steventlamb:
Works after removing that `raise`. (Conditional +1)

jwalgran:
Sorry. I removed that ``raise`` right after I made the PR but did not force push. Will do.

PR #1997, opened 2015-04-14
Prevent inadvertent modification of cached objects by returning copies

### LINKED ISSUE #1764: Protect cached objects from modification
In OpenTreeMap/otm2-addons#354, we discovered code that fetched a cached object, changed one of its attributes, and tried to save. When the save failed the changed attribute remained in the cache, with the incorrect attribute value.

Proposed solution is to always `deepcopy` a cached object before returning it.

OpenTreeMap/otm2-addons#354 fixed the issue with a workaround. That workaround should be removed as part of implementing the general fix.

maurizi:
Also adds support for deepcopy to `DotDict`, which should fix issues with
pickling it.

Fixes #1764

### DIFF
--- a/opentreemap/treemap/DotDict.py
+++ b/opentreemap/treemap/DotDict.py
@@ -1,7 +1,12 @@
+from copy import deepcopy
+
+
 class DotDict(dict):
     """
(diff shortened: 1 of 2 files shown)


RickMohr:
+1

PR #2002, opened 2015-04-27
Allow resetting of diameter/circumeference fields to undefined.

### LINKED ISSUE #978: Diameter Calculator sets other field to 0 when a field is cleared
If you clear out the diameter for a field, it's circumference gets set to `0.0` (and vice versa when clearing circumference).
Deleting the text in one field should do the same in the other.
Presumably, if all diameters are empty, the "calculated diameter" should be as well (as opposed to 0)

maurizi:
Previously, clearing a field would change it's value to 0.  Thus, if you
accidentally saved a diameter for a tree when you were unsure of the
diameter, there was no way to undo your action.  This change allows empty
values to pass through, so the backend can clear the field and set it
back to None.

Fixes #978

### DIFF
--- a/opentreemap/treemap/js/src/diameterCalculator.js
+++ b/opentreemap/treemap/js/src/diameterCalculator.js
@@ -21,14 +21,14 @@ function elementToValue(el) {
     return $(el).data('value');
 }
 
-function zeroToEmptyString(n) {
-    return n === 0 ? '' : n;
-}
-
 function textToFloat(t) {
     var f = parseFloat(t);
 
-    return isNaN(f) ? 0 : f;
+    return isValidNumber(f) ? f : 0;
+}
+
+function isValidNumber(num) {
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/js/src/diameterCalculator.js
```diff
@@ -21,14 +21,14 @@ function elementToValue(el) {
...
-}
-
 function textToFloat(t) {
     var f = parseFloat(t);
 
-    return isNaN(f) ? 0 : f;
+    return isValidNumber(f) ? f : 0;
+}
+
+function isValidNumber(num) {
```
steventlamb: because we used Ramda, if desired, you could probably convert this to:
`var isValidNumber = R.and(_.isNumber, R.not(isNan));`

steventlamb:
+1, tested a bunch of cases and this works

PR #2012, opened 2015-05-21
Add validation for Instance.mobile_api_fields configuration

maurizi:


RickMohr:
+1. Thanks for the thorough unit tests.

Not tested. The testing that counts (other than the unit tests) will happen when we build the UI.

Issue #2010, opened 2015-05-19
Change boundary filters to use within rather than contained


jwalgran:
Boudary search is currently using a ``contained`` clause

https://github.com/OpenTreeMap/OTM2/blob/master/opentreemap/treemap/search.py#L282-L284.

``contained`` uses bounding boxes, which is too general for the filtering we are trying to do.

https://docs.djangoproject.com/en/1.6/ref/contrib/gis/geoquerysets/#std:fieldlookup-contained

We should switch to ``within``

https://docs.djangoproject.com/en/1.6/ref/contrib/gis/geoquerysets/#within

PR #2014, opened 2015-05-28
Fix boundary filter on ecobenefits/counts.

maurizi:
We were using `contained`, which determines if the geometry is within
the bounding box of a boundary.  The more appropriate filter to use is
`within`, which will only match when the geometry is entirely within
the boundary.

Connects to #2010


steventlamb:
+1, didn't test

Issue #2004, opened 2015-05-12
Triggers significantly slowing down bulk import process


hectcastro:
Part of the bulk import process inserts rows into the `treemap_mapfeature` table:

```sql
INSERT INTO "treemap_mapfeature" 
  ( 
    "udfs", 
    "instance_id", 
    "the_geom_webmercator", 
    "address_street", 
    "address_city", 
    "address_zip", 
    "readonly", 
    "updated_at", 
    "feature_type" 
  ) 
    VALUES 
  ( 
    hstore(array['Owned By'], array['Joker']),
    21, 
    st_geomfromewkb('\x0101000020110f00003bcac6d0e32d69c1508edbf6b0f04e41'::bytea),
    NULL, 
    'Gotham City', 
    NULL, 
    FALSE, 
    '2015-05-12 19:22:11.300790+00:00', 
    'Plot' 
  ) 
    returning "treemap_mapfeature"."id"
```

This insert statement activates the following trigger, which activates yet another trigger:

```sql
CREATE TRIGGER revinserttrigger
  AFTER INSERT
  ON treemap_mapfeature
  FOR EACH ROW
  WHEN ((new.the_geom_webmercator IS NOT NULL))
  EXECUTE PROCEDURE revupdate();
```

Body of `revupdate()`:

```sql
CREATE OR REPLACE FUNCTION revupdate()
  RETURNS trigger AS
$BODY$
 DECLARE
   iid integer;
 BEGIN
 IF (TG_OP='INSERT') THEN
   iid = NEW.instance_id;
 ELSIF (TG_OP='UPDATE') THEN
   iid = NEW.instance_id;
 ELSIF (TG_OP='DELETE') THEN
   iid = OLD.instance_id;
 END IF;
 UPDATE treemap_instance SET geo_rev=geo_rev+1 WHERE id=iid ;
 Return NEW;
 END;
 $BODY$
  LANGUAGE plpgsql VOLATILE
  COST 100;
ALTER FUNCTION revupdate()
  OWNER TO otm;
```

This series of triggers yields very poor performance for any insert for the `treemap_mapfeature` table:

```
Insert on treemap_mapfeature  (cost=0.00..0.01 rows=1 width=0) (actual time=1.150..1.154 rows=1 loops=1)
  ->  Result  (cost=0.00..0.01 rows=1 width=0) (actual time=0.162..0.163 rows=1 loops=1)
Trigger revinserttrigger: time=95071.921 calls=1
[...]

PR #2015, opened 2015-05-28
Remove triggers

RickMohr:
Inserting a row into `treemap_mapfeature` on my development VM now takes 1 ms instead of 100 ms.

Importing 10,000 trees on my development VM now takes 8 minutes instead of 17 minutes. 
(Or, excluding verification time, 5 minutes instead of 14 minutes.)

Connects #2004 

To test:
* Verify that adding, moving, or deleting a tree is shown correctly on the map.
* Verify that the `INSERT` query in #2004 takes 1ms


public_key
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ==
```
maurizi: I don't think you want to include this file....

opentreemap/treemap/instance.py
```diff
@@ -442,6 +442,10 @@ def static_page_names(self):
 
         return names
 
+    def update_geo_rev(self):
+        self.geo_rev += 1
+        self.save()
```
maurizi: To avoid race conditions, consider using [an F expression here](https://docs.djangoproject.com/en/1.7/ref/models/queries/), like we do with `object_caches`

RickMohr: Good idea, done. I actually think it would be fine if two servers got the same updated `geo_rev` back in a race situation, but the F expression is  still the right choice.

jwalgran:
+1 I verified that adds, moves, deletes, and bulk imports update the geo rev and inserts are < 1ms in dev.

Issue #1998, opened 2015-04-16
Can't import CSV with Mac line endings


RickMohr:
Error message is "new-line character seen in unquoted field - do you need to open the file in universal-newline mode?"

jwalgran:
Are we talking about a file from classic Mac OS that has CR instead of LF endings? Do we have a sample file, or steps to reproduce?

The comments on the SO answer hint that CSVs saved by Excel may use CR. http://superuser.com/a/439443

RickMohr:
I have a sample file but can't attach it to this issue and was hesitant to post an internal file path to a public repo. But here it is:

`\\fileshare\users\rmohr\bugs\OTM_Bulk_Uploader_Webinar.csv`

jwalgran:
Looks like opening the file in ``'rU'`` mode may fix the issue

http://stackoverflow.com/questions/6726953/open-the-file-in-universal-newline-mode-using-csv-module-django

PR #2018, opened 2015-05-28
Update importer to handle CR terminated CSV files

jwalgran:
Using ``csv.Sniffer`` provides us with alternate line ending detection, and we get things like tab separated file detection and support for free.

Files I used to test the imports:
http://cl.ly/1B3b1I0X2Z11/1998_csv_import_test_files.zip

Connects to #1998

### DIFF
--- a/opentreemap/importer/tasks.py
+++ b/opentreemap/importer/tasks.py
@@ -3,7 +3,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
-import csv
 import json
 
 from celery import task, chord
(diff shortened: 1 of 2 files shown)


RickMohr:
+1. I verified that without your changes `tree_points_cr.csv` and `OTM_Bulk_Uploader_Webinar_TABS.csv` get an "Invalid File Structure" error, and that with your changes all files import successfully.

I did notice that the first row of `OTM_Bulk_Uploader_Webinar.csv` has a value of `Agué` that the bulk uploader displays as `AguÃ©` both before and after your changes.

Issue #1985, opened 2015-03-30
makemessages does not work with translation functions aliased to "trans"


jwalgran:
See this SO answer: http://stackoverflow.com/a/16510454

The root of the problem is the fact that ``makemessages`` calls ``xgettext`` with a fixed set of "keywords." ``trans`` is not one of those keywords.

https://github.com/django/django/blob/master/django/core/management/commands/makemessages.py#L98

Identified by Filipe Marques and discussed on the opentreemap-user mailing list Mon, 30 Mar 2015 11:54:22

What is the tradeoff between switching to ``_`` vs switching to ``ugettrans/ugetrans_lazy``?

RickMohr:
I propose using `_` instead of `trans` for translations, changing e.g. 
```py
header = trans('Tree Benefits')
```
to
```py
header = _('Tree Benefits')
```
Currently we use underscore (`_`) for unused variables; I propose changing to double underscore (`__`), so instead of
```py
_, field = pair.split('.')`
```
we would have
```py
__, field = pair.split('.')`
```
To me that looks better than the other alternatives I saw for unnamed variables:
```py
_model, field = pair.split('.')`
```
```py
unused_model, field = pair.split('.')`
```

jwalgran:
Double underscore looks good to me.

PR #2022, opened 2015-05-29
Use underscore for translations instead of for unused variables

RickMohr:
* Change e.g. `trans('Tree Benefits')` to `_('Tree Benefits')`
* Change e.g. `_, field = pair.split('.')` to `__, field = pair.split('.')`

Changed via `sed`:
```
find . -name "*.py" -print | xargs sed -b -i 's/ _, / __, /g'
find . -name "*.py" -print | xargs sed -b -i 's/, _ /, __ /g'
find . -name "*.py" -print | xargs sed -b -i 's/, _)/, __)/g'
find . -name "*.py" -print | xargs sed -b -i 's/trans(/_(/g'
find . -name "*.py" -print | xargs sed -b -i 's/import ugettext as trans/import ugettext as _/g'
find . -name "*.py" -print | xargs sed -b -i 's/import ugettext_lazy as trans/import ugettext_lazy as _/g'
```

Connects #1985

To test:
```sh
cd opentreemap
mkdir locale
./manage.py makemessages -l en
```


maurizi:
+1

Issue #1991, opened 2015-04-06
Adding a tree to a planting site via Quick Edit does not trigger an update to the count/eco


jwalgran:
1. Add an empty planting site. Observe the "planting sites" count increments.
1. Select the added planting site
1. Click quick edit.
1. Add a species
1. Click save

Expected: The "tree" count increments
Actual: No change

PR #2023, opened 2015-05-29
Update the ecobenefits and plot/tree counts when saving via quick edit.

maurizi:
Ensures that adding a tree or changing diameter or species is reflected
in tree counts and ecobenefits results.

Connects to #1991 

To test:
1. Ensure that adding a tree to an existing plot via "Quick Edit" will increment the plot/tree counts and update the ecobenefits.

### DIFF
--- a/opentreemap/treemap/js/src/treeMapModes.js
+++ b/opentreemap/treemap/js/src/treeMapModes.js
@@ -108,6 +108,7 @@ function init(config, mapManager, triggerSearchBus) {
             activateBrowseTreesMode(true);
         }
     });
+    form.saveOkStream.onValue(triggerSearchBus.push);
 
     plotMarker.init(config, mapManager.map);
 


jwalgran:
Looks good. Testing this now.

jwalgran:
On this branch, the benefits and tree count are updated after adding a tree, but a new dot does not appear because new tiles are not requested. I verified that tiles are refreshed after adding a new tree on the ``master`` branch.

maurizi:
@jwalgran Are you sure that is the behaviour you are seeing?

This fix is not related to adding a new plot, only for adding a tree *to an existing plot*, via the "Quick Edit" button in the sidebar, which should not cause a new dot to appear on the map.

(Unless you mean that you don't see tiles requested for an advanced search for "Trees"?)

jwalgran:
Running backwards migrations to restore the deleted triggers that this branch expects resolved the issue.

+1

PR #2026, opened 2015-06-01
During an import, update geo_rev for each batch of trees imported

RickMohr:


RickMohr:
This is a straightforward change and I tested it with a large import, so I think it's okay not to test.

jwalgran:
+1

PR #2032, opened 2015-06-03
Increase maxZoom to 21

steventlamb:
Originally, turning the zoom level up to 21 caused bugs on the satellite
layer when zooming in on areas that don't support such high
levels. Fixes in the google plugin for leaflet made it possible to go up
to 21 universally and it will disallow levels that aren't supported.

### DIFF
--- a/opentreemap/treemap/js/shim/leaflet.google.js
+++ b/opentreemap/treemap/js/shim/leaflet.google.js
@@ -1,7 +1,8 @@
 /*
  * Google layer using Google Maps API
  */
-//(function (google, L) {
+
+/* global google: true */
 
 L.Google = L.Class.extend({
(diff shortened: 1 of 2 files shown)


steventlamb:
@RickMohr updated to use the new solution

RickMohr:
Can you update your PR comment?

RickMohr:
+1 Verified that we can zoom Tokyo imagery to zoom level 21 but Los Angeles won't go past 20.

Whew.

RickMohr:
There is a funny animation when plots are visible and you try to go from 20 to 21 -- the plots zoom but then snap back. This is not a big problem and would be difficult to fix so I don't think we should worry about it.

Issue #2030, opened 2015-06-02
Importer: "Cancel Import" doesn't work


RickMohr:
Seen on the development server:

1) Create an instance centered on Edmonton
2) Do a tree import of \\fileshare\users\rmohr\notes\otm\importer\trees10k.csv
3) Once it starts, click "View" and "Cancel Import"

Expected: import is canceled.
Actual: import continues.

jwalgran:
This may be a duplicate of https://github.com/OpenTreeMap/otm-core/issues/1980

RickMohr:
Let's fix it.

PR #2034, opened 2015-06-05
Restore ability to cancel imports

RickMohr:
I'm not sure how this ever worked. Notes:
* `Chord` returns an `AsyncResult` (not documented)
* Use `AsyncResult.parent` to get a `GroupResult`
* Use `save()` and `restore()` on the `GroupResult` as suggested in http://stackoverflow.com/questions/13685344/retrieving-groupresult-from-taskset-id-in-celery

Connects #2030


RickMohr:
waffle

RickMohr:
re-waffle

maurizi:
Looking at this now

maurizi:
+1, tested

Issue #2020, opened 2015-05-28
Regression: Tooltips not shown explaining why buttons are disabled


RickMohr:
Go to https://www.opentreemap.org/treeocodeniagara/map/?z=18/43.18697/-79.23359 and click on a tree. The "Edit Details" button is disabled, but the provided reason ("Editing a tree's details is not available to all users") is not shown on hover.

If you then click "View All Details", the "Edit" and "Delete" buttons have the same problem.

PR #2040, opened 2015-06-08
Wrap disabled buttons in a titled label element

steventlamb:
connects #2020 on github

Title (help) text is not shown for disabled buttons. However, we want to
explain why buttons are disabled so the user can take action to enable
them.

The recommended fix for twitter bootstrap tooltips is to wrap disabled
buttons in a div or label. In this case, we are using this technique,
but not activating the bootstrap tooltip. This gives us native title
text instead of tooltip windows. This is advantageous because running
$container.tooltip() is expensive, and with how our code is currently structured,
would have to be done on each map_feature click. The downside is that
bootstrap tooltips look prettier than native title text. This change
could me made easily at a later time.

This commit breaks styling in at least two known places:
* The counts/export/add titlebar is misaligned. Unchecking 'display:
  inline-block' on the label fixes this, but breaks the newly added title
  text.
* The 'Add a Tree' option in the topmost navbar is broken. @jfrankl helped
  me identify that this is because we use '>' selectors in the CSS, which
  don't respond flexibly to heirarchy changes.

### DIFF
--- a/opentreemap/treemap/js/src/buttonEnabler.js
+++ b/opentreemap/treemap/js/src/buttonEnabler.js
@@ -15,12 +15,15 @@
 
 var $ = require('jquery'),
     _ = require('lodash'),
+    format = require('util').format,
 
     enablePermAttr = 'data-always-enable',
     disabledTitleAttr = 'data-disabled-title',
     redirectUrlAttr = 'data-redirect-url',
     hrefAttr = 'data-href',
     enablePermSelector = '[' + enablePermAttr + ']',
+    disabledButtonWrapperAttrPair = 'data-class="disabled-button-wrapper"',
+    disabledButtonWrapperSelector = '[' + disabledButtonWrapperAttrPair + ']',
     config;
 
 function removeActionableDataAttributes($el) {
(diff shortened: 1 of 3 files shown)


opentreemap/treemap/js/src/buttonEnabler.js
```diff
@@ -52,9 +55,15 @@ function fullyEnable($el, href) {
 }
 
 function fullyDisable($el, disabledTitle) {
+    var wrapperTemplate =
+            '<label ' + disabledButtonWrapperAttrPair +
+            ' data-placement="bottom" title="%s"></label>';
```
maurizi: I don't think we need `data-placement="bottom"`.  Is that a vestigial Bootstrap thing?

steventlamb: Ah yeah, when I pulled out the bootstrap portion I left that there. Will remove.

dboyer:
@ctaylo37 I think we have some OTM set aside in your sprint. If so, we could use some styling on this before we merge it.

ctaylo37:
Sounds good.  I will take a look.

RickMohr:
@ctaylo37, our sprint ends tomorrow. Any chance of getting to this soon?

ctaylo37:
I'm pretty swamped with Model My Watershed, but I will try to take a look this afternoon.

ctaylo37:
I ran into a provisioning issue and couldn't get to this.  Come find me when someone has time to help out.

steventlamb:
this is ready for final review by a dev, thanks for the design fix @ctaylo37

maurizi:
Looking at this now

maurizi:
Looks good.  I found one (probably unrelated) bug, which I made an issue for (#2103).

+1

PR #2045, opened 2015-06-09
Add backend support for changing mobile api fields

maurizi:


maurizi:
Something I just realized: We should change the UDF deletion code to remove that UDF from the `mobile_api_fields`, or that instance will become unsaveable, due to having a non-existent UDF in it's mobile_api_fields` config.

jwalgran:
Yes, deleting a field does need to clean up the mobile config.

jwalgran:
Adding the model name to each field group looks redundant, because all the field names within the group have the same value as a prefix. If this makes things simpler on the UI side (or has some other benefit), it would be good to expand the commit message and/or add a comment, explaining to a future refactorer why this duplication is helpful.

maurizi:
@jwalgran Adding the model name to the config helps on the UI side because it lets you know what fields can be added to that group *when there are no fields in the group*.  I couldn't think of any other way to handle an empty field group.

jwalgran:
In the production database, there are instances with mobile field groups that have both ``field_keys`` and ``collection_udf_keys`` entries, which results in a crash because of if/elif order

```python
    for group in field_groups:
        if 'field_keys' in group:
            group['disabled_fields'] = get_disabled_fields(group)
        elif 'collection_udf_keys' in group:
            group['disabled_fields'] = get_disabled_cudfs(group)
```

Here is an example of a problematic mobile field config

```
[{u'field_keys': [u'tree.species',
   u'tree.diameter',
   u'tree.height',
   u'tree.date_planted',
   u'tree.udf:Assessment'],
  u'header': u'Tree Information',
  u'model': u'tree'},
 {u'field_keys': [u'plot.width', u'plot.length'],
  u'header': u'Planting Site Information',
  u'model': u'plot'},
 {u'collection_udf_keys': [u'plot.udf:Stewardship', u'tree.udf:Stewardship'],
  u'field_keys': [],
  u'header': u'Stewardship',
  u'sort_key': u'Date'}]
```

The cases I found were from old instances, and ``field_keys`` was always an empty array.

Flipping the if/elif would prevent the crash, but altering the migration to clean up these cases may be the better solution.

maurizi:
@jwalgran This has been updated a bit.

I've rebased against master, uncommented the mobile fields validation and unskipped the unit tests, added code to UDF deletion to also remove references from `mobile_api_fields`, and updated the data migration to fix incorrect configs

jwalgran:
+1 This works, but the migration "removes" fields that are currently available in the mobile apps, which must have relaxed case sensitivity. We will need to notify affected customers, or restore the field visibility for them.

Issue #2019, opened 2015-05-28
Uncaught JavaScript exception 


RickMohr:
Open the chrome developer tools and load https://www.opentreemap.org/treeocodeniagara/map/ -- we get:

```
Uncaught QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of '__/treeocodeniagara/species/__data' exceeded the quota.
```

We should at least catch the exception so it doesn't prevent other code from executing.

maurizi:
This is a duplicate of https://github.com/OpenTreeMap/OTM2/issues/1857

Issue #1857, opened 2014-12-10
In an instance with all OTM species the species typeahead is too big for browser local storage


RickMohr:
I made a treemap outside the US. Since it has no i-Tree region we initialize it with all known species.

Loading any page with a species typeahead gives this JS error (if you have the developer tools open):

    Uncaught QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of '__/edmonton2/species/__data' exceeded the quota.

The typeahead still works fine.

@maurizi suggests that we stop using browser-local storage for the typeahead species data and instead include it as a script tag on the page.

RickMohr:
If we're not going to fix this in the short term we should at least catch the exception so that other code on the page still runs.

maurizi:
We will also fix https://github.com/OpenTreeMap/otm-core/issues/199 as part of this.

Issue #199, opened 2013-08-15
Add support for caching species and boundaries list


maurizi:
At the minimum we can add an etag hash.

We can also add a hash in the config.js and have typeahead.js use HTML5's localStorage to cache if the config's hash matches the copy in localStorage.

PR #2049, opened 2015-06-10
Cache species and boundary autocomplete data

RickMohr:
See https://github.com/twitter/typeahead.js/blob/master/doc/bloodhound.md
for a description of cache parameters discussed here.

We were re-fetching an instance's species and boundary data on every page load
because we were using `ttl` ("time to live") of 0.
Fix is to use a default `ttl` of one day, and a `thumbprint` (cache buster) based on
the latest update time of the appropriate table (Species or Boundary).

Also, visiting too many instances was causing a "quota exceeded" error for browser local storage.
Fix is to use an instance-independent `cacheKey` ("species" or "boundary").
That means switching instances invalidates the cache, but that's rare compared to
visiting many consecutive pages on the same instance.

Connects #1857
Connects #2019
Connects #199

To test:

0) Clear your browser local storage
* window.localStorage.clear()
* window.localStorage

1) Verify that species and boundary autocomplete data is cached
* Open developer tools, network tab
* Load a treemap
* Note that `/instance/species/` and `/instance/boundaries/` are fetched.
  (The species are fetched twice. Hard to avoid.)
* Go to a plot detail page. Note that species and boundaries are not re-fetched.

2) Verify that loading a different treemap invalidates the cache
* Load a different treemap
* Note that the search dropdowns contain different species and boundaries

3) Verify that changing data invalidates the appropriate cache
* Start a Django shell
* `from treemap.models import Species, Boundary`
* `s = Species.objects.get(instance__url_name='latreemap', common_name='Afghan pine')`
* `s.common_name='Afghan pony'`
* `s.save_with_system_user_bypass_auth()`
* `s = Boundary.objects.get(name='Alhambra')`
* `s.name = 'Alhamburger'`
* `s.save()`
* Reload the appropriate treemap
* See your changes in the dropdowns

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -439,6 +439,44 @@ class Instance(models.Model):
 
         return names
 
+    @property
+    def species_thumbprint(self):
+        # Species autocomplete data lives in browser local storage.
+        # It must be invalidated when a different instance is loaded,
+        # or when the current instance's species are updated.
+        #
+        # To get a unique thumbprint across instances and species updates
+        # we use the instance's latest species update time if available,
+        # and otherwise its url name.
+        from treemap.models import Species
+        thumbprint = None
+        my_species = Species.objects \
+            .filter(instance_id=self.id) \
+            .order_by('-updated_at')
+        try:
+            thumbprint = my_species[0].updated_at
+        except IndexError:
+            pass
+        return "%s_%s" % (self.url_name, thumbprint)
+
+    @property
+    def boundary_thumbprint(self):
+        # Boundary autocomplete data lives in browser local storage.
+        # It must be invalidated when a different instance is loaded,
+        # or when the current instance's species are updated.
+        #
+        # To get a unique thumbprint across instances and boundary updates
+        # we use the latest boundary update time if available,
+        # and otherwise the instance's url name.
+        from treemap.models import Boundary
+        thumbprint = None
+        my_boundaries = Boundary.objects.order_by('-updated_at')
+        try:
+            thumbprint = my_boundaries[0].updated_at
+        except IndexError:
+            pass
+        return "%s_%s" % (self.url_name, thumbprint)
+
     def update_geo_rev(self):
         qs = Instance.objects.filter(pk=self.id)
 
--- a/opentreemap/treemap/templatetags/species_thumbprint.py
+++ b/opentreemap/treemap/templatetags/species_thumbprint.py
@@ -1,0 +1,25 @@
+from django import template
+
+import threading
+
+register = template.Library()
+
+_thread_local = threading.local()
+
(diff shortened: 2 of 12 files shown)


maurizi:
Looks like there is some lint in this PR (according to TravisCI at least)

opentreemap/treemap/templatetags/species_thumbprint.py
```diff
@@ -0,0 +1,23 @@
+from django import template
+
+register = template.Library()
+
+last_request_and_thumbprint = (None, None)
+
+# Some pages contain many species autocompletes. To avoid re-querying
+# we hang on to the last request and thumbprint. This isn't thread-safe,
+# but the the worst that could happen with a data race is an extra query.
```
maurizi: I _think_ your concerns about thread-safety are unlikely, because AFAIK request objects are never shared between threads.  Like you say though, a data race here isn't so bad.

opentreemap/treemap/instance.py
```diff
@@ -439,6 +439,44 @@ def static_page_names(self):
...
+        try:
+            thumbprint = my_species[0].updated_at
+        except IndexError:
+            pass
+        return thumbprint or self.url_name
+
+    @property
+    def boundary_thumbprint(self):
+        # Boundary autocomplete data lives in browser local storage.
+        # It must be invalidated when a different instance is loaded,
```
maurizi: > It must be invalidated when a different instance is loaded

I don't see any code tying this thumbprint to a particular instance.  How does it get invalidated when switching instances?

opentreemap/treemap/templatetags/species_thumbprint.py
RickMohr: I was thinking about two threads making interleaved calls to `species_thumbprint`. Neither would get the wrong answer, but they would keep invalidating each other's cache.

This might solve it:

```python
import threading

_thread_local = threading.local()

def species_thumbprint(request):
    last_request, last_thumbprint = _thread_local.getattr(
        last_request_and_thumbprint, (None, None))

    if request == last_request:
        thumbprint = last_thumbprint
    else:
        thumbprint = request.instance.species_thumbprint
        _thread_local.last_request_and_thumbprint = (request, thumbprint)

    return thumbprint
```

opentreemap/treemap/instance.py
RickMohr: Good point. It worked when there were no boundary changes, using `self.url_name`. Updated to always prefix thumbprint with `self.url_name`.

RickMohr:
@maurizi, all comments addressed.

maurizi:
@RickMohr Looks like your lint still isn't fixed

maurizi:
+1, once the lint is fixed.  Works well.

Issue #2051, opened 2015-06-10
Release: Photo not accessible to Pintrest sharing when using S3 storage


jwalgran:
To reproduce, add a photo to a tree and click the Pintrest share button in the modal.

![screen shot 2015-06-10 at 9 41 47 am](https://cloud.githubusercontent.com/assets/17363/8088314/37b05644-0f55-11e5-8cd2-8a2d22847835.png)

Requesting the image link returns an XML error. 

```xml
<Error>
  <Code>AccessDenied</Code>
  <Message>
    Query-string authentication requires the
    Signature, Expires and AWSAccessKeyId parameters
  </Message>
  <RequestId>A99CC1BEEAF99DE6</RequestId>
  <HostId>
    Xe2U9jeV0sZhB3tJ8p2OMBJHUosrD0o6QSjFMBGBDWafkl3u3nj5Q3jkYDipK/pOGsO8nYBHTk8=
  </HostId>
</Error>

RickMohr:
Sharing makes no sense for private instances. We don't show sharing buttons for trees, but currently we do for images. We shouldn't.

jwalgran:
That is true, but I am going to open a separate issue for it.

I produced the error condition described in this issue using an internet-facing, public instance.

steventlamb:
chopping the `?signature=<...>` portion off the S3 URL fixes the issue. Looking at how to accomplish that programatically now.

PR #2054, opened 2015-06-10
Don't send any url params to photo sharing s3 urls

steventlamb:
connects #2051 on github


maurizi:
+1, didn't test

jwalgran:
I confirmed that an S3 photo is now accessible to Pintrest

Issue #2039, opened 2015-06-06
Header not working on Mobile?


collinanderson:
Hi All,

I just get a blank screen. No errors in the console. We're using the vagrant version. Any ideas?

Thanks,
Collin

maurizi:
We don't have 100% support for mobile browsers, but you shouldn't be seeing a blank screen.

What phone/browser are you using?

collinanderson:
Both Nexus 4 and iPhone 5. It's not 100% blank. It just doesn't have the header.

collinanderson:
It looks like it just doesn't load the header, so there's no way to login.

![image](https://cloud.githubusercontent.com/assets/129234/8021015/df412758-0c5d-11e5-86f7-04faa113b1a5.png)

maurizi:
Thanks for reporting this!

I've made a fix for this issue in https://github.com/OpenTreeMap/otm-core/pull/2055/.

PR #2055, opened 2015-06-10
Fix styling error causing header to not appear on smaller screen sizes.

maurizi:
Connects to #2039


collinanderson:
works for me. Thanks!

maurizi:
@ctaylo37 I don't *think* the `collapse` class was necessary here, but I'd appreciate your input.

maurizi:
Got a verbal confirmation from @ctaylo37 that this fix should be OK.

PR #2056, opened 2015-06-11
Add initial app/schema for stormwater features

steventlamb:


RickMohr:
Looks like there are a few broken tests.

opentreemap/treemap/audit.py
```diff
@@ -140,13 +140,13 @@ def approve_or_reject_audits_and_apply(audits, user, approved):
...
-    from treemap.models import MapFeature
+    from treemap.models import MapFeaturePhoto
     if roles is None:
         roles = Role.objects.filter(instance=instance)
     if models is None:
-        # MapFeature is "Authorizable", but it is effectively abstract
-        # Only it's subclasses should have permissions added
-        models = all_subclasses(Authorizable) - {MapFeature, PendingAuditable}
+        # MapFeaturePhoto is an exception to the usual pattern - it is
+        # the only class that we use directly *and* inherit.
```
RickMohr: Much better. How about this rewritten comment:

We need permissions only on those subclasses of `Authorizable` which we instantiate. Those are the leaf nodes of the subclass tree, plus `MapFeaturePhoto` (which has subclass `TreePhoto`).

steventlamb:
@RickMohr updated

RickMohr:
+1 thanks.

PR #2057, opened 2015-06-11
Temporarily disable mobile api field validation

maurizi:


jwalgran:
+1 my non-savable instances are now savable

PR #2059, opened 2015-06-12
Feature/2036/add bioswale to map

steventlamb:


maurizi:
I was unable to view a popup for a bioswale, or see it in the sidebar.  That doesn't have to be fixed in this PR, but we should open an issue for it.

Otherwise, +1, works well.

steventlamb:
https://github.com/OpenTreeMap/otm-core/issues/2062 created

PR #2063, opened 2015-06-16
Print debug-friendly repr of APIAccessCredential

steventlamb:


jwalgran:
+1 I like it (not manually tested)

PR #2064, opened 2015-06-16
Fix issue preventing global errors from being shown.

maurizi:
We were correctly writing non-field errorss to a global error message
section, but then accidentally overwriting those global errors when
attempting to show inline field errors.

Excluding the global error section when clearing the field errors fixes
this bug.

### DIFF
--- a/opentreemap/treemap/js/src/inlineEditForm.js
+++ b/opentreemap/treemap/js/src/inlineEditForm.js
@@ -253,7 +253,7 @@ exports.init = function(options) {
         },
 
         showValidationErrorsInline = function (errors) {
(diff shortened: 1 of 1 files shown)


steventlamb:
It looks like a similar issue can be found on the plot detail page when putting bogus data in the tree height field of the plot/tree edit page. Assigning back to you for review, as per our verbal discussion.

steventlamb:
+1 otherwise, code behaves as expected

maurizi:
@steventlamb That appears to be unrelated, so I am opening a separate issue to fix the bug you found.

Issue #2061, opened 2015-06-15
Performance: don't request "boundary" tiles if none will be displayed


RickMohr:
We often request boundary tiles when no boundaries will be displayed.

Fixes:
* Don't create boundary layer if map intersects no boundaries (`len(instance.boundaries) == 0`)
* Put min zoom 15 on the boundary layer

Issue #2037, opened 2015-06-05
Add polygonal map feature layer to the map page


jwalgran:
Depends on https://github.com/OpenTreeMap/otm-tiler/issues/60

If the instance does not have polygonal map feature types, do not bother adding this layer to the map.

Coordinate min zoom level with the CartoCSS

RickMohr:
Since this layer will only be shown at higher zoom levels and probably won't have many features we could consider using a GeoJSON layer instead of a tiled layer. Advantage -- fewer HTTP requests, faster load. Disadvantage -- requires custom implementation for mobile apps.

maurizi:
@RickMohr I can't speak for how things are for iOS, but it looks like it's straightforward  to add a polygon to a Google Map on android (https://developers.google.com/android/reference/com/google/android/gms/maps/model/Polygon), but we'd have to write our own GeoJSON parser (thats not *too* hard though).

jwalgran:
Creating tappable, polygonal overlays with iOS MapKit is possible.

https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKPolyline_class/index.html#//apple_ref/occ/cl/MKPolyline

http://stackoverflow.com/questions/20858108/detecting-touches-on-mkoverlay-in-ios7-mkoverlayrenderer

http://stackoverflow.com/questions/4354130/how-to-determine-if-an-annotation-is-inside-of-mkpolygonview-ios

https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/CGPath/index.html#//apple_ref/c/func/CGPathContainsPoint

http://stackoverflow.com/questions/19014926/detecting-a-point-in-a-mkpolygon-broke-with-ios7-cgpathcontainspoint

https://github.com/mattt/GeoJSONSerialization/

jwalgran:
Working with vectors will requires performance tweaks sooner that raster tiles. Performance issues appear with thousands of vector features (hundreds on older IEs). It generally takes 10s of thousands of features before performance issues appear in raster layers.

In TreesCount we saw, when attempting to encode all the geometry in the UTF grid, that ST_AsGeoJSON start dominating the feature query time as the number of features increased.

Using an alternate renderer rules out unified styling of features. We plan to expose custom map feature styling options to map owners, so will need to restrict these options to a lowest common denominator supported by all three renderers (Leaflet, Android, iOS) and convert our style definitions to the syntax required for each of the renderers.

maurizi:
[...]

PR #2065, opened 2015-06-16
Show layer for polygonal map features

RickMohr:
Show layer only if
* instance has polygonal map feature types, and
* zoom is 15 or higher

And while we're at it, likewise show boundary layer only if
* instance has boundaries, and
* zoom is 15 or higher

Communicate `has_polygons` and `has_boundaries` via data attributes on the map DOM element, for both the main map and detail map.

To test:
0) Get OpenTreeMap/otm-tiler#63 and restart the tiler
1) Add the `Bioswale` map feature type to an instance
2) Create a bioswale -- its polygon should be displayed
3) Open network tab and zoom out to level 14. There should be no tile requests for the boundary or polygonal map feature
layers.
4) Go to an instance that doesn't have the `Bioswale` map feature type.
5) Zoom in to level 15. There should be no tile requests for the polygonal map feature layer.

Connects #2037
Connects #2061

Requires OpenTreeMap/otm-tiler#63

### DIFF
--- a/opentreemap/treemap/js/src/MapManager.js
+++ b/opentreemap/treemap/js/src/MapManager.js
@@ -204,17 +204,16 @@ function createBoundariesTileLayer(config) {
     return L.tileLayer(url, options);
 }
 
-// Leaflet uses {s} to indicate subdomains
+function getPlotLayerURL(config, extension) {
+    return getLayerURL(config, 'treemap_mapfeature', extension);
+}
+
 function getLayerURL(config, layer, extension) {
     var host = config.tileHost || '';
     return host + '/tile/' +
         config.instance.rev +
         '/database/otm/table/' + layer + '/{z}/{x}/{y}.' +
         extension + '?instance_id=' + config.instance.id;
-}
-
-function getPlotLayerURL(config, extension) {
-    return getLayerURL(config, 'treemap_mapfeature', extension);
 }
 
 function deserializeZoomLatLngAndSetOnMap(mapManager, state) {
--- a/opentreemap/treemap/js/src/MapManager.js
+++ b/opentreemap/treemap/js/src/MapManager.js
@@ -25,7 +25,7 @@ MapManager.prototype = {
         var config = options.config,
             plotLayer = createPlotTileLayer(config),
             allPlotsLayer = createPlotTileLayer(config),
-            boundsLayer = createBoundsTileLayer(config),
+            boundariesLayer = createBoundariesTileLayer(config),
             utfLayer = createPlotUTFLayer(config);
 
         this._config = config;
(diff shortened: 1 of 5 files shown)


opentreemap/treemap/js/src/MapManager.js
```diff
@@ -8,7 +8,9 @@ var $ = require('jquery'),
     makeLayerFilterable = require('treemap/makeLayerFilterable'),
     urlState = require('treemap/urlState'),
 
-    _ZOOM_OPTIONS = {maxZoom: 21};
+    MAX_ZOOM_OPTION = {maxZoom: 21},
```
steventlamb: why drop the underscore? doesn't look like these have changed from internal to exported.

maurizi: Personally I've never seen a need to prefix private JS variables with `_`.  Since only explicitly exported items can be used anyway, it isn't necessary to distinguish public/private items (unlike Python, where everything is exported).

steventlamb:
+1, tested, works

opentreemap/treemap/js/src/MapManager.js
```diff
@@ -150,25 +175,33 @@ function getBasemapLayers(config) {
...
 function createPlotTileLayer(config) {
     var url = getPlotLayerURL(config, 'png'),
-        layer = L.tileLayer(url, _ZOOM_OPTIONS);
+        layer = L.tileLayer(url, MAX_ZOOM_OPTION);
+    makeLayerFilterable(layer, url, config);
+    return layer;
+}
+
+function createPolygonTileLayer(config) {
+    var url = getPlotLayerURL(config, 'png'),
```
steventlamb: is this a typo? Do we want the plot url here, or the polygon url? This will be sent to `makeLayerFilterable`.

opentreemap/treemap/js/src/MapManager.js
RickMohr: That's exactly why I changed it.

opentreemap/treemap/js/src/MapManager.js
```diff
@@ -104,9 +123,15 @@ MapManager.prototype = {
 
         if (!this._allPlotsLayer.map) {
```
steventlamb: do you also want:
```
        if (this._hasPolygons) {
            this._polygonLayer.setFilter(filter);
        }
```
?

steventlamb: (I've added this in my working branch anyway, so no worries)

opentreemap/treemap/js/src/MapManager.js
RickMohr: Good catch. Filtering of polygon layers would not have worked. Fixed.

opentreemap/treemap/js/src/MapManager.js
RickMohr: Right. Thanks for adding this in your work.

PR #2071, opened 2015-06-16
Fix accidental mutation of default api fields.

maurizi:
We were accidentally mutating the DEFAULT_MOBILE_API_FIELDS module-level
variable when handling UDF deletion.

Depending on the order the unit tests were run in, this could cause a
test to fail, and almost certainly could have occurred in production
as well.

### DIFF
--- a/opentreemap/treemap/udf.py
+++ b/opentreemap/treemap/udf.py
@@ -584,17 +584,18 @@ class UserDefinedFieldDefinition(models.Model):
                          .filter(model='udf:%s' % self.pk)\
                          .delete()
 
-            for group in self.instance.mobile_api_fields:
-                if self.full_name in group.get('collection_udf_keys', []):
-                    # If this is the only collection UDF with this name,
-                    # we remove the entire group, since there would be no
-                    # eligible items to go *in* the group after deletion
-                    if len([udf for udf in udf_defs(self.instance)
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/udf.py
```diff
@@ -584,17 +584,18 @@ def delete(self, *args, **kwargs):
...
-                    # If this is the only collection UDF with this name,
-                    # we remove the entire group, since there would be no
-                    # eligible items to go *in* the group after deletion
-                    if len([udf for udf in udf_defs(self.instance)
-                            if udf.name == self.name]) == 1:
-                        self.instance.mobile_api_fields.remove(group)
-                    # Otherwise, just remove this UDF from the group
-                    else:
-                        group['collection_udf_keys'].remove(self.full_name)
+            if 'mobile_api_fields' in self.instance.config:
```
RickMohr: As we discussed, it would be nice to have a comment here explaining what this `if` is doing.

jwalgran:
+1

This was my test:

1. Create an instance
1. Use the Django shell to delete one of the stewardship fields and verify that it is no longer accessible in the app.
1. Create a second instance and verify that I can add a tree with both plot and tree stewardship fields, meaning the defaults were not mutated.

PR #2072, opened 2015-06-17
Send filter strings to polygon layer

steventlamb:
connects #2038 on github

requires https://github.com/OpenTreeMap/otm-tiler/pull/65 for testing

### DIFF
--- a/opentreemap/treemap/js/src/MapManager.js
+++ b/opentreemap/treemap/js/src/MapManager.js
@@ -121,6 +121,10 @@ MapManager.prototype = {
     setFilter: function (filter) {
         this._plotLayer.setFilter(filter);
 
+        if (this._hasPolygons) {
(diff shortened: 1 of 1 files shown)


RickMohr:
If I do an advanced search with "Bioswales" unchecked the blue dots are lightened but not the blue polygons.

steventlamb:
I believe this is working for me.

After searching with it unchecked:
screenshot from 2015-06-17 18 41 52
after searching with it checked:
screenshot from 2015-06-17 18 42 10

Do you also have the tiler branch checked out? Do you need to restart the tiler after checking out these branches?

RickMohr:
I did do all those things. But I did them again, and also moved a tree to reset the cache buster. Now it works. Playing around, I'm unable to reproduce the problem.

+1!

PR #2079, opened 2015-06-18
Feature/2062/bioswale popup

steventlamb:


opentreemap/stormwater/templates/stormwater/Bioswale_detail.html
```diff
@@ -0,0 +1,6 @@
+{% extends "treemap/resource_detail.html" %}
+{% load i18n %}
+{% load l10n %}
+{% load form_extras %}
```
RickMohr: I think you could eliminate these last two `load`s.

opentreemap/treemap/models.py
```diff
@@ -695,7 +695,21 @@ def cast_to_subtype(self):
...
+            # don't know how to cast along a heirarchy, but classes
+            # know how fetch subtype and a query easily gets an
+            # instance of the subtype.  This is a single row query not
+            # likely to happen as part of a list-loop, and will only
+            # happen for features that are not *direct* descendents of
+            # MapFeature.
+            return (self
+                    .__class__
+                    .get_subclass(ft)
+                    .objects.get(pk=self.id))
```
RickMohr: Alternative:
```py
elif hasattr(self, 'polygonalmapfeature'):
    pmf = getattr(self, 'polygonalmapfeature')
    return getattr(pmf, ft.lower())

RickMohr:
FYI - unfortunately waffle won't pick up your "connects" comments unless they're in the PR message.

opentreemap/treemap/models.py
steventlamb: Fix pushed. I dropped the non-exhaustive elif clause because it would have produced a `DoesNotExist` error the same as attempting to access it directly.

opentreemap/stormwater/templates/stormwater/Bioswale_detail.html
steventlamb: fixed.

RickMohr:
+1, works great.

PR #2081, opened 2015-06-18
Initial first draft of public API documentation

jwalgran:
Building comprehensive documentation of an API is a large task. This
initial README is not complete, but does mention all of the endpoints used
by the mobile applications. There are some endpoints in api/urls.py that
are not covered in this README that are either broken or out of date and
not used.

The most glaring omission in this initial pass is the lack of a table
describing the datatypes and allowable values for all of the input
parameters and output data.

The Markdown is structured in a format that is compatible with
https://github.com/tripit/slate, which produces output similar to the Stripe
documentation, which I feel is some of the best API documentation
available. https://stripe.com/docs/api

I do not know if adding the Ruby dependencies of tripit/slate is good
for the project. We can devise our own python-based script to build
hostable HTML, if we choose to.

### DIFF
--- a/opentreemap/api/README.md
+++ b/opentreemap/api/README.md
@@ -1,0 +1,1015 @@
+**NOTE: This is DRAFT documentation and is subject to change.**
+
+# Introduction
+
+TODO
+
+# Terminology
+
+## Instance
+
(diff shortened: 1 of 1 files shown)


jwalgran:
Will reopen to connect to the related issue.

PR #2082, opened 2015-06-18
Initial first draft of public API documentation

jwalgran:
Building comprehensive documentation of an API is a large task. This
initial README is not complete, but does mention all of the endpoints used
by the mobile applications. There are some endpoints in api/urls.py that
are not covered in this README that are either broken or out of date and
not used.

The most glaring omission in this initial pass is the lack of a table
describing the datatypes and allowable values for all of the input
parameters and output data.

The Markdown is structured in a format that is compatible with
https://github.com/tripit/slate, which produces output similar to the Stripe
documentation, which I feel is some of the best API documentation
available. https://stripe.com/docs/api

I do not know if adding the Ruby dependencies of tripit/slate is good
for the project. We can devise our own python-based script to build
hostable HTML, if we choose to.

Connects to #2033

### DIFF
--- a/opentreemap/api/README.md
+++ b/opentreemap/api/README.md
@@ -1,0 +1,1015 @@
+**NOTE: This is DRAFT documentation and is subject to change.**
+
+# Introduction
+
+TODO
+
+# Terminology
+
+## Instance
+
+A single installation of OpenTreeMap supports an unlimited number of
+"maps." To avoid confusion with actual map displayed on the explore
+page, the term "instance" is used internally, and in the API.
+
+## Plot
+
+OpenTreeMap tracks the site at which a tree is planted separately from
+the tree itself. For street trees, it is not uncommon for a damaged
+tree to be removed, and a new tree added in its place. The web
+application uses the more general "planting site" to refer to the
+location where trees are planted. Internally, and in the API, "plot"
+is used to refer to these locations.
+
+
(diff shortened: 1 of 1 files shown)


opentreemap/api/README.md
```diff
@@ -0,0 +1,1015 @@
...
+# Introduction
+
+TODO
+
+# Terminology
+
+## Instance
+
+A single installation of OpenTreeMap supports an unlimited number of
+"maps." To avoid confusion with actual map displayed on the explore
```
RickMohr: "with actual maps"

RickMohr: I'll just fix things like this and make a PR to your PR.

opentreemap/api/README.md
```diff
@@ -0,0 +1,1015 @@
...
+Definition:
+
+```
+GET /api/{version}/user/
+```
+
+Example Request:
+
+```
+curl "https://opentreemap.org/api/v3/user?access_key=AN_ACCESS_KEY&timestamp=2015-06-16T17%3A59%3A37&signature=Ybtw...="
```
RickMohr: Did you mean for the signature to end with `=`?

RickMohr:
+1, great to have this well-structured doc.

(See https://github.com/jwalgran/otm-core/pull/2)

PR #2107, opened 2015-06-24
New modeling toolbar

jwalgran:
The new design for the modeling application unifies prioritization and scenarios into a "plan." We no longer need a drop down menu to access the different modeling modes because they we will each be accessible from the same single-page interface.

The design on the modeling application calls for a left-aligned toolbar in the header. The "search" sections of the header are all right-aligned, necessitating the creation of a new "toolbar" block.


RickMohr:
+1

Issue #2089, opened 2015-06-18
Bioswale polygon remains on map after zooming out past the minZoom


maurizi:
We encountered a similar issue on TreesCount, we worked around it here: https://github.com/azavea/nyc-trees/blob/develop/src/nyc_trees/js/src/progressPage.js#L192-L205

PR #2108, opened 2015-06-24
Work around Leaflet afterimage issue.

maurizi:
There is a known Leaflet issue which causes an "afterimage" when going one
zoom level beyond a layer's min/max zoom level.
(https://github.com/Leaflet/Leaflet/issues/1905)

This fix works around the issue.  We should be able to remove this once
we upgrade to Leaflet 1.0, once it is out.

Connects to #2089


jwalgran:
+1

I verified that after zooming from 15 out to 14 the polygonal map feature layer was cleaned up.

Issue #2099, opened 2015-06-18
Add ability to add stewardship entries from the bioswale detail page


maurizi:

dboyer:
Includes making the resource detail page editable.

PR #2109, opened 2015-06-24
Feature/resource edit page

steventlamb:
connects #2099 on github


steventlamb:
this has broken ui tests from missing dom element name updates. fix coming soon.

jwalgran:
+1

- I was able to add stewardship to Bioswales when logged in as the map owner. The edit buttons appeared in both the map popups and on the detail page.
- Editing was disabled when logged in as a non-owner.
- Stormwater features from other apps remained non-editable.

Issue #2069, opened 2015-06-16
Return JSON error messages when there is an unhandled exception in an API request


jwalgran:
The API sometimes returns HTML content when there is an error.

maurizi:
@jwalgran does this bug include fixing this: https://github.com/OpenTreeMap/otm-core/blob/master/opentreemap/api/user.py#L103-L106?

Looks like we would need to make a few changes in the iOS application as well.

jwalgran:
I don't think addressing that is a requirement for this issue. Returning a structured JSON body would require bumping the API version and doing a "translation" so that older app versions do not break.

PR #2111, opened 2015-06-26
Return JSON for unhandled errors to API views.

maurizi:
I investigated changing our explicit error responses from API views to be JSON instead of plain text strings, but since the existing mobile apps use the textual responses in several places to report errors to the user, I opted to leave them as is.

To test, you will have to change your settings in Django settings to set `DEBUG=False` and `ALLOWED_HOSTS=['localhost']`.

You should see that any URL starting with `/api/` should return JSON for an unhandled error like a 404 or a 500 error, and that other URLs continue to return our custom error pages.

Connects to #2069

### DIFF
--- a/opentreemap/treemap/views/misc.py
+++ b/opentreemap/treemap/views/misc.py
@@ -6,13 +6,15 @@ from __future__ import division
 import string
 import re
 import sass
+import json
 
 from django.utils.translation import ugettext as _
 from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.http import HttpResponse, HttpResponseRedirect
-from django.shortcuts import get_object_or_404
+from django.shortcuts import get_object_or_404, render_to_response
+from django.template import RequestContext
 
 from stormwater.models import PolygonalMapFeature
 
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/views/misc.py
```diff
@@ -227,3 +229,20 @@ def instance_geojson(instance):
                  }))
 
     return [instance_geojson(instance) for instance in instances]
+
+
+def error_page(template, status_code):
```
steventlamb: consider
```python
def error_page(status_code):
    template = '%s.html' % stauts_code
    def inner_fn(request):
        ...
```

opentreemap/treemap/views/misc.py
```diff
@@ -227,3 +229,20 @@ def instance_geojson(instance):
...
+    def inner_fn(request):
+        # API requests with an unhandled error should return JSON, not HTML
+        if request.path.startswith('/api/'):
```
steventlamb: would it make more sense to gate on `Accept: application/json` ?

opentreemap/treemap/views/misc.py
maurizi: I do like that better, I think I'll make that change.

opentreemap/treemap/views/misc.py
maurizi: I think we should _always_ return JSON for API requests.  That said, returning JSON errors for _other_ requests that are expecting a JSON response is almost certainly a good idea, will do.

maurizi:
@steventlamb I made the 2 fixes you suggested.

opentreemap/treemap/views/misc.py
```diff
@@ -227,3 +229,23 @@ def instance_geojson(instance):
...
+
+def error_page(status_code):
+    template = '%s.html' % status_code
+
+    def inner_fn(request):
+        # API requests with an unhandled error should return JSON, not HTML
+        if ((request.path.startswith('/api/') or
+             'application/json' in request.META.get('HTTP_ACCEPT'))):
+            response = HttpResponse(json.dumps(
+                {'status': 'Failure', 'reason': 'An unhandled error occured'}),
```
jwalgran: Saying there was an "unhanded error" in the 404 case is incorrect. I would suggest checking ``status_code`` to decide what the value of ``reason`` should be.

maurizi: Done.

steventlamb:
+1, tested

PR #2115, opened 2015-06-29
Add a settings switch for etag caching

steventlamb:
I've found it useful to add this feature many times during development


jwalgran:
Abstractly, I agree that having a switch to turn off caching in development is a good idea. Can you share a specific use case where this smoothed out the workflow? I have never felt the need to do it before.

I have some thoughts on the name of the setting

* As I understand it, the ``etag`` decorator generates simply generates a header. Including ``_CACHING`` may be subtly misleading, since Django is not actually caching anything.

* The name does not capture the dependency on debug mode. If I have ``DEBUG = False`` in my settings I might still expect setting ``ENABLE_ETAG_CACHING = False`` to disable generating etag headers. ``DEBUG_ETAG_ENABLED = False`` may clarify that this will only work in debug mode.

jwalgran:
I verified this on the tree detail page. When I click edit, then click save, I see an AJAX request for updated sidebar content. When the setting is ``True``, this sidebar response includes an etag header. When the setting is ``False`` the etag header is not returned with the response.

steventlamb:
The use case that comes up for me repleatedly is debugging things related to cached ecobenefit calculations, like mapfeature popups, and model counts. I'll attempt to change something in a template and won't see the result until I remember etag caching is on and I clear my cache.

steventlamb:
re: naming - good points.

First, happy to drop `_CACHING`, that makes sense.

As for the dependency on debug mode, I threw that in there as a silent safeguard to make sure that this is a development opt-in, not something that can accidentally be switched on in production. I feel like this is a common convention, for example, when you switch `DEBUG` to `False`, you will then need to set `ALLOWED_HOSTS`, even though you may not have realized that the server was taking care of that detail for you. That was my reasoning. I don't have strong feelings about silently requiring `DEBUG`, so I'm happy to make it more explicit. However, I find `DEBUG_ETAG_ENABLED = False` to be a tricky thing to decipher quickly, it might imply that etags are disabled by default when in DEBUG, and this is required to enable them.

steventlamb:
@maurizi mentioned that this can be solved on the browser side instead:
screenshot from 2015-07-07 12 16 43

PR #2116, opened 2015-06-29
Feature/2100/generic udfc search

steventlamb:


opentreemap/treemap/udf.py
```diff
@@ -1008,6 +1006,17 @@ def audits(self):
         all_audits = udf_collection_audits | regular_audits
         return Audit.objects.filter(all_audits).order_by('created')
 
+    def search_slug(self):
+        return to_object_name(self.__class__.__name__)
+
+    @classproperty
+    def search_display_name(self):
+        "this should be overridden"
```
maurizi: If this really _needs_ to be overridden, consider having it raise an exception when called, instead of trying to calculate a reasonable default.

maurizi:
+1, works well.

PR #2119, opened 2015-06-30
Clarify empty csv error text

steventlamb:
@dboyer @maurizi what do you think of this, as a holdover? Looks like it can't hurt.


maurizi:
+1 :deciduous_tree:

PR #2132, opened 2015-07-06
Upgrade dragula JS dependency

jwalgran:
This version include a fix to a dependency issue that was breaking the bundling of our javascript assets.

https://github.com/bevacqua/dragula/commit/d7342ace10b65dbe45be109139f1d7162e7e3efe


steventlamb:
+1 tested

PR #2133, opened 2015-07-06
Don't require more data attributes than necessary

steventlamb:
connects #2129 on github

They dynamically generated udfFieldDefId for the given modelName is
enough to build the name.


RickMohr:
+1

PR #2134, opened 2015-07-06
Hide point marker on PolygonalMapFeature detail

steventlamb:
connects #2130 on github

The TODO in here is something of a discussion point. Should this be done? It might involve far-reaching API changes.

### DIFF
--- a/opentreemap/treemap/js/src/mapFeature.js
+++ b/opentreemap/treemap/js/src/mapFeature.js
@@ -49,13 +49,15 @@ exports.init = function(options) {
         $(window).asEventStream('popstate')
             .map(function() { return U.getLastUrlSegment() === 'edit'; }));
 
+    var currentPlotMover;
+
     var form = inlineEditForm.init(
             _.extend(options.inlineEditForm,
                      { config: options.config,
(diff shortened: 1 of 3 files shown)


jwalgran:
What do you think about keeping the details of the ``plotMover`` interface inside of ``plotMover.js``?

mapFeature.js

```javascript
if (options.hidePointMarker) {
    currentPlotMover = plotMover.disabled();
} else {
    ...
    currentPlotMover = plotMover.init({
        mapManager: mapManager,
        plotMarker: plotMarker,
        inlineEditForm: form,
        editLocationButton: options.location.edit,
        cancelEditLocationButton: options.location.cancel,
        location: options.location.point
    });
    ...
}

function onSaveBefore(data) {
    currentPlotMover.onSaveBefore(data);
}
```

plotMover.js

```javascript
exports.disabled = function() {
    return {
        onSaveBefore: _.noop
    };
};

exports.init = function ...
```

jwalgran:
I tested and confirmed that a marker does not appear on the bioswale detail page, but still appears on the  tree detail page. I am ok with the current version,  or any of the variations discussed.

+1

PR #2136, opened 2015-07-08
Styling tweaks for scenario actions

RickMohr:


jwalgran:
+1 I like using obviously wrong colors to signal that some design attention is needed.

PR #2137, opened 2015-07-08
Feature/2126/bioswale delete

steventlamb:


maurizi:
+1, tested, looks good

Issue #2073, opened 2015-06-17
The "share" section of a plot returned after creating a plot via the API has API-specific URLs


jwalgran:
A request like this

```
curl -H "Content-Type: application/json"\
     -X POST\
     -d '{"plot": {"geom": {"srid": 4326, "x": -75.15653, "y": 39.97426}, "width": 10}, "tree": {"height": 10}}'\
     "https://opentreemap.org/api/v3/instance/myinstance/plots/?access_key=AN_ACCESS_KEY&timestamp=2015-06-16T19%3A48%3A05&signature=ybtw..."
```

Returns plot details that include a "share" section like this

```
  "share": {
    "url": "https:\/\/opentreemap.org\/api\/v3\/instance\/myinstance\/plots?access_key=LIoY8sU8SvCDdIRwmaL0fg",
    "image": "https:\/\/opentreemap.org\/static\/img\/tree.png",
    "description": "This Missing Species is mapped on My Instance",
    "title": "Missing Species on My Instance"
  }
```

The ``url`` value should point to the web app, not the API.

PR #2140, opened 2015-07-09
Fix share link

steventlamb:
connects #2073 on github

the share context was using the same url used to produce the feature,
which is incorrect for the API. Here we explicitly use the url to the
map_feature_detail page in the webapp.


jwalgran:
+1

I see a proper absolute URL in the API response

screen shot 2015-07-09 at 2 19 19 pm

PR #2141, opened 2015-07-09
Remove HTML prototypes directory.

maurizi:
I don't think these are needed anymore.


RickMohr:
I think we should keep his around for two reasons:
1) It's been a useful vehicle for communicating prototypes and it would be nice not to have to reinvent its structure.
2) There is a piece of `_template.html` that we never integrated -- clicking on the little picture icon at the upper right shows a panel with a bunch of user information

jwalgran:
I agree that we should keep these, and possibly expand on them as a design
reference as part of the management overhaul.

Issue #2114, opened 2015-06-29
Bioswales not clickable with zoom >= 14


steventlamb:
Introduced by https://github.com/OpenTreeMap/otm-tiler/commit/f7f2f8754b7029f20f901ee72e677c1848418980

maurizi:
Well the polygon should still be clickable right?

RickMohr:
We need to either add a UTF grid for the polygon layer or add logic to select polygons server-side.

PR #2144, opened 2015-07-13
Make bioswale polygons on map pages clickable

maurizi:
I explored using multiple UTF grids but ran into several issues:
 - clicks do not "fall through" to the lower UTF grid
 - max/min zoom properties are not respected

Instead, I implemented a WMS-style server backend for determining if a
polygon is present at a given lat/lng.  UTF grids are still used for point
based map features, and no network requests are made if the map does not
support polygonal map features.

Connects to #2114

### DIFF
--- a/opentreemap/stormwater/views.py
+++ b/opentreemap/stormwater/views.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from django.conf import settings
+from django.contrib.gis.geos import Point
+from django.contrib.gis.measure import D
+
+from django_tinsel.exceptions import HttpBadRequestException
+
+from stormwater.models import PolygonalMapFeature
+
+
+def polygon_for_point(request, instance):
+    lng = request.GET['lng']
+    lat = request.GET['lat']
+    point = Point(float(lng), float(lat), srid=4326)
+
+    try:
+        distance = float(request.GET.get(
+            'distance', settings.MAP_CLICK_RADIUS))
+    except ValueError:
+        raise HttpBadRequestException(
+            'The distance parameter must be a number')
+
+    features = PolygonalMapFeature.objects.distance(point)\
+        .filter(instance=instance)\
+        .filter(polygon__distance_lte=(point, D(m=distance)))\
+        .order_by('distance')[0:1]
+
+    if len(features) > 0:
+        # This is currently only being used to complement UTF grid plot data,
+        # so the structure of the response is designed to mirror the utf grid
+        return {
+            'data': {
+                'id': features[0].pk
+            }
+        }
+
+    return {
+        'data': None
+    }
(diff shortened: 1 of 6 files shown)


opentreemap/treemap/js/src/MapManager.js
```diff
@@ -58,6 +61,27 @@ MapManager.prototype = {
...
+                    emptyUtfEventStream = baseUtfEventStream.filter(shouldCheckForPolygon),
+
+                    polygonDataStream = emptyUtfEventStream.map(function(e) {
+                        return config.instance.polygonForPointUrl + format('?lng=%d&lat=%d', e.latlng.lng, e.latlng.lat);
+                    }).flatMap(BU.getJsonFromUrl);
+
+                map.utfEvents = Bacon.mergeAll(
+                    plotUtfEventStream,
+                    polygonDataStream
+                ).log();
```
jwalgran: Leftover ``log`` statement from debugging.

opentreemap/stormwater/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+    features = PolygonalMapFeature.objects.distance(point)\
+        .filter(instance=instance)\
+        .filter(polygon__distance_lte=(point, D(m=settings.MAP_CLICK_RADIUS)))\
+        .order_by('distance')[0:1]
+
+    if len(features) > 0:
+        # This is currently only being used to complement UTF grid plot data,
+        # so the structure of the response is designed to mirror the utf grid
+        return {
```
jwalgran: I like the idea of matching the UTF grid schema so that the downstream handler does not have to care.

opentreemap/stormwater/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def polygon_for_point(request, instance):
+    lng = request.GET['lng']
+    lat = request.GET['lat']
+    point = Point(float(lng), float(lat), srid=4326)
+
+    features = PolygonalMapFeature.objects.distance(point)\
+        .filter(instance=instance)\
+        .filter(polygon__distance_lte=(point, D(m=settings.MAP_CLICK_RADIUS)))\
```
jwalgran: I don't see MAP_CLICK_RADIUS defined as part of this PR.

In general, a click buffer is most effectively defined as a function of the current zoom level, since the distance-to-pixel ratio is cut in half each time you zoom in. Because we are only dealing with levels 15 through 19, it may be possible to choose a reasonable, single value.

jwalgran: Nevermind, this reused the radius parameter established by the API

jwalgran:
Tested. Trees within the bounds of a polygonal map feature are still clickable.

The functionality is good, but I think we should address the radius issue before we ship it.

maurizi:
@jwalgran I've removed the stray log statement, and I added a `distance` parameter, which the client halves for every zoom level you go in.  Please take another look.

jwalgran:
+1 (did not retest)

PR #2146, opened 2015-07-13
Feature/2125/bioswale fields

steventlamb:


opentreemap/treemap/instance.py
```diff
@@ -371,6 +371,17 @@ def get_udfc_search_fields(self, user):
 
         return data
 
+    def unlocked_udf_models(self):
```
jwalgran: What do you think about ``editable_udf_models`` instead of ``unlocked_udf_models``. "Lock" is not a verb we have used anywhere else in the python code.

steventlamb: ah, totally. This is an oversight, we already had this discussion about `is_editable` vs `is_unlocked`. You were correct and we went with `is_editable`. Will fix.

opentreemap/treemap/lib/map_feature.py
```diff
@@ -282,6 +282,15 @@ def context_dict_for_resource(request, resource, **kwargs):
...
     _add_share_context(context, request, photos)
 
+    object_name_alias = to_object_name(context['feature'].__class__.__name__)
+    # some features that were originally written to support plot and tree
+    # have grown to support other resource types, but they expect a context
+    # entry for their type, not just for 'feature'.
+    # For example:
+    # * Plot detail expects 'plot' and 'tree'
+    # * Foo detail would expect 'foo'
+    context[object_name_alias] = context['feature']
```
jwalgran: Should we ``del context['feature']`` to avoid confusion? Does that break template rendering expectations unnecessarily?

steventlamb: I don't think so. This addition makes `context_dict_for_resource` more consistent with `context_dict_for_plot`. This is because they both call `context_dict_for_map_feature`, which inserts a feature key/value into the context. I didn't write this structure but I assume it's because `feature` is used in the generic places, and then `plot` is used in the plot-specific places. The same is going on now for 'bioswale'.

PR #2158, opened 2015-07-17
Fix a few bugs

maurizi:
- Modals in the `endbody` block need to come before the `scripts` block
   in order for JS to set up event handlers.
 - Fixed importer polling


jwalgran:
+1

PR #2163, opened 2015-07-20
Feature/scalar udf defaults

steventlamb:


maurizi:
This works well and in general I think the code looks good.

I have reservations about temporarily rolling back our validations on empty choice lists though.  I would rather wait until we get a default list to merge this.

steventlamb:
I agree, and will hold off until @dboyer can provide a set of defaults, or confirm that it's totally impossible.

dboyer:
I'll follow up on getting a default list. I also submitted the question to Mr. Smarty Plants - https://www.wildflower.org/expert/

steventlamb:
closing in favor of #2181, comments addressed there

Issue #2122, opened 2015-07-01
Show "trees within" on bioswale detail


steventlamb:
For this feature, as well as some others, it would make the most sense to put a class attribute on bioswale like `can_contain_trees` and then on any mapfeature subclass detail page, have a section like:
```
{% if feature.can_contain_trees %}
<h3>Trees within:</h3>
{% for tree in feature.trees_within %}
<a href="{{ tree.detail_page_url }}">{{ tree.id }}</a>
{% endfor %}
{% endif %}
```

This will need some design work, but it should be possible to get started.

RickMohr:
If `can_contain_trees` were false

maurizi:
It should be possible to reuse the existing styling of the page, in consultation with design.

PR #2166, opened 2015-07-20
Show trees in bioswales on detail page

maurizi:
Also includes some general cleanups to the resource detail page.

Connects to #2161
Connects to #2122

### DIFF
--- a/opentreemap/treemap/lib/map_feature.py
+++ b/opentreemap/treemap/lib/map_feature.py
@@ -294,26 +294,6 @@ def context_dict_for_resource(request, resource, **kwargs):
     return context
 
 
-def title_for_map_feature(feature):
-    # Cast allows the map feature subclass to handle generating
-    # the display name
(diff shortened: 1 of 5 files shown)


maurizi:
A bioswale with trees:
with-trees

A bioswale without trees:
no-trees

I used the same title we are currently using for the map popups as the text of the link.  I'm not sure if we need more information on the link beyond that.

dboyer:
If possible, it would be useful to show the dbh next to the species since there may be multiple trees of the same species in the bioswale. If there's no dbh, we can just leave a blank or put "No diameter" - whatever's simpler.

maurizi:
@dboyer How does this look?  Note that we won't show diameter (not even "No diameter") if the plot has no tree or if the current user does not have permissions to see diameter.
with-trees

steventlamb:
+1

dboyer:
That looks great.

Issue #2162, opened 2015-07-20
Release: Cannot register for new account


dboyer:
I can complete the registration form at /accounts/register/ on the release branch but get the error page when I click "create account."

maurizi:
Relevant log output:
```
015-07-20 12:54:53,475 [ERROR] django.request: Internal Server Error: /accounts/register/
Traceback (most recent call last):
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/otm/app/opentreemap/registration_backend/views.py", line 84, in dispatch
    request, *args, **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/registration/views.py", line 91, in dispatch
    return super(RegistrationView, self).dispatch(request, *args, **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/registration/views.py", line 46, in post
    return self.form_valid(request, form)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/registration/views.py", line 94, in form_valid
    new_user = self.register(request, form)
  File "/usr/local/otm/app/opentreemap/registration_backend/views.py", line 138, in register
    email=email, password=password)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/registration/models.py", line 91, in create_inactive_user
    new_user.save()
  File "/usr/local/otm/app/opentreemap/treemap/models.py", line 411, in save
[...]

PR #2167, opened 2015-07-20
Fix incorrect initial Django Migrations

maurizi:
There were two backwards incompatible changes to the Django Abstract User
model, which were not properly captured when recreating initial
migrations for Django 1.8.

Because OTM uses a custom user model, we have to handle the migrations
for this change ourselves.

Connects to #2162


steventlamb:
+1, tested, these run, tests pass, and this removes the following message from my debugserver:
```
System check identified 2 issues (0 silenced).

You have unapplied migrations; your app may not work properly until they are applied.
Run 'python manage.py migrate' to apply them.
```

Issue #2165, opened 2015-07-20
Release: New roles only have one permission


dboyer:
I added a new role on a map which had permissions to add new roles. I could add the role but the only permission available was ability to upload photos on the tree.

PR #2168, opened 2015-07-20
Fix object_caches issue with creating new FieldPermissions

maurizi:
Since bulk_create does not send signals, the object_caches for roles
were not invalidated when creating a new role, and the roles page would
incorrectly appear blank.

Connects to #2165


maurizi:
Closing to open against the release branch

Issue #2123, opened 2015-07-01
Hide all plot/stewardship functionality on trees within bioswales


steventlamb:
A tree within a bioswale should have a limited detail page.

The results of this can be far reaching, especially affecting:
* delete tree/plot
* move tree/plot
* permissions
* quick edit

RickMohr:
It seems simpler if a tree in a bioswale is no different than any other tree. It just happens to overlap the bioswale polygon, so we can show it on the bioswale detail page.

dboyer:
When a tree is in a bioswale or other green infrastructure feature (maybe on a green roof? don't know if that's actually logistically possible), the GI feature effectively becomes the tree's planting site. The tree could be removed and the bioswale would still be there, just like a planting site.

This is particularly relevant for stewardship since actions normally done to the tree (pruning, removing litter, watering) are more accurately tracked as stewardship actions on the bioswale since they impact the bioswale as a whole. Tracking stewardship on the tree runs the risk of the action being entered twice (on the tree and on the bioswale) or just once (on the tree) where it's not the most accurate.

That said, this is on the edge of what falls within the scope of the bioswale project so we can discuss what is and is not possible.

steventlamb:
Here's a proposal/recap based on our discussions:

* If a user adds or moves a tree to an area within a bioswale's polygon, we will hide that tree's information for planting site, stewardship and (when rolled out) alerts. In its place, the planting site header will remain and a well with the following text will be present: "This tree is part of bioswale <id_num_and_link>. For stewardship actions, please go to the bioswale detail page". Styling should resemble this:
![screenshot from 2015-07-20 17 05 51](https://cloud.githubusercontent.com/assets/1699455/8787348/9b7520d0-2f01-11e5-9ab1-6af8e404cd9f.png)
* Deletion will not be changed. Even though you cannot view/edit planting site information, you can/must delete the tree and then the planting site to remove the feature from the map entirely.
* Moving will not be changed. If a tree is moved outside of a bioswale, it will resume its normal behavior.
* Quick edit will not be changed. It does not appear to have planting site details.
* Permissions will not be changed. Because this is not a security concern, hiding these fields in the UI is sufficient to effectively disable editing them. Their underlying permissions will still prevent unauthorized users from editing them.

PR #2171, opened 2015-07-21
Hide plot/udf data on trees within pmfs

steventlamb:
connects #2123 on github

Also refreshes the page when a plot is moved in or out of a polygonal
map feature.

### DIFF
--- a/opentreemap/treemap/templates/treemap/plot_detail.html
+++ b/opentreemap/treemap/templates/treemap/plot_detail.html
@@ -49,6 +49,7 @@
     </tbody>
   </table>
 
+  {% if not containing_polygonalmapfeature %}
   {# Render collection UDF fields for the Tree #}
   {% if tree %}
     {% for udf in tree.collection_udfs %}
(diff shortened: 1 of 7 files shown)


opentreemap/treemap/templates/treemap/plot_detail.html
```diff
@@ -57,10 +58,21 @@
...
+  {% endif %}
 </div>
 
+<h3>{% trans "Planting Site Information" %}</h3>
+{% if containing_polygonalmapfeature %}
+<div class="well" id="containing-polygonalmapfeature">
+  {% with pmf_id=containing_polygonalmapfeature.pk|unlocalize %}
+  {% url 'map_feature_detail' instance_url_name=request.instance.url_name feature_id=pmf_id as polygonal_url %}
+  {% blocktrans with display_name=containing_polygonalmapfeature.display_name %}
+  This tree is part of <a href="{{ polygonal_url }}">{{ display_name }} #{{ pmf_id }}.</a> For stewardship actions, please go to the bioswale detail page.
```
jwalgran: "please go to the bioswale detail page" should be something like "please go to the {{ display_name }} detail page" so the same text will work for another polygonal type. 

It is redundant, but it may be nice if the "go to..." text is also a link to to the polygonal map feature.

steventlamb: fine by me

jwalgran:
This works nicely.

steventlamb:
changes made.

PR #2174, opened 2015-07-27
Cleanups to Field Permission and Roles

maurizi:
- Update FieldPermission choices and translate them.
 - Split inlineEditForm.js into 2 modules, and add a "simple" version
   The simple version is intended to be used for forms that want the
   edit/save/cancel behaviour but want to handle form serialization and
   error reporting in a custom way.
 - Add basic styling for updated roles page
 - Fix potential object_caches invalidation issue with field permissions


RickMohr:
+1

PR #2179, opened 2015-07-29
Feature/multiselect widget

steventlamb:


maurizi:
The multiselect widget is not properly handling canceling.  If I go into edit mode, check some boxes, cancel, and then go into edit mode again, those boxes are still checked, when they should go back to how they were.

opentreemap/treemap/udf.py
```diff
@@ -358,11 +358,28 @@ def _update_choice_scalar(self, old_choice_value, new_choice_value):
...
+            for model in Model.objects\
+                              .filter(**{'udf:%s' % self.name:
+                                         old_choice_value}):
+                model.udfs[self.name] = new_choice_value
+                model.save_base()
+
+        else:
+            udf_filter = {'instance': self.instance,
+                          # grab anything with that key
+                          'udf:%s__contains' % self.name: ''}
```
maurizi: There is a hstore-specific way of filtering by key, which is more efficient and more accurate (http://djangonauts.github.io/django-hstore/#_python_api).

Basically, you just wrap the RHS in a list and it will search the hstore column for rows containing those keys:
```
 udf_filter = {'instance': self.instance,
               # grab anything with that key
               'udfs__contains': [self.name]}

File: opentreemap/treemap/udf.py [hunk 2]
```diff
@@ -358,11 +358,28 @@ def _update_choice_scalar(self, old_choice_value, new_choice_value):
...
         Model = safe_get_udf_model_class(self.model_type)
-        for model in Model.objects\
-                          .filter(**{'udf:%s' % self.name:
-                                     old_choice_value}):
-            model.udfs[self.name] = new_choice_value
-            model.save_base()
+
+        if self.datatype_dict['type'] == 'choice':
+            for model in Model.objects\
+                              .filter(**{'udf:%s' % self.name:
```
maurizi: Consider using `self.canonical_name` here instead of `'udf:%s' % self.name`

opentreemap/treemap/udf.py
```diff
@@ -375,15 +392,35 @@ def _update_choice_scalar(self, old_choice_value, new_choice_value):
...
+            if new_choice_value is None:
+                cval_audits.delete()
+                pval_audits.delete()
+            else:
+                cval_audits.update(current_value=new_choice_value)
+                pval_audits.update(previous_value=new_choice_value)
         else:
-            cval_audits.update(current_value=new_choice_value)
-            pval_audits.update(previous_value=new_choice_value)
+            for audit in audits.filter(field='udf:%s' % self.name,
```
maurizi: Consider using `self.canonical_name` here instead of `'udf:%s' % self.name`

steventlamb:
@maurizi all comments addressed

steventlamb:
@maurizi rebased, tests pass, poke test still works

maurizi:
+1

Issue #2172, opened 2015-07-23
Bioswale does not have comment disclaimer text


steventlamb:
![screenshot from 2015-07-23 14 04 25](https://cloud.githubusercontent.com/assets/1699455/8858143/cfd3a24a-3143-11e5-9322-7c9ff2c19c88.png)

should look like:

![screenshot from 2015-07-23 14 05 13](https://cloud.githubusercontent.com/assets/1699455/8858156/dcae4ae2-3143-11e5-93af-30765a6ba75f.png)

dboyer:
"a tree" should probably be replaced with either "a resource" or with whatever terminology the map owner sets.

PR #2180, opened 2015-07-29
Add comment disclaimer to all resources

steventlamb:
connects #2172 on github

### DIFF
--- a/opentreemap/treemap/templates/treemap/map_feature_detail.html
+++ b/opentreemap/treemap/templates/treemap/map_feature_detail.html
@@ -193,13 +193,19 @@
             {% endif %}
           </div>
 
-          {% if feature.is_plot %}
           <div id="comment_disclaimer">
             <p class="text-muted">
-              <em>The comment system does not serve as a way to report problems with a tree.</em>
+              <em>
+                {% if feature.is_plot %}
+                {% trans "The comment system does not serve as a way to report problems with a tree." %}
+                {% else %}
+                {% blocktrans with resource=term.resource %}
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/templates/treemap/map_feature_detail.html
```diff
@@ -193,13 +193,21 @@ <h4 class="address" id="map-feature-address">{{ feature.address_full }}</h4>
             {% endif %}
           </div>
 
-          {% if feature.is_plot %}
           <div id="comment_disclaimer">
             <p class="text-muted">
-              <em>The comment system does not serve as a way to report problems with a tree.</em>
+              <em>
+                {% if feature.is_plot %}
+                {% blocktrans with resource=term.resource %}
```
jwalgran: We can drop the unused "with" variable in this first ``blocktrans``.

jwalgran:
This does show disclaimers for resources, but there is problem with a/an agreement when using terminology that starts with a vowel. I don't know if it is an edge case worth worrying about.

screen shot 2015-07-29 at 3 06 12 pm

jwalgran:
+1 tested

steventlamb:
@jwalgran @RickMohr it looks like the problem of 'a' vs 'an' would not be exclusive to this new text:
https://github.com/OpenTreeMap/otm-core/blob/master/opentreemap/opentreemap/context_processors.py#L60
for example:
```
             data-action='addresource'>{% trans "Add a" %} {{ term.Resource }}</a>
```

Should we revisit this with a new issue?

jwalgran:
:+1:  to a new issue.

PR #2190, opened 2015-07-31
Feature/mw/scenario protoype

designmatty:
Supported by https://github.com/OpenTreeMap/otm-addons/tree/feature/mw/scenario-protoype


RickMohr:
+1 Looks great!

Issue #2189, opened 2015-07-30
The audit detail for a multi-value field shows as a python list instead of a plain comma-separated list


jwalgran:
<img width="293" alt="screen shot 2015-07-30 at 11 45 08 am" src="https://cloud.githubusercontent.com/assets/17363/8991984/d99fb580-36b0-11e5-8f61-a0b1c79c2f6e.png">

PR #2192, opened 2015-08-04
Format lists as readable text in audit description

steventlamb:
connects #2189 on github

before:
screenshot from 2015-08-04 15 14 30

after:
screenshot from 2015-08-04 15 21 30

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -1319,6 +1319,10 @@ class Audit(models.Model):
                 value = value.area
         elif isinstance(value, datetime):
             value = dformat(value, settings.SHORT_DATE_FORMAT)
+        elif isinstance(value, list):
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/audit.py
```diff
@@ -1319,6 +1319,8 @@ def _unit_format(self, value):
                 value = value.area
         elif isinstance(value, datetime):
             value = dformat(value, settings.SHORT_DATE_FORMAT)
+        elif isinstance(value, list):
+            value = '(%s)' % ', '.join(value) if value else 'none'
```
maurizi: You should probably wrap `'none'` in `_(...)` for translation, with a translation comment for context.

RickMohr:
Works well. Agree with Mike's comment.

PR #2196, opened 2015-08-07
Fix a few issues with audits/permissions for Bioswales

maurizi:
Stop tracking `PolygonalMapFeature` subclass pointer
Change full_name property of `FieldPermission`


RickMohr:
+1!!

Issue #2198, opened 2015-08-10
Can't add bioswale via menu


RickMohr:
The "+" icon in the page-top menu is supposed to control a dropdown containing "Add Tree", "Add Bioswale", etc. But currently it does nothing.

PR #2201, opened 2015-08-12
Fix add resource menu

maurizi:
- The add resource menu was broken due to using `true` instead of `True`.  Since it should always be enabled, I removed all of the button enabler related attributes.
 - Add plot and resource permissions were incorrect for _all_ of the add tree/resource buttons on the site.  I registered the `plot_is_creatable` helper function as a template filter, and switched to using it in several places.
 - We were checking plot permissions instead of resource permissions when enabling/disabling buttons for resources.  I added separate helper functions for these, and fixed some typos in other places where they were already being used.

Connects to #2198


maurizi:
Closing to reopen against the release branch.

PR #2203, opened 2015-08-12
When adding a feature, clicking "Done" twice no longer adds twice

RickMohr:
Also, show toast if adding causes an unexpected error (e.g. a 500).

Connects #2188


maurizi:
+1, works well.

Issue #2155, opened 2015-07-16
Add a “continue editing” option to the last step of the add bioswale wizard


jwalgran:
Gate the availability of this option using the “is_editable” property of the map feature.

PR #2204, opened 2015-08-12
Enable continue editing link when adding Bioswales

maurizi:
Connects to #2155


RickMohr:
+1. I could "continue editing" when adding a tree or bioswale but not an LA rain garden.

Issue #2184, opened 2015-07-30
Bioswale sidebar keeps stale data


steventlamb:
'Tree Details' shows, and ecobenefits don't collapse/clear:
![screenshot from 2015-07-30 13 35 20](https://cloud.githubusercontent.com/assets/1699455/8990375/d7707222-36bf-11e5-8f78-a086c1a373e1.png)

PR #2207, opened 2015-08-12
Genericize the map page browse mode detail header

jwalgran:
I have changed the first header on the sidebar of the browse trees mode of the map page to "Details" instead of the overly specific "Tree Details." We are showing the details for green infrastructure features in this same area.

Tree selected 

screen shot 2015-08-12 at 3 23 20 pm

Stormwater feature selected

screen shot 2015-08-12 at 3 23 38 pm

The eco benefits being shown are for the entire map, and are always shown. Single-tree benefits are correctly removed from the sidebar when switching from displaying a tree to displaying a stormwater feature.

Connects to #2184

### DIFF
--- a/opentreemap/treemap/templates/treemap/map-browse-trees.html
+++ b/opentreemap/treemap/templates/treemap/map-browse-trees.html
@@ -4,7 +4,7 @@
 <div class="panel" id="map-info">
   <div class="panel-group">
     <div class="panel-heading">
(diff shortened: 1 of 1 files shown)


jwalgran:
@steventlamb assigned to you because I implemented a very simple solution to the issue and wanted to get your feedback.

@dboyer I would also like to get your opinion on this simple solution to the issue.

steventlamb:
+1 I think this is a nice way to handle this.

PR #2218, opened 2015-08-17
Fix saving models when collection UDFs are not writable.

maurizi:
Setting collection UDFs to read-only would prevent editing a model if
that model already had collection UDF data on it.  The fix is rather
simple - we no longer attempt to save collection UDF data if there have
been no changes.

I also fixed an issue with clean_value for collection UDFs where it would
attempt to parse a datetime object as a formatted string and turn it into
a date.  This was mostly a symptom, not a cause of the underlying issue,
but it is more correct anyways.

### DIFF
--- a/opentreemap/treemap/udf.py
+++ b/opentreemap/treemap/udf.py
@@ -6,6 +6,7 @@ from __future__ import division
 import json
 import copy
 import re
+from datetime import date, datetime
 
 from django.core.exceptions import ValidationError, FieldError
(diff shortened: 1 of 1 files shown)


steventlamb:
please link this commit message to the issue it fixes

steventlamb:
Needs a fix for instances that have multiple udfs per model

maurizi:
@steventlamb Code fixed, unit tests pass.

steventlamb:
+1

PR #2222, opened 2015-08-19
Feature/2213/delete any map feature

steventlamb:


maurizi:
+1, works well, code looks good.

PR #2224, opened 2015-08-19
Feature/2217/filter creatable resources

steventlamb:


maurizi:
This goes against what we usually do with things a user doesn't have permission to do, which is to leave the field/button there but disable it with an explanatory tooltip.

I would suggest changing this to continue following that pattern, by disabling the checkboxes you don't have permissions for instead of hiding them.

RickMohr:
@maurizi is right that usually we show people what they're not allowed to do. In this case it doesn't seem so bad to not be able to see GSI types you're not allowed to create. @dboyer, any opinion?

RickMohr:
+1 Works well and code looks good.

dboyer:
It should be hidden. Public users don't need to see the green infrastructure features that the map administrator is preventing them from adding to the map.

steventlamb:
@RickMohr how about this change instead of a comment (see newest commit)?

RickMohr:
Awesomely clear.

PR #2227, opened 2015-08-25
Add more GSI models

steventlamb:


opentreemap/stormwater/templates/stormwater/RainBarrel_add.html
```diff
@@ -0,0 +1,11 @@
+{% extends "treemap/resource_add.html" %}
+{% load i18n %}
+{% load l10n %}
+{% load form_extras %}
+
+{% block resource_add_details %}
+<div class="resource-question">
+  {% trans "How many gallons does the tank hold?" as capacity %}
```
RickMohr: This label should avoid the term "gallons" so it will still make sense on a site using liters. Could change to "How much water does the tank hold?"

RickMohr:
To make rain barrel capacity participate in unit conversions you'll need to add something like
https://github.com/OpenTreeMap/otm-build/blob/master/ansible/roles/app/templates/local_settings.j2#L277
to
https://github.com/OpenTreeMap/otm-core/blob/master/opentreemap/opentreemap/settings/default_settings.py#L281

When successful you should be able to choose e.g. "liters" on the management "Units" page and see it when adding a rain barrel or displaying the field on the detail page.

opentreemap/stormwater/templates/stormwater/RainBarrel_detail.html
```diff
@@ -0,0 +1,9 @@
+{% extends "treemap/resource_detail.html" %}
+{% load i18n %}
+{% load l10n %}
+{% load form_extras %}
+
+{% block resource_details %}
+{% trans "Capacity in gallons" as capacity %}
```
maurizi: Similar to Rick's comments above, it would be ideal to avoid mentioning the units in the hardcoded label, since it can be changed to liters.

opentreemap/treemap/search.py
```diff
@@ -406,7 +411,10 @@ def _parse_dict_value_for_mapping(mapping, valuesdict, field=None):
 
 
 def _parse_predicate_pair(key, value, mapping):
-    model, search_key = _parse_predicate_key(key, mapping)
+    try:
+        model, search_key = _parse_predicate_key(key, mapping)
+    except ModelParseException:
+        return FilterContext()
```
maurizi: Consider adding a comment here explaining why this is being done.

Also, you may be able to be a little more careful here, so that we don't swallow exceptions for simple typos:
```python
try:
    model, search_key = _parse_predicate_key(key, mapping)
except ModelParseException:
    if _is_udf(key):
        # We should return no results for collection UDF searches on unrelated models
        return FilterContext()
    else:
        raise
```

maurizi:
When going through the add process for `Rain Barrel`, it indicates that there are 5 steps, but the last step is step 4 out of 5.

maurizi:
I was able to add edit and search for rain barrels and rain gardens.  I want to take a quick look through the code again after the few remaining issues are sorted out, but overall :+1:

maurizi:
:+1: after commits are squashed.

Issue #2220, opened 2015-08-17
Moving a tree in/out of a bioswale causes bioswale detail to load


steventlamb:
it looks like this has to do with the bioswale polygon receiving a click event

RickMohr:
+1 Good solution, works well. 

Too bad the enable/disable object code is complex, but without using true inheritance I don't think it could be simplified.

PR #2228, opened 2015-08-26
Disable clicking other objects while moving a geom

steventlamb:
connects #2220 on github


opentreemap/treemap/js/src/mapFeature.js
```diff
@@ -135,6 +136,7 @@ exports.init = function(options) {
     clickedIdStream
         .filter(BU.not, options.featureId)
         .map(_.partialRight(U.appendSegmentToUrl, detailUrlPrefix, false))
+        .filter(R.not(currentMover.isEnabled))
```
RickMohr: You could consider defining `isDisabled()` instead of `isEnabled()` and removing this `R.not`

PR #2229, opened 2015-08-26
Increase margin on management page content sections.

maurizi:
Prevents the content at the bottom of the page from being obscured by
the footer.


RickMohr:
+1

PR #2232, opened 2015-08-28
Replace FreeNode IRC link with Gitter channel

maurizi:


jwalgran:
This documentation leads me to believe that only members of the OpenTreeMap organization can use this room

https://gitter.zendesk.com/hc/en-us/articles/200178951-Gitter-rooms

I think we need to use ``opentreemap/otm-core`` as our public discussion room.

dboyer:
Yes, I'd like this to be as public as possible.

maurizi:
@jwalgran Updated to use `OpenTreeMap/otm-core` as the chat room.

jwalgran:
+1

PR #2234, opened 2015-09-01
Saner units methods

RickMohr:


jwalgran:
Downloads are not working for me in the latest Safari (version 8, Yosemite). The feature test on this line is not passing: https://github.com/OpenTreeMap/otm-core/pull/2234/files#diff-99b22f02ada35d2ddcfc00c562eeae9dR187

Googleing points to data URIs as a solution. This lib, which advertised Safari support, may help us deal with the issue. https://github.com/eligrey/FileSaver.js/

RickMohr:
@jwalgran, could you please try the FileSaver demo (http://eligrey.com/demos/FileSaver.js)? The docs say that on Safari you won't get a file name with your download. But do you get a file type? i.e. when you try to open the downloaded file does it offer you appropriate app choices or not? If not I can duplicate that functionality with a few lines of code and avoid bringing in another dependency.

jwalgran:
The text demo is not a good experience:

https://drive.google.com/file/d/0B5hgVC8cVVtwNkR3NVZrUDk4ZVE/view?usp=sharing

RickMohr:
Can you please try this fixup?

jwalgran:
The good new is that we do not have to bring in the library. The bad new is, it has the same crummy behavior as the FileSaver.js demo. It is possible to save a file this way, but you have to know that you can File > Save As and you have to know to switch the dropdown at the bottom of the save sheet to "Page Source"

screen shot 2015-09-03 at 1 29 22 pm

I don't think this can be our final implementation, but I am not opposed to creating an a separate issue for improving the export experience on Safari.

RickMohr:
Entered #2243 for the Safari issue.

PR #2236, opened 2015-09-02
Update gunicorn and django debug toolbar

maurizi:
Gunicorn got the reload option in version 19, which is handy.
The Django Debug Toolbar got Django 1.8 support in 1.3, and it wouldn't
work until I upgraded it.

It _looks_ like Gunicorn got a big version increase, but I think what happened is they decided to drop the leading `0.`


RickMohr:
+1

PR #2241, opened 2015-09-02
Cleanup some cruft with UDFs

maurizi:
- Validate the choices list for multichoice UDFs
 - Allow UDFs with the same name if they are on different models
 - Add a uniqueness constraint to UserDefinedFieldDefinition

### DIFF
--- a/opentreemap/treemap/lib/udf.py
+++ b/opentreemap/treemap/lib/udf.py
@@ -39,10 +39,7 @@ def udf_create(params, instance):
 
     udfs = UserDefinedFieldDefinition.objects.filter(
         instance=instance,
-        # TODO: why isn't this also checking model name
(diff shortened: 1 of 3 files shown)


opentreemap/treemap/lib/udf.py
```diff
@@ -39,10 +39,7 @@ def udf_create(params, instance):
 
     udfs = UserDefinedFieldDefinition.objects.filter(
         instance=instance,
-        # TODO: why isn't this also checking model name
-        # is there some reason the same name can't appear
-        # on more than one model?
-        # Too scared to change this.
+        model_type=model_type,
```
steventlamb: !

steventlamb:
+1, tested

PR #2244, opened 2015-09-04
Rename 'supports_resource' -> 'has_resources'

steventlamb:
Determining if an instance has resources is now a poor fit for
determining support for resources because it imposes an inception
problem. An instance must support resources for a resource to be added,
but a resource must be added for an instance to support
resources. Renaming this function will make it semantically clearer to
use other criteria for testing resource support.


RickMohr:
+1

PR #2245, opened 2015-09-08
Add styling and refactor for management page notifications

maurizi:


jwalgran:
The notification bubbles for community moderation are a different color than the ones for custom fields. Is this by design?

screen shot 2015-09-09 at 6 34 45 pm

maurizi:
@jwalgran Yes, they're colored differently in the wireframes.

I'm not sure, but I think the intention was to differentiate between notifications from user activity (like someone posting a new photo or comment), and notifications from admin activity (adding a new field or user, imports)?

jwalgran:
Are the UI test failures legitimate? If not, :+1:

PR #2252, opened 2015-09-11
Fix comment moderation UI tests.

maurizi:
Because the Bacon.js stream was change from using onValue to using flatMap
to support further usage, it stopped working when it was *not* used,
due to Bacon's lazy evaluation.

Adding a noop onValue fixes the issue.

Connects to #2247


RickMohr:
+1 (did not test)

Issue #2248, opened 2015-09-10
Rain Barrel units not honoring decimal point limits in all places


dboyer:
to reproduce:
add rainbarrel to a map
set units on rainbarrel capacity to n
add a rainbarrel
visit rainbarrel detail
edit capacity, with a number of decimal points greater than n
save
observe that the inlineeditform displays the correct units but the recent edits sidebar does not.

PR #2253, opened 2015-09-11
Fix audit unit format for multi-word names

steventlamb:
connects #2248 on github

As was the case with bb11caad41f9b2feb1da8c387f9ddef6be5f90aa, unit
field lookup was failing for case-sensitivity reasons on multi-word
models. This is because we were calling `.lower()` on a camel-case
string. In this case, we were producing 'rainbarrel' when we wanted 'rainBarrel'.


jwalgran:
I think the test failures will be fixed by rebasing a recent master.

jwalgran:
+1 (did not test)

PR #2254, opened 2015-09-11
Add RainGarden stewardship defaults

steventlamb:


maurizi:
I think we have another case sensitivity issue going on.

I tried to save a stewardship action on a rain garden, and got the following issue:
```
Exception at /mike/features/3/
Malformed request - invalid field raingarden.udf:Stewardship
```

RickMohr:
Looks like a good case for a unit test.

steventlamb:
this is ready for another look

jwalgran:
I added RainGarden to a newly created test map, saw the new stewardship choices, and was able to add a rain garden with stewardship and search for the specific stewardship type.

:+1:

PR #2255, opened 2015-09-11
[requires.io] dependency update on master branch

maurizi:


maurizi:
So this is an automated pull request from [Requires.io](https://requires.io/), which I signed us up for.

It's supposed to send us a PR whenever our dependencies change (bundled up into a digest if more than 1 changes in a day).

This one has a lot of changes because many of our dependencies are out-of-date, but I'm hoping if we get this taken care of then we'll only get a few small updates every once in a while, and we can keep things up-to-date fairly easily.

maurizi:
It looks like this PR isn't working because of something weird with `modgrammar-py2`, which says on it's PyPi page that there is a `0.9.2` version, but when you try and install it says the most recent version is `0.9.1`.  I've created an issue on the maintainer's repo alerting them of this, and if they don't get back to us we can keep it pinned to `0.9.1` and add a commit to tell `requires.io` that we want it to stay at that version.

PR #2262, opened 2015-09-14
Fix running JS tests via testem CI.

maurizi:
The `before_tests` script specified in our testem config was failing
since Grunt needs to run in the virtualenv in order to run collectstatic.

When this script fails, testem just hangs forever for some reason.

I fixed the issue by adding a new Grunt task which just runs the `js` and
`css` tasks (which used to be the default Grunt task).


RickMohr:
+1

Issue #2264, opened 2015-09-15
Importer: fix annoyances


RickMohr:
* Compute completed row counts more efficiently
* Sort import tables by date (most recent first)
* Show "(None)" if an import table has no rows

PR #2265, opened 2015-09-15
Add pagination controls for tables on main importer page

RickMohr:
Also fix several small importer issues.

Connects #2259 
Connects #1918
Connects #2264


maurizi:
This doesn't play well with the auto-refreshing when there are imports in progress.  If I'm viewing page 2, when the table refreshes it puts me back at page 1.

RickMohr:
@maurizi, good catch, all fixed.

maurizi:
+1

Issue #2261, opened 2015-09-11
Use a more performant query for tree import proximity validation


jwalgran:
~~Make use of PostGIS nearest neighbor index~~

The same performance can be achieved using a bounding box filter

PR #2267, opened 2015-09-15
[DO NOT MERGE] Use indexed search for import proximity validation

jwalgran:
**NOTE: When an instance has zero trees or 1 tree, this new query takes > 3s to run per import row. This is probably related to the use of ``LIMIT 2``. This situation must be dealt with before merging the PR**

Using the PostGIS nearest neighbor index to find nearby trees is 2 orders of magnitude faster than the ST_Distance based query generated by the Django ORM.

These are some example queries, with timings from a development VM with an instance containing > 200,000 plots.

```
SELECT id, ST_Distance("treemap_mapfeature"."the_geom_webmercator",
  ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860), 4326),3857)) as "distance"
FROM treemap_mapfeature
WHERE "treemap_mapfeature"."instance_id" = N
AND ST_distance("treemap_mapfeature"."the_geom_webmercator",
  ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860), 4326),3857)) <= 3.048
ORDER BY distance ASC LIMIT 5;

491.045 ms
```

```
SELECT id, ST_Distance(the_geom_webmercator,
  ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860),4326), 3857)) AS distance
FROM treemap_mapfeature
WHERE "treemap_mapfeature"."instance_id" = N
ORDER BY the_geom_webmercator <-> ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860),4326), 3857)
LIMIT 5;

1.482 ms
```

The nearest neighbor search is not a 1-to-1 replacement, since it returns trees no matter how far away they are. Since the record count is always fixed and small, post-processing in Python code is not an issue.

To test this I used import files with 1000 randomly placed trees. The default import batch size of 250 created 4 celery tasks per operation. Here are my timings on a app VM with 3GB of RAM running 2 celery workers querying a database VM with 1GB of RAM.

Original Validation
298s
296s
287s
287s

Original Commit
315s
315s
310s
311s

Raw SQL Validation
34s
34s
36s
36s

Raw SQL Commit
61s
60s
59s
59s

If I replace the body of the ``proximity_validation`` function with ``return True`` the batches run a few seconds faster, but it is only a 5-10% decrease in processing time in this sample set.

---

Reference: http://boundlessgeo.com/2011/09/indexed-nearest-neighbour-search-in-postgis/

Connects to #2261

### DIFF
--- a/opentreemap/importer/models/trees.py
+++ b/opentreemap/importer/models/trees.py
@@ -6,7 +6,7 @@ from __future__ import division
 from django.core.exceptions import ValidationError
 from django.contrib.gis.db import models
 from django.contrib.gis.geos import Point
-from django.contrib.gis.measure import D
+from django.db import connection
 from django.utils.translation import ugettext as _
 
 from treemap.models import Species, Plot, Tree
@@ -250,30 +250,59 @@ class TreeImportRow(GenericImportRow):
         return True
 
     def validate_proximity(self, point):
-        plot_ids_from_this_import = TreeImportRow.objects\
-            .filter(import_event=self.import_event)\
-            .filter(plot__isnull=False)\
-            .values_list('plot__pk', flat=True)
+        """Return False if the point is within 10ft of an existing
+        plot, excluding plots that are part of the same import event.
 
-        nearby = Plot.objects\
-                     .filter(instance=self.import_event.instance)\
-                     .filter(geom__distance_lte=(point, D(ft=10.0)))\
-                     .distance(point)\
-                     .exclude(pk__in=plot_ids_from_this_import)\
+        This is written with as a raw SQL query to take advantage of
+        the PostGIS <-> operator which provides indexed nearest
+        neighbor searching.
 
-        oid = self.cleaned.get(fields.trees.OPENTREEMAP_PLOT_ID, None)
-        if oid:
-            nearby = nearby.exclude(pk=oid)
+        The `AND NOT IN` clause skips plots created by the same import
+        event, assuming that your import file is correct, even though
+        trees are close together.
 
-        nearby = nearby.order_by('distance')[:5]
+        `LIMIT 2` because we may be updating an existing plot, in
+        which case the nearest plot will be the plot itself, and we
+        will skip over it looking for the next nearest plot.
+        """
+        sql = ('SELECT id, '
+               '  ST_Distance(the_geom_webmercator, '
+               '              ST_SetSRID(ST_MakePoint(%(x)s, %(y)s), '
+               '                         %(srid)s)) '
+               '  AS distance '
+               'FROM treemap_mapfeature '
+               'WHERE instance_id = %(instance_id)s '
+               'AND NOT id IN ( '
+               '  SELECT plot_id '
+               '  FROM importer_treeimportrow tir '
+               '  WHERE tir.import_event_id = %(event_id)s '
+               '  AND tir.plot_id IS NOT NULL) '
+               'ORDER BY the_geom_webmercator <-> '
+               '  ST_SetSRID(ST_MakePoint(%(x)s, %(y)s), %(srid)s) '
+               'LIMIT 2;')
 
-        if len(nearby) > 0:
+        nearby = []
+        with connection.cursor() as c:
+            c.execute(sql, {'instance_id': self.import_event.instance.id,
+                            'event_id': self.import_event.id,
+                            'x': point.x,
+                            'y': point.y,
+                            'srid': point.srid})
+            nearby = c.fetchall()
+
+        existing_plot_id = self.cleaned.get(fields.trees.OPENTREEMAP_PLOT_ID,
+                                            None)
+        if existing_plot_id:
+            # `p` is a tuple of (id, distance)
+            nearby = [p for p in nearby if p[0] != existing_plot_id]
+
+        if len(nearby) > 0 and nearby[0][1] < 3.048:  # 10ft in meters
             flds = (fields.trees.POINT_X, fields.trees.POINT_Y)
-            if nearby[0].distance.m < 0.001:
+            if nearby[0][1] < 0.001:
                 self.append_error(errors.DUPLICATE_TREE, flds)
             else:
                 self.append_error(errors.NEARBY_TREES, flds,
-                                  [p.pk for p in nearby])
+                                  [row[0] for row in nearby])
             return False
         else:
             return True


opentreemap/importer/models/trees.py
```diff
@@ -250,30 +250,59 @@ def validate_otm_id(self):
...
+               'FROM treemap_mapfeature '
+               'WHERE instance_id = %(instance_id)s '
+               'AND NOT id IN ( '
+               '  SELECT plot_id '
+               '  FROM importer_treeimportrow tir '
+               '  WHERE tir.import_event_id = %(event_id)s '
+               '  AND tir.plot_id IS NOT NULL) '
+               'ORDER BY the_geom_webmercator <-> '
+               '  ST_SetSRID(ST_MakePoint(%(x)s, %(y)s), %(srid)s) '
+               'LIMIT 2;')
```
maurizi: > `LIMIT 2` because we may be updating an existing plot, in which case the nearest plot will be the plot itself

If we filter out any possible existing plot in the SQL query, can we remove the ``LIMIT 2`?

Maybe by adding something like this to the `WHERE` clause (with `existing_plot_id` possibly being null)?
`AND id IS DISTINCT FROM %(existing_plot_id)`

jwalgran: We need a `LIMIT`, otherwise the query will return every plot on the map that is not part of the current import, ordered by distance. Pushing the existing plot check into the query would let us use `LIMIT 1`. Without measuring I am not sure what effect this will have.

opentreemap/importer/models/trees.py
```diff
@@ -250,30 +250,59 @@ def validate_otm_id(self):
...
+        This is written with as a raw SQL query to take advantage of
+        the PostGIS <-> operator which provides indexed nearest
+        neighbor searching.
 
-        oid = self.cleaned.get(fields.trees.OPENTREEMAP_PLOT_ID, None)
-        if oid:
-            nearby = nearby.exclude(pk=oid)
+        The `AND NOT IN` clause skips plots created by the same import
+        event, assuming that your import file is correct, even though
+        trees are close together.
```
RickMohr: I suggest you change this comment to:
The `AND NOT IN` clause skips plot ID's present in the import. That's because an import may update an existing plot or tree by referencing its existing plot ID. We don't want to flag those as duplicates.

RickMohr: Sorry, I was confused.

But since this clause excludes plot ID's in the import I think it is also handling the case of updating an existing plot. Which means we could use `LIMIT 1` instead of `LIMIT 2`. Which should mean the only slow edge case is an instance with 0 trees? If so we could start the import with an `exists()` query on `Plot` and use it to skip proximity checks.

RickMohr:
Here's another option not suffering from the `LIMIT` issue -- since we only care about plots within 10 feet of the given plot, find all plots within a 20 foot bounding box around the given plot. That will be very fast and return few results, which can be quickly processed in Python code as in your method.

opentreemap/importer/models/trees.py
```diff
@@ -250,30 +250,59 @@ def validate_otm_id(self):
...
+        nearby = []
+        with connection.cursor() as c:
+            c.execute(sql, {'instance_id': self.import_event.instance.id,
+                            'event_id': self.import_event.id,
+                            'x': point.x,
+                            'y': point.y,
+                            'srid': point.srid})
+            nearby = c.fetchall()
+
+        existing_plot_id = self.cleaned.get(fields.trees.OPENTREEMAP_PLOT_ID,
```
RickMohr: I think any `existing_plot_id` has already been excluded by the `NOT id IN` clause and so this code can be omitted.

jwalgran:
Using a subquery produces the same plan. Still slow for an empty map. I think the only way out of this is to skip the proximity check for an empty map. We don't want to count all the trees on the map each time we process a tree, so the empty/not empty check needs to be done once, and passed into the celery task as a boolean.

RickMohr:
Agreed, compute once and pass in. What about the bounding box approach, which should be simplest of all?

jwalgran:
It looks like an `EXISTS` test is nearly free.

```
otm=# explain analyze  select case when  exists ( select id from treemap_mapfeature where instance_id = 21) then true else false end;
                                                         QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.06..0.07 rows=1 width=0) (actual time=0.239..0.239 rows=1 loops=1)
   InitPlan 1 (returns $0)
     ->  Seq Scan on treemap_mapfeature  (cost=0.00..32109.68 rows=504979 width=0) (actual time=0.238..0.238 rows=1 loops=1)
           Filter: (instance_id = 21)
           Rows Removed by Filter: 66
 Planning time: 0.086 ms
 Execution time: 0.275 ms
(7 rows)
```

But the bounding box filter is on the same order of magnitude.

```
explain analyze SELECT id,                                                                       ST_Distance(the_geom_webmercator,                                                                                ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860), 4326), 3857))                   AS distance                                                                                        FROM treemap_mapfeature                                                                              WHERE instance_id = 21                                                                               AND id != 803762                                                                                     AND ST_intersects(the_geom_webmercator, st_geomfromtext('POLYGON((-13240855.7139095 3868151.74246754,-13240855.7139095 4139894.06157828,-13096334.4335878 4139894.06157828,-13096334.4335878                3868151.74246754,-13240855.7139095 3868151.74246754))', 3857))                                     AND NOT id IN (                                                                                        SELECT plot_id                                                                                       FROM importer_treeimportrow tir                                                                      WHERE tir.import_event_id = 1                                                                        AND tir.plot_id IS NOT NULL)                                                                       ORDER BY the_geom_webmercator <->                                                                      ST_Transform(ST_SetSRID(ST_MakePoint(-118.1289,34.0860), 4326), 3857)                              LIMIT 1;
                                                                                                                                                                  QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=242.15..248.13 rows=1 width=36) (actual time=0.283..0.283 rows=1 loops=1)
   ->  Index Scan using treemap_mapfeature_the_geom_webmercator_id on treemap_mapfeature  (cost=242.15..245383.99 rows=40983 width=36) (actual time=0.283..0.283 rows=1 loops=1)
         Index Cond: (the_geom_webmercator && '0103000020110F00000100000005000000BC58D8F6424169C1252D09DFFB824D41BC58D8F6424169C10DCCE107BB954F4186F3DFCDB1FA68C10DCCE107BB954F4186F3DFCDB1FA68C1252D09DFFB824D41BC58D8F6424169C1252D09DFFB824D41'::geometry)
         Order By: (the_geom_webmercator <-> '0101000020110F000081FBDE1FEC1469C1E5CB17CA51D34E41'::geometry)
         Filter: ((id <> 803762) AND (NOT (hashed SubPlan 1)) AND (instance_id = 21) AND _st_intersects(the_geom_webmercator, '0103000020110F00000100000005000000BC58D8F6424169C1252D09DFFB824D41BC58D8F6424169C10DCCE107BB954F4186F3DFCDB1FA68C10DCCE107BB954F4186F3DFCDB1FA68C1252D09DFFB824D41BC58D8F6424169C1252D09DFFB824D41'::geometry))
         Rows Removed by Filter: 1
         SubPlan 1
           ->  Index Scan using importer_treeimportrow_import_event_id on importer_treeimportrow tir  (cost=0.43..241.36 rows=201 width=4) (actual time=0.006..0.006 rows=0 loops=1)
                 Index Cond: (import_event_id = 1)
                 Filter: (plot_id IS NOT NULL)
 Planning time: 0.559 ms
 Execution time: 0.307 ms
(12 rows)
```

I think I will go with adding a bbox filter to the query.

jwalgran:
Oh. I got some "tunnel vision" and did not pay attention to @RickMohr 's actual suggestion of synthesizing a 10ft query polygon and filtering on the intersection. That would vastly simplify things. I will try that out.

RickMohr:
Query should keep its "ignore all ID's in the import" clause

jwalgran:
Superseded by #2285, which avoids the use of raw SQL and provides the same performance boost.

PR #2270, opened 2015-09-16
Cleanup styling for small screens.

maurizi:
Fixes a styling issue with the header bar on the comment moderation page
as well as a styling issue on the global navbar.


jwalgran:
I follow the changes in size classes and padding. What does removing the overflow from the .map class at small widths do for us?

:+1:

maurizi:
@jwalgran I saw an unrelated responsive styling bug while I was fixing some other things, where the map div had scrollbars on it.  Removing `overflow: auto` fixed it and didn't seem to cause any other issues.  ShouldI update the commit message to reflect this?

jwalgran:
A good idea for future ``git blame``ers

PR #2271, opened 2015-09-17
Fix collectstatic sometimes missing updated files.

maurizi:
Django's `collectstatic` command uses file modification time to determine
if it should overwrite a file.  Unfortunately inside VirtualBox shared
folder file modification times are often inconsistent, and this can lead
to newly updated files not being collected, especially when running
`grunt watch`.

Adding `-c` to the command tells Django to always overwrite every file,
which fixes the problem, though it does add a slight amount of overhead.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -100,7 +100,7 @@ module.exports = function(grunt) {
         },
         shell: {
             collect_static: {
-                command: 'python manage.py collectstatic --noinput'
+                command: 'python manage.py collectstatic --noinput -c'
             }
         },
         browserify: {


jwalgran:
Could you quantify "slight?" The edit/build/test loop is already longer than I would like it to be. I am not sure if we have a choice, since we need a collectstatic to do the right thing, but maybe trying to attack the problem through VirtualBox (NFS shares?) could help avoid the performance penalty.

maurizi:
I ran `time ./scripts/manage.sh collectstatic --noinput` several times both with and without the `-c` option with no files that needed updating, which seems like a best case baseline.

Without `-c`: 
`./scripts/manage.sh collectstatic --noinput  1.98s user 0.44s system 40% cpu 6.011 total`
`./scripts/manage.sh collectstatic --noinput  1.92s user 0.49s system 40% cpu 5.987 total`
`./scripts/manage.sh collectstatic --noinput  1.97s user 0.52s system 43% cpu 5.734 total`

With `-c`:
`./scripts/manage.sh collectstatic --noinput -c  2.04s user 0.54s system 36% cpu 6.986 total`
`./scripts/manage.sh collectstatic --noinput -c  1.86s user 0.54s system 36% cpu 6.668 total`
`./scripts/manage.sh collectstatic --noinput -c  2.09s user 0.53s system 37% cpu 6.952 total`

So about 1 second longer give or take.

As far as how to improve things without taking a performance hit, I can think of a few options:
 - Refactor our JS build scripts to move any _changed_ bundled assets into the `STATIC_ROOT` directly, so we don't have to involve Django or `collectstatic`.
 - Refactor our JS build scripts to place compiled assets into a directory outside of the VirtualBox shared folder, and add that directory to the list of places Django looks for static JS files, which should prevent VirtualBox from messing with the file modification times.

RickMohr:
+1 Works.

PR #2272, opened 2015-09-18
Adds rate limiting to importer commit tasks to prevent database overutilization

maurizi:
Based on the timing information reported in OpenTreeMap/otm-core#2267
we expect a single batch of 250 trees running on a single worker to take
approximately 1 minute per batch, once the distance validation refactor
in that PR is integrated.

Based on this, I reduced the batch size to 85 which should take around 20
seconds, and I added a rate limit of 1 batch per minute.  Since we have
two Celery workers in production, this should provide a combined rate
limit of 2 batches/minute (since rate limits affect workers individually),
with an expected "quiet period" of 20 seconds per minute.

Since our production environment differs significantly from our
development environments, we should obviously test these assumptions
and timings after this change is deployed to production.

Connects to #2258

### DIFF
--- a/opentreemap/importer/tasks.py
+++ b/opentreemap/importer/tasks.py
@@ -7,6 +7,7 @@ import json
 
 from celery import task, chord
 from django.core.exceptions import ObjectDoesNotExist
+from django.conf import settings
 from django.db import transaction
 
 from treemap.models import Species
@@ -17,8 +18,6 @@ from importer.models.trees import TreeImportEvent, TreeImportRow
 from importer import errors, fields
 from importer.util import (clean_row_data, clean_field_name,
                            utf8_file_to_csv_dictreader)
-
-BLOCK_SIZE = 250
 
 
 def _create_rows_for_event(ie, csv_file):
@@ -58,7 +57,8 @@ def _create_rows(ie, reader):
         rows.append(RowModel(data=data, import_event=ie, idx=idx))
 
         idx += 1
-        if int(idx / BLOCK_SIZE) * BLOCK_SIZE == idx:
+        if ((int(idx / settings.IMPORT_BATCH_SIZE) *
+             settings.IMPORT_BATCH_SIZE == idx)):
             RowModel.objects.bulk_create(rows)
             rows = []
 
@@ -92,8 +92,9 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
 
     try:
         row_set = ie.rows()
-        validation_tasks = (_validate_rows.subtask(row_set[i:(i+BLOCK_SIZE)])
-                            for i in xrange(0, ie.row_count, BLOCK_SIZE))
+        validation_tasks = (
+            _validate_rows.subtask(row_set[i:(i+settings.IMPORT_BATCH_SIZE)])
+            for i in xrange(0, ie.row_count, settings.IMPORT_BATCH_SIZE))
 
         final_task = _finalize_validation.si(import_type, import_event_id)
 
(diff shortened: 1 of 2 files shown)


maurizi:
This can be merged before https://github.com/OpenTreeMap/otm-core/pull/2267, but it will likely not have any affect (positive or negative) until that PR is merged.

jwalgran:
What do you think about putting the rate limit in a setting that would let us use different values for different deployments, or more easily change it in a production deploy when we need or want to temporarily slow things down?

maurizi:
@jwalgran I like the idea.  I also pulled `BLOCK_SIZE` out into the Django settings as well.

jwalgran:
Sorry for sitting on this. I wanted to get the proximity check optimization in place before checking out the timings. Could you please rebase this on top of the latest master?

maurizi:
@jwalgran It's been rebased.

jwalgran:
Launched a single celery process on my VM and committed 1000 trees to a map with 500,000 existing trees. I see mostly the expected output. 1m between tasks, except for the first two.

```
2015-09-30 13:21:37,216 tasks._commit_rows succeeded in 18.751179653s: None
2015-09-30 13:21:37,243 tasks._commit_rows succeeded in 18.797136631s: None
2015-09-30 13:22:35,057 tasks._commit_rows succeeded in 15.198042358s: None
2015-09-30 13:23:34,928 tasks._commit_rows succeeded in 14.997731673s: None
2015-09-30 13:24:35,786 tasks._commit_rows succeeded in 15.770342432s: None
2015-09-30 13:25:35,374 tasks._commit_rows succeeded in 15.30674078s:  None
2015-09-30 13:26:33,981 tasks._commit_rows succeeded in 15.162252383s: None
2015-09-30 13:27:34,496 tasks._commit_rows succeeded in 15.608256426s: None
2015-09-30 13:28:35,114 tasks._commit_rows succeeded in 16.163152147s: None
2015-09-30 13:29:34,499 tasks._commit_rows succeeded in 15.478169244s: None
2015-09-30 13:30:30,532 tasks._commit_rows succeeded in 11.446474297s: None
2015-09-30 13:31:34,233 tasks._commit_rows succeeded in 15.745661248s: None
```

I launched 2 celery tasks and committed another 1000 trees. On both workers I saw the same unexpected pair of initial tasks, then expected behavior.

```
# WORKER A
2015-09-30 14:14:37,237 tasks._commit_rows succeeded in 41.385178064s: None
2015-09-30 14:14:37,543 tasks._commit_rows succeeded in 41.740854767s: None
2015-09-30 14:15:10,871 tasks._commit_rows succeeded in 14.91725806s: None
2015-09-30 14:16:12,004 tasks._commit_rows succeeded in 15.953025384s: None
2015-09-30 14:17:11,518 tasks._commit_rows succeeded in 15.373559372s: None
2015-09-30 14:18:12,443 tasks._commit_rows succeeded in 16.227509245s: None
```

```
# WORKER B
2015-09-30 14:14:37,313 tasks._commit_rows succeeded in 41.522724785s: None
2015-09-30 14:14:38,112 tasks._commit_rows succeeded in 42.295789376s: None
2015-09-30 14:15:12,392 tasks._commit_rows succeeded in 16.571708086s: None
2015-09-30 14:16:11,789 tasks._commit_rows succeeded in 15.953622956s: None
2015-09-30 14:17:08,562 tasks._commit_rows succeeded in 12.718369406s: None
```

I would like to investigate why we are seeing this doubling up of tasks in the log output, but I don't think it should block merging this config.

:+1:

RickMohr:
This looks like it will cause a 5-fold slowdown in imports (15 sec per batch, then wait a minute). If those timings reflect what we'll see in production we might consider lowering the `rate_limit` to around 15 sec.

maurizi:
@RickMohr That's not how Celery rate limiting works.  Setting a `rate_limit` of  15 seconds for a task that could take between 15-45 seconds would mean that tasks would *never* wait.  When you set a rate-limit it sets the _maximum_ amount of time a task will wait, so in this case we're setting a `rate_limit` of `1/m`, and so for a tasks that takes 15 sec. it will wait 45 sec., for a task that takes 40 sec. it will wait 20 sec., etc.  For a task that takes longer than a minute, there will be no wait time.

Also, since the rate limits apply to each worker individually, and we have 2 Celery workers, we should see 2 batches per minutes, not 1.

I do think we may want to reevaluate our batch size and rate limiting after seeing how some production imports turn out, but for now this should be fast enough to import 10,000 trees an hour (w/ 85 per batch, 2 batches per minute).

PR #2278, opened 2015-09-23
Feature/remove featureset

steventlamb:


steventlamb:
reopening against `modular-pricing` https://github.com/OpenTreeMap/otm-core/pull/2279

PR #2279, opened 2015-09-23
Rename references to tier

steventlamb:


jwalgran:
+1 did not test. Grepped across all repos for `tier_tree_limit` and did not find any python or template references.

steventlamb:
Added two commits that cap off the 'remove featureset' task

jwalgran:
I ran this branch locally and connected to it with the iOS application running in the simulator and verified that all publicly accessible instances appeared in the list.

:+1:

PR #2281, opened 2015-09-24
Link to static importer help files

jwalgran:
The old links are no longer valid. Using Django static file URLs gives us flexibility on how, and from where, the help documentation and templates are served.


steventlamb:
+1 tested

PR #2282, opened 2015-09-24
Add missing terminology

steventlamb:
It was originally planned to dynamically generate plural terminology if
it was absent but it was then decided to be more complex than
necessary. Plural terminology was supposed to be added in when that
decision was made, but fell through for a few models.

### DIFF
--- a/opentreemap/stormwater/models.py
+++ b/opentreemap/stormwater/models.py
@@ -137,6 +137,10 @@ class RainBarrel(MapFeature):
         'plural': _('Rain Barrels'),
     }
 
+    @classproperty
+    def benefits(cls):
+        return CountOnlyBenefitCalculator(cls)
+
     @property
     def is_editable(self):
         # this is a holdover until we can support editing for all resources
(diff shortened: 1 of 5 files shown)


steventlamb:
I threw another commit on here that is a little bit off topic.

jwalgran:
I tried testing this by hacking in a new plural value and did not see it working because this block is naïvely using `singular` and appending an "s." We can use the new plural support now.

https://github.com/OpenTreeMap/otm-core/blob/master/opentreemap/treemap/models.py#L609-L610

```python
        if not self.instance.bounds.contains(self.geom):
            raise ValidationError({
                "geom": [
                    _(
                        "%(model)ss must be created inside the map boundaries")
                    % {'model': self.terminology(self.instance)['singular']}]
            })
```

jwalgran:
I confirmed that there are no errors or missing terminology text.

I also confirmed that adding a Rain Barrel causes the resource count in the header to increase.

:+1:

Issue #2283, opened 2015-09-24
Update to Django 1.8.4


jwalgran:

jwalgran:
I misread the release notes. 1.8.5 is still in development.

PR #2286, opened 2015-09-30
Upgrade from django 1.8.2 to 1.8.4

jwalgran:
1.8.3 and 1.8.4 include security fixes.

https://docs.djangoproject.com/en/1.8/releases/1.8.3/
https://docs.djangoproject.com/en/1.8/releases/1.8.4/

Connects to #2283


maurizi:
+1

PR #2287, opened 2015-10-01
Refresh finished import table on completion

jwalgran:
This fixes the bug where a completed import would disappear because the active import table refreshed, but the finished import table did not.

I fixed the issue by keeping track of which tables are linked together, and the current state of the primary tables, and then refreshing the linked table when the primary table transitions out of the refreshing state.


jwalgran:
@RickMohr I simplified the refreshing as we discussed, and changed the `has_pending` check to look across all pages.

RickMohr:
Thanks, much simpler! I clicked "add to tree map" for a verified import, went to page 2 so its updates weren't visible, and saw it appear in the "Finished" table (...eventually, slow with the rate limiting).

:+1:

PR #2288, opened 2015-10-02
Refactor get_cached_plot_count to make it easier to use.

maurizi:
Removes the need to pass `compute_value` to get_cached_plot_count and
adds a helper method to Instance to get the total number of plots from the
cache.


RickMohr:
+1 with comment consideration

maurizi:
@RickMohr They've both been rebased.

Issue #2142, opened 2015-07-13
Fix warnings uncovered by Django 1.8 migration


maurizi:
Running `manage.py runserver` or `manaqge.py check` shows the following warnings:
```
WARNINGS:
treemap.Instance.boundaries: (fields.W340) null has no effect on ManyToManyField.
treemap.Instance.users: (fields.W340) null has no effect on ManyToManyField.
```

PR #2294, opened 2015-10-09
Eliminate `fields.W340` warnings

RickMohr:
No need for `null=True` on many-to-many fields `Instance.users` and `Instance.boundaries`

Connects #2142


maurizi:
+1,000

PR #2295, opened 2015-10-12
Improvements to inline edit forms

RickMohr:
* Clicking "Save" disables "Save" and "Cancel" and shows spinner (if present)
* Pages configuring an inline edit form may omit options by following conventions
* New template partial allows DRY buttons

### DIFF
--- a/opentreemap/treemap/templates/treemap/partials/form_buttons.html
+++ b/opentreemap/treemap/templates/treemap/partials/form_buttons.html
@@ -1,0 +1,14 @@
+{% load i18n %}
+
+<div class="col-xs-4 text-right">
+    <img class="spinner" src="{{ STATIC_URL }}img/spinner.gif" style="display: none;">
+    <button data-class="display" class="btn btn-info editBtn">
+        {% trans "Edit" %}
+    </button>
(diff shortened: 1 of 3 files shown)


opentreemap/treemap/templates/treemap/partials/form_buttons.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+
+<div class="col-xs-4 text-right">
+    <img class="spinner" src="{{ STATIC_URL }}img/spinner.gif" style="padding: 0 5px 5px 0; display: none;">
```
maurizi: Since there is already a `spinner` class on here, consider moving the inline `padding` styling out into SASS.

RickMohr: There are many uses of the `spinner` class, and changing the SASS isn't compatible with all of them. This spinner needs some design attention anyway since it's on a dark background, so I thought the designer could sort it out.

RickMohr:
@maurizi, I switched to SASS styling as you suggested.

maurizi:
+1

PR #2296, opened 2015-10-13
Restore mistakenly-removed saveStream option to inlineEditForm.init

RickMohr:


jwalgran:
I confirmed that that I was unable to save credit card information on the Billing Info management page and then, after checking out this branch, I was once again able to save changes.

:+1:

PR #2297, opened 2015-10-13
Feature/gsi opt out

steventlamb:


maurizi:
With this change, when I opted out of Bioswales, I still saw Bioswale photos on the `/edits` page, but Bioswale edits and Bioswale Stewardship edits were gone.

maurizi:
When I opted out of Bioswales, Bioswales were still an option in the advanced search stewardship search.

opentreemap/treemap/views/map_feature.py
```diff
@@ -111,9 +111,8 @@ def context_map_feature_detail(request, instance, feature_id, **kwargs):
 
 
 def map_feature_photo_detail(request, instance, feature_id, photo_id):
-    photo = get_object_or_404(MapFeaturePhoto,
-                              pk=photo_id,
-                              map_feature=feature_id)
+    feature = get_map_feature_or_404(feature_id, instance)
+    photo = get_object_or_404(MapFeaturePhoto, pk=photo_id, feature=feature)
```
maurizi: Here it is

maurizi:
Other than the comment about `feature` needing to be `map_feature`, I think this looks good.

+1 after fixup commits are rebased away.

PR #2298, opened 2015-10-13
Fix error validation for species columns with numeric values.

maurizi:
There was a "hack" in place to support changing the species values in a
tree import to an existing species. When updating a species, the 'species'
column was updated to reflect the PK of a species in the database, and
then later during validation the species related fields (including the one
which held the PK) would be overwritten with the correct info from the
database if the 'species' field was a number

Because it piggybacked the species id in the species name column of the
ImportRow object, there were conflicts when a number was uploaded as the
value for the species column in the original CSV, and no error would be
reported to the user.

I cleaned up the work-around significantly, by sending the species id in
a different field name, and fetching the species information from the DB
as part of the view function, so that there is no need for validation to
be concerned with it.


steventlamb:
code looks good. I can confirm that with bad species data, this stops the importer from crashing and makes species correction possible. +1

PR #2299, opened 2015-10-14
Expose inlineEditForm functions

jwalgran:
This allows code outside the context of the inlineEditForm
to serialize the state of the inlineEditForm and push validation errors into the inlineEditForm for display. 

The use case driving this change is wanting to "preview" the results of a change, which could return the same server-side validation errors that the actual save operation would generate.

Another potential use case would be adding client side validations that could make use of the existing error display plumbing.


RickMohr:
+1

PR #2302, opened 2015-10-19
Correctly use model display names on the Edits page.

maurizi:
Also adds a helper method to get the singular terminology name more
easily, and improves the display_name template filter to take an
(optional) instance parameter to use to look up the display name.

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -22,7 +22,8 @@ from django.db import IntegrityError, connection, transaction
 from django.conf import settings
 
 from treemap.units import (is_convertible, is_convertible_or_formattable,
-                           get_display_value, get_units, get_unit_name)
+                           get_display_value, get_units, get_unit_name,
+                           Convertible)
 from treemap.util import (all_models_of_class, leaf_models_of_class,
                           to_object_name)
 
@@ -1391,11 +1392,18 @@ class Audit(models.Model):
 
         format_string = cls.action_format_string_for_audit(self)
 
-        model_display_name = cls.terminology(self.instance)['singular']
+        model_display_name = self.model_display_name()
 
         return format_string % {'field': self.field_display_name,
                                 'model': model_display_name.lower(),
                                 'value': self.current_display_value}
+
+    def model_display_name(self):
+        cls = get_auditable_class(self.model)
+        if issubclass(cls, Convertible):
+            return cls.display_name(self.instance)
+        else:
+            return cls.__name__
 
     def dict(self):
         return {'model': self.model,
--- a/opentreemap/treemap/models.py
+++ b/opentreemap/treemap/models.py
@@ -697,7 +697,7 @@ class MapFeature(Convertible, UDFModel, PendingAuditable):
             else:
                 title = _("Empty Planting Site")
         else:
-            title = feature.terminology(self.instance)['singular']
+            title = feature.display_name(self.instance)
 
         return title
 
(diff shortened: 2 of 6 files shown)


opentreemap/treemap/models.py
```diff
@@ -1071,6 +1071,12 @@ def delete_with_user(self, *args, **kwargs):
         thumb.delete(False)
         image.delete(False)
 
+    @classmethod
+    def display_name(cls, instance):
```
steventlamb: Both Tree and Plot use the `_terminology` API despite also being not configurable. I don't see a reason to introduce this layer on top of terminology, but if we do, we should convert Tree and Plot as well.

opentreemap/treemap/units.py
```diff
@@ -31,6 +31,10 @@ def terminology(cls, instance=None):
                          .get(cls.__name__, {}))
         return terms
 
+    @classmethod
+    def display_name(cls, instance):
```
steventlamb: the `display_name` method was phased on on several models in favor of using the terminology API.

https://github.com/steventlamb/otm-core/commit/286fe6c85e295607dc6acd360b407fe2ba420b7f#diff-eb9b8ea2fa5bef48515b23b723bbd6d4
https://github.com/steventlamb/otm-core/commit/d7a866b12e1dd9174dbc49607e0ace751e938783#diff-eb9b8ea2fa5bef48515b23b723bbd6d4
https://github.com/steventlamb/otm-core/commit/286fe6c85e295607dc6acd360b407fe2ba420b7f#diff-eb9b8ea2fa5bef48515b23b723bbd6d4

maurizi: As we discussed, I'd like to leave this in as a helper method, to make it easier to get the singular name.

maurizi:
@steventlamb I added the `Convertible` mixin to `MapFeaturePhoto` and I'm using the `_terminology` config for it, as you suggested.

steventlamb:
```
src/otm-core/opentreemap/treemap/lib/map_feature.py:31:51:            feature.terminology(feature.instance)['singular'].lower())
```
should be `feature.display_name(feature.instance).lower()`, I think.

steventlamb:
also:
```
src/otm-core/opentreemap/treemap/models.py:700:56:            title = feature.terminology(self.instance)['singular']
```

opentreemap/treemap/audit.py
```diff
@@ -1391,12 +1391,19 @@ def short_descr(self):
...
-        model_display_name = cls.terminology(self.instance)['singular']
+        model_display_name = self.model_display_name()
 
         return format_string % {'field': self.field_display_name,
                                 'model': model_display_name.lower(),
                                 'value': self.current_display_value}
 
+    def model_display_name(self):
+        cls = get_auditable_class(self.model)
+        if hasattr(cls, 'display_name'):
```
steventlamb: consider using `issubclass(cls, Convertible)`. I'm slightly concerned that because we have a `display_name` method on `Species` with a different signature, it could produce a tricky bug in the future.

opentreemap/treemap/templatetags/util.py
```diff
@@ -109,18 +110,18 @@ def display_name(audit_or_model_or_name):
...
-    # I am deferring this fix because this feature is not planned asnd it's
-    # non-trivial to put instance in scope here, which is required for
-    # getting terminology
-    if name.lower() == 'plot':
-        return Plot.terminology()['singular']
+    if audit:
+        return audit.model_display_name()
+
+    cls = safe_get_model_class(name)
+    if hasattr(cls, 'display_name'):
```
steventlamb: (same, consider using `issubclass`)

steventlamb:
+1

PR #2304, opened 2015-10-20
Add the unlocalize template filter to PKs used in Django templates.

maurizi:
I went through the source and grepped for `'(id|pk) ?}}'` and added an
`unlocalize` filter to everywhere it seemed appropriate.  I'm sure there
are other places where the template variables are named differently that
I've missed, but this should prevent a slew of bugs waiting to happen.

### DIFF
--- a/opentreemap/treemap/templates/treemap/field/species_typeahead.html
+++ b/opentreemap/treemap/templates/treemap/field/species_typeahead.html
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load l10n %}
 {% load instance_config %}
 {% load species_thumbprint %}
 
@@ -6,7 +7,7 @@
        data-typeahead-hidden
        id="{{ field.label }}-hidden"
        name="{{ field.identifier }}"
-       value="{{ field.value.id }}"/>
+       value="{{ field.value.id|unlocalize }}"/>
 <input type="text"
        data-typeahead-input="{{ field.identifier }}"
        id="{{ field.label }}-typeahead"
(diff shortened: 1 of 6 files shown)


opentreemap/importer/templates/importer/partials/merge_species.html
```diff
@@ -21,7 +21,7 @@
                             <input type="radio"
                                    id="{{ value.id }}"
                                    name="{{ field.id }}"
-                                   value="{{ value.value }}"
+                                   value="{{ value.value|unlocalize }}"
```
jwalgran: If I am looking in the right place, this `value` is not a number

screen shot 2015-10-21 at 7 56 09 am

opentreemap/treemap/templates/treemap/edits.html
```diff
@@ -72,7 +72,7 @@
           </td>
           <td>
             {% if audit.model|is_filterable_audit_model %}
-            <a href="?models={{ audit.model }}&model_id={{ audit.model_id }}">
+            <a href="?models={{ audit.model }}&model_id={{ audit.model_id|unlocalize }}">
```
jwalgran: This is not a problem on `master` because of the `{% localize off %}` block above. It does not look like it does any harm to add this, though.

opentreemap/importer/templates/importer/partials/merge_species.html
maurizi: Good catch! from looking at the view function I thought it was the species id, but clearly I was wrong. I'll remove this.

jwalgran:
It looks like that species merge one is the only problematic one.

:+1:

PR #2305, opened 2015-10-20
highlight color intensity fix and notification number lineheight fix

designmatty:


designmatty:
fix for https://github.com/OpenTreeMap/otm-addons/issues/770

and fix for https://github.com/OpenTreeMap/otm-addons/issues/838

opentreemap/treemap/css/sass/partials/pages/_admin.scss
```diff
@@ -256,7 +257,8 @@
             border-bottom: 1px solid #EEE;
 
             &.active {
-                background-color: #F7F3DF;
+                background-color: $warning-color;
+                background-color: rgba($warning-color, .75);
```
maurizi: Did you intend to include `background-color` here twice? An explanatory comment might be helpful if this is intended.

designmatty: Yeah I did. line 260 is a fallback for ie8 which I wasn't sure if we supported which I just asked Deb. She said not to worry about it but it also can't hurt to have it in place.

maurizi: Yeah, we don't even support IE9 :smile: Can't hurt though...

designmatty: Yeah that's good to know hah!

maurizi:
+1

PR #2306, opened 2015-10-20
Improve RainBarrel.capacity error messages

steventlamb:


RickMohr:
I can't make the error message appear, either on `ec2-otm-dev` or with this PR. I see it in the network tab when I click "Done" but it is not displayed. What am I missing?

RickMohr:
+1 I see that the message is correctly returned from the server. Yay using concise Django conventions. We need a separate issue for the error message not displaying on some machines.

steventlamb:
@RickMohr I also confired this fix on the mf detail page:
screenshot from 2015-10-20 16 10 22

PR #2307, opened 2015-10-20
Update fontello icons.

maurizi:
The only icon that changed is icon-tree, which is now filled in instead
of an outline, matching the marketing site.


jwalgran:
I verified that `icon-tree` is now filled and that our icon heavy sections (map page eco and management left sidebar) don't have any missing icons.

:+1:

PR #2309, opened 2015-10-20
optimized spinners for both dark and light backgrounds

designmatty:
Should fix https://github.com/OpenTreeMap/otm-addons/issues/810


designmatty:
This is just a suggestion @RickMohr but I think we should use an svg with a gif fallback instead of just using a gif. Let me know if this is something worth doing.

maurizi:
The new spinners are _bigger_ than the ones they are replacing, and overflow their surrounding areas.  I think we need to have the new images match the size of the existing ones.

designmatty:
I will fix this by the end of day friday.

maurizi:
+1

Issue #2303, opened 2015-10-19
Cached counts continue to show map feature types removed from instance until tree added


steventlamb:

steventlamb:
to reproduce:
1) create an instance
2) add bioswales to map feature types
3) add a bioswale
4) remove bioswales from the instance's map feature types

At this point the tiler will correctly hide the bioswale but the counts will register "1 resource" until you add a tree.

PR #2312, opened 2015-10-22
Include the instance map feature types in eco etags.

maurizi:
This ensures that changing the map features for the instance invalidates
the browser cached resource counts.

Connects to #2303


jwalgran:
Looking at this now

jwalgran:
:+1:

PR #2319, opened 2015-10-23
fixed bottom padding on admin page overflows

designmatty:
Fixes issue reported on add-ons repo https://github.com/OpenTreeMap/otm-addons/issues/843

Firefox now has correct padding-bottom. Every other browser will have slightly bigger padding-bottom due to this *fix*. It still looks good though


maurizi:
> Every other browser will have slightly bigger padding-bottom due to this fix. It still looks good though
Should we reduce the padding by `16px` then so every browser will look consistent?

designmatty:
That would be logical ha. One sec.

maurizi:
+1

PR #2320, opened 2015-10-23
clean up the suspend map page

designmatty:
connected to https://github.com/OpenTreeMap/otm-addons/pull/875

### DIFF
--- a/opentreemap/treemap/css/sass/partials/pages/_admin.scss
+++ b/opentreemap/treemap/css/sass/partials/pages/_admin.scss
@@ -317,7 +317,6 @@
 
     .alert{
         padding: 15px;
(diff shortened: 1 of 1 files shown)


designmatty:
This fixes the margin-top of the alerts across all management pages.

RickMohr:
Do you know if other uses of `.alert` on management pages continue to look good with this change?

designmatty:
I checked on 3-4 other management pages including those that have 2 alerts in a row and they all looked fine. The margin was added months ago when Christopher first did his work so I assume the reason for the margin no longer exists.

RickMohr:
+1 thanks for checking out the other uses.

PR #2323, opened 2015-10-26
 Rename 'TILE_HOST' -> 'TILER_HOST'

steventlamb:
'Tiler' is more conventional than 'Tile' and more consistent with other uses.


RickMohr:
The original name seems fine to me, but if you change it I think you should also change `otm.settings.tileHost`.

steventlamb:
good idea, done.

PR #2327, opened 2015-10-28
Refresh Instances after incrementing the adjuncts_timestamp

maurizi:
Prevents caching issues where related objects were saved before the Instance
object was saved, causing the adjuncts_timestamp to be reverted back to the old
value.

As an example of what this prevents, consider the following code snippet:

```
role = Role.objects.all()[0]  # role.instance.adjuncts_timestamp is 1

role.perm_level = 2
role.save()  # This triggers `increment_adjuncts_timestamp`, increasing it to 2

role.instance.save()  # This reverts the `adjuncts_timestamp` to 1
```

Connects to OpenTreeMap/otm-addons#891
Connects to OpenTreeMap/otm-addons#890

### DIFF
--- a/opentreemap/treemap/lib/object_caches.py
+++ b/opentreemap/treemap/lib/object_caches.py
@@ -69,6 +69,7 @@ def increment_adjuncts_timestamp(instance):
     from treemap.models import Instance
     Instance.objects.filter(pk=instance.id)\
                     .update(adjuncts_timestamp=F('adjuncts_timestamp') + 1)
+    instance.refresh_from_db()
 
 
 # ------------------------------------------------------------------------


RickMohr:
+1
I was unable to reproduce this issue with or without your change. However, the change seems good. It would be possible to write code that defeats this fix, but I don't expect them to arise in practice.

opentreemap/treemap/lib/object_caches.py
```diff
@@ -69,6 +69,7 @@ def increment_adjuncts_timestamp(instance):
     from treemap.models import Instance
     Instance.objects.filter(pk=instance.id)\
                     .update(adjuncts_timestamp=F('adjuncts_timestamp') + 1)
+    instance.refresh_from_db()
```
RickMohr: @maurizi, on further reflection I think it would be safer to do:
``` python
    qs = Instance.objects.filter(pk=instance.id)
    qs.update(adjuncts_timestamp=F('adjuncts_timestamp') + 1)

    # Fetch updated timestamp so it's not stale
    instance.timestamp = qs[0].timestamp
```

Otherwise we risk clobbering values in `instance` that haven't been saved yet.

PR #2328, opened 2015-10-28
Improve timestamp refreshing in object_caches.

maurizi:
Commit 6222295e fixed a problem with accidentally saving stale
adjuncts_timestamp values, but @RickMohr pointed out that it could cause
changes to be missed.  Only refreshing the timestamp but keeping other data in
the instance is a bit safer while still preventing the original problem.


RickMohr:
+1, still seems to work!

PR #2329, opened 2015-10-29
Only show plots from the current instance on the map feature detail page

maurizi:
Connects to OpenTreeMap/otm-addons#900


RickMohr:
+1 (did not test.) I did do a quick scan for other uses of `Plot.objects` and saw no similar issues.

PR #2338, opened 2015-11-04
Improve helpers

RickMohr:
* `buttonEnabler.js` no longer requires `options.config` argument
* Add JS Utility `modalsFocusOnFirstInputWhenShown`
* Allow adding buttons to `form_buttons.html` partial


maurizi:
+1

Issue #1990, opened 2015-04-06
Export from map with single planting site fails with "Query returned no results"


jwalgran:
1. Create a map
1. Take the map out of trial mode
1. Add a planting site with no information other than location
1. Refresh the page
1. Click export search results

Expected: CSV with single row
Actual: "Query returned no results" message

![screen shot 2015-04-06 at 2 09 59 pm](https://cloud.githubusercontent.com/assets/17363/7012574/f1e6148a-dc66-11e4-92b2-10caff28f406.png)

jwalgran:
Adding a tree to the planting site makes it show up in the export.

PR #2346, opened 2015-11-09
Include empty plots in tree exports

RickMohr:
* Query `Plot` instead of `Tree`
* Adjust CSV headers to remain tree-centric instead of becoming plot-centric
* Use latest version of `django-queryset-csv` for necessary bug fix

Also fixes a longstanding bug where exports ignored "Display Filters"

Connects #1990

### DIFF
--- a/opentreemap/exporter/tasks.py
+++ b/opentreemap/exporter/tasks.py
@@ -15,7 +15,7 @@ from django.core.files import File
 from treemap.lib.object_caches import permissions
 
 from treemap.search import Filter
-from treemap.models import Species, Tree
+from treemap.models import Species, Plot
 from treemap.util import safe_get_model_class
 from treemap.audit import model_hasattr, FieldPermission
 from treemap.udf import UserDefinedCollectionValue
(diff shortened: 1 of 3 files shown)


RickMohr:
@steventlamb, I incorporated the change from #2348 into this PR and it all works for me. Ready for another look.

RickMohr:
waffle

steventlamb:
Code looks good and on testing, it now shows empty plots. However, instead of export `plot__geom__x` and `plot__geom__y`, it exports `plot__geom` in a seralized format (wkt?) that is probably not end-user friendly.

steventlamb:
When I go to advanced search and uncheck `show trees`, it reduces my counts accordingly but the export is still exporting plots that *do* have trees, inconsistent with the count.

RickMohr:
@steventlamb, the advanced search issue was a longstanding bug -- we weren't passing the `show` parameter to the export endpoint.

All fixed now if you want to take another look.

steventlamb:
+1, geoms look good and I tried a few different advanced search combinations and got correct export results.

PR #2348, opened 2015-11-10
Receive bugfixes from latest version of djqscsv

steventlamb:
see https://github.com/azavea/django-queryset-csv/issues/76 for more details


RickMohr:
+1 works with #2346

RickMohr:
I incorporated this change into #2346, so I think we can just close this.

PR #2349, opened 2015-11-10
Require config argument to button enabler once again

RickMohr:
07c776c introduced an error by making the `config` argument to `buttonEnabler` optional and eliminating the top-level `config` variable. This reverts the first change, but retains the second so the usage of `config` is clearer.

### DIFF
--- a/opentreemap/treemap/js/src/buttonEnabler.js
+++ b/opentreemap/treemap/js/src/buttonEnabler.js
@@ -33,9 +33,9 @@ function removeActionableDataAttributes($el) {
     $el.removeAttr('data-action');
 }
 
-function makeRedirectToLogin($el, href) {
+function makeRedirectToLogin(loginUrl, $el, href) {
 
-    var fullHref = config.loginUrl + href;
+    var fullHref = loginUrl + href;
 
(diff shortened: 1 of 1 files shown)


maurizi:
+1

steventlamb:
+1, didn't test

opentreemap/treemap/js/src/buttonEnabler.js
```diff
@@ -68,7 +68,7 @@ function fullyDisable($el, disabledTitle) {
 
 exports.run = function (options) {
     var $elements = $(enablePermSelector),
-        loggedIn = options && options.config ? options.config.loggedIn : true;
+        config = options.config;
```
jwalgran: We may want to protect against mistakenly forgetting to pass in `options` 

```javascript
config = options ? options.confg : {}
```

RickMohr: Merged before seeing your comment. Not passing in `options` was a momentary bad idea of mine; all uses of `buttonEnabler` supply `options.config` so future users aren't likely to omit it.

PR #2350, opened 2015-11-12
Photo review updates

maurizi:
- Adjust text on photo review page and add an "Archived" filter.
 - Add batch moderation UI to photo review page.

Connects to OpenTreeMap/otm-addons#926
Connects to OpenTreeMap/otm-addons#542

### DIFF
--- a/opentreemap/otm_comments/tests.py
+++ b/opentreemap/otm_comments/tests.py
@@ -155,7 +155,7 @@ class CommentReviewTest(CommentTestCase):
 
 
 def _comment_ids_to_params(*args):
-    return {'comment-ids': ','.join(str(arg) for arg in args)}
+    return {'ids': ','.join(str(arg) for arg in args)}
 
 
 def make_post_request(user, params=None):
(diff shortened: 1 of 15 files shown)


opentreemap/otm_comments/tests.py
```diff
@@ -155,7 +155,7 @@ def test_sorting(self):
 
 
 def _comment_ids_to_params(*args):
-    return {'comment-ids': ','.join(str(arg) for arg in args)}
+    return {'ids': ','.join(str(arg) for arg in args)}
```
steventlamb: curious about this rename. Based on it's uses/changes in this PR, it looks like the `comment-` prefix is useful for context.

steventlamb:
+1 total

opentreemap/otm_comments/tests.py
maurizi: I changed this to `ids` because the JS that calls this view function is now shared by photo and comment moderation, so it didn't make sense to call it `comment-ids` any more.

PR #2351, opened 2015-11-12
Feature/misc

steventlamb:


maurizi:
+1 after fixing the translated string to use named parameters

Issue #2342, opened 2015-11-05
Support new field types in advanced search UI


maurizi:
- Text fields
 - Choice fields
 - Multichoice fields

PR #2352, opened 2015-11-13
Prepare for custom field search

RickMohr:
Make sure front end and back ends can handle searching for custom field types:
* Text
* Choice
* Multichoice

Changes:
* `applyFilterObjectToDom` now sets values on dropdown (`select`) elements
* In advanced search UI, render multichoice fields as regular dropdowns

Requires OpenTreeMap/otm-tiler#87
Connects #2342

To test:
1) Create one custom field with each type (text, choice, multichoice)
2) Temporarily add them to the default advanced search fields (https://github.com/OpenTreeMap/otm-core/blob/2f54593390a668ce5e7a7643f03a9ef0f6778f28/opentreemap/treemap/instance.py#L291):
```python
{'identifier': 'tree.udf:MyTextField, 'search_type': 'LIKE'},
{'identifier': 'tree.udf:MyChoiceField, 'search_type': 'IS'},
{'identifier': 'tree.udf:MyMultiChoiceField, 'search_type': 'LIKE'},
```

Test advanced searches against these fields, making sure:
* Count is shown correctly
* Tree dots are shown correctly
* Refreshing the page preserves your search choices


RickMohr:
@maurizi, updated to handle the case of a multichoice field containing e.g. both "car" and "carp".

maurizi:
+1, works well.

PR #2353, opened 2015-11-18
Bug/fix search caching

steventlamb:


opentreemap/treemap/instance.py
```diff
@@ -526,21 +535,26 @@ def add_map_feature_types(self, types):
 
         add_default_permissions(self, models=classes)
 
+    def update_all_revs(self):
+        for rev in ('geo_rev', 'eco_rev', 'universal_rev'):
+            self._update_rev(rev)
```
jwalgran: Do we need to be concerned about making 3 SQL `UPDATE` queries one after another? We could write an `_update_revs` that takes a list of attributes and makes a single `qs.update` call.

opentreemap/treemap/instance.py
```diff
@@ -186,24 +186,29 @@ class Instance(models.Model):
...
+    The "geometry revision" is incremented whenever a map feature is
+    modified in a way that would affect rendered tiles for non-search
+    map rendering. Its value is part of all non-search tile URLs.
+
+    The "ecobenefit revision" is incremented whenever a tree is
+    modified in a way that would affect ecobenefit calculations.
+    summaries. Its value is part of cache keys for non-search
+    ecobenefit summaries.
     """
+    universal_rev = models.IntegerField(default=1)
```
jwalgran: Adding `null=True` here would make the migration backward compatible.

opentreemap/treemap/search.py
```diff
@@ -56,6 +56,9 @@ class Filter(object):
     def __repr__(self):
         return "(%s, %s)" % (self.filterstr, self.displaystr)
 
+    def __nonzero__(self):
+        return bool(self.filterstr) and bool(self.displaystr)
```
jwalgran: I would personally prefer a named `@property` instead of defining truthyness for a filter. We don't rely on type coercion much in this codebase so I can imagine myself scratching my head when `bool(filter)` returned `False` for a non-None object.

This isn't a nice looking, but it is not bad.:

```python
if filter and filter.is_set:
    ...
```

RickMohr:
+1 nice work! Don't forget "Connects #161".

PR #2354, opened 2015-11-19
Add ecobenefit data to CSV search exports

maurizi:
Connects to #2209


jmorrison1847:
+1 from Niagara --where was this left off?

PR #2355, opened 2015-11-19
Faster species export

RickMohr:
It was slow because:
* It used a large `Species.objects.` query
* `Species` subclasses `Auditable`, so its constructor calls `populate_previous_state`, which fetches UDFs
* That hits the object cache, which requires `instance.adjunct_timestamp`
* So `species.instance` was fetched from the database
* `select_related` and `prefetch_related` didn't help

At @maurizi's suggestion I rewrote it to use a values queryset, which was easy because
* Changing `SpeciesImportRow.SPECIES_MAP` to an `OrderedDict` made it easy to get the right field order
* The header row comes out right because `django-queryset-csv` removes underscores

### DIFF
--- a/opentreemap/importer/models/species.py
+++ b/opentreemap/importer/models/species.py
@@ -4,6 +4,9 @@ from __future__ import unicode_literals
 from __future__ import division
 
 import itertools
+
+from collections import OrderedDict
+
 from django.contrib.gis.db import models
 from django.db import transaction
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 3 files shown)


maurizi:
Looks good to me

RickMohr:
waffle

RickMohr:
unwaffle

RickMohr:
@maurizi, ready for another look.

maurizi:
+1, tested

PR #2358, opened 2015-11-23
Switch from 'help_text' to 'verbose_name' for field name defaults.

maurizi:
`verbose_name` is the more correct Django-ism to use for a field name.

Also removes a few errant uses of Plot in favor of Planting Site


RickMohr:
+1

PR #2361, opened 2015-11-30
Include correct protocol in activation email template.

maurizi:
Passes the request object into the activation email template, so it can
be used to template the activation link with either HTTP or HTTPS,
depending on whether the signup page was accessed over HTTPS.

Connects to https://github.com/OpenTreeMap/otm-cloud/issues/44


hectcastro:
:+1:

PR #2365, opened 2015-12-02
Make advanced search work with config from DB

maurizi:
- Removes search_type from the advanced search configuration dict
Instead, the search_type is inferred based on the underlying field type when
displaying fields.


RickMohr:
+1

Issue #2364, opened 2015-12-02
Remove extra items in Google Street View box


dboyer:
Looks like the Google Street View box changed again and the amount of overlay material (info box, directional, etc) makes it difficult to actually see the image. Can we minimize the info box and perhaps remove the circular rotation and full screen boxes?
![streetview](https://cloud.githubusercontent.com/assets/516764/11541141/9de66678-98fe-11e5-87c1-c5be14fbf07a.png)

jwalgran:
It looks like there are some more options that we pass to the `StreetViewPanorama` constructor to disable some widgets

Example from https://developers.google.com/maps/documentation/javascript/examples/streetview-controls

```javascript
 var panorama = new google.maps.StreetViewPanorama(
      document.getElementById('map'), {
        position: {lat: 42.345573, lng: -71.098326},
        addressControlOptions: {
          position: google.maps.ControlPosition.BOTTOM_CENTER
        },
        linksControl: false,
        panControl: false,
        enableCloseButton: false
  });

PR #2368, opened 2015-12-03
Hide address and panning controls on street view.

maurizi:
I attempted to hide the full screen button but as far as I can tell
there is no option to do so.

There is also no way to minimize the address info box, only hide it or
change its position, so I opted to always hide it.

Connects to #2364

### DIFF
--- a/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss
+++ b/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss
@@ -98,10 +98,14 @@
     border-radius: 9px;
 }
 
-// Needed because bootstrap breaks street view on firefox
-// by setting max-width: 100% on all <img> tags
-.street-view-small img {
-    max-width: none;
+.street-view-small {
+    height: 440px;
+
+    // Needed because bootstrap breaks street view on firefox
+    // by setting max-width: 100% on all <img> tags
+    .street-view-small img {
+        max-width: none;
+    }
 }
 
 #sidebar {
--- a/opentreemap/treemap/js/src/streetView.js
+++ b/opentreemap/treemap/js/src/streetView.js
@@ -32,7 +32,8 @@ exports.create = function(options) {
                 if (panorama === null) {
                     panorama = new google.maps.StreetViewPanorama(div, {
                         position:pos,
-                        addressControl: showAddress
+                        addressControl: showAddress,
+                        panControl: false,
                     });
                 } else {
                     panorama.setPosition(pos);
--- a/opentreemap/treemap/js/src/addMapFeature.js
+++ b/opentreemap/treemap/js/src/addMapFeature.js
@@ -151,8 +151,8 @@ function init(options) {
                 container = streetView.create({
                     streetViewElem: $streetViewContainer[0],
                     noStreetViewText: options.config.trans.noStreetViewText,
-                    hideAddress: true,
-                    location: latlng
+                    location: latlng,
+                    hideAddress: true
                 });
             }
 


dboyer:
I'm somewhat concerned about always hiding the address as that info is really helpful if you're using the street view to navigate down the street or if the street view is at a location near but not exactly at the address of the tree.

maurizi:
> I'm somewhat concerned about always hiding the address as that info is really helpful if you're using the street view to navigate down the street or if the street view is at a location near but not exactly at the address of the tree.

I was concerned about this as well, but I'm not sure what else to do.  At the size we're using for the street view box, the address information takes up more than 1/4 of the street view element and makes it very hard to see anything.

We could increase the size of the element on the main map and show the address there (we were not showing the address on the map page previously) but we're constrained on the plot detail by the size of the right column.

We could also try styling the address box differently, but it isn't under our control so I'm worried about that breaking in the future and making things worse.

RickMohr:
We could also consider making the street view box bigger on the plot detail page. 

Or removing controls from the current small street view box but adding a button to raise a popup showing a big version with all the controls.

dboyer:
That right column width does constrain what we can do with the size of the street view box and I don't want to even think about reworking the layout of that page overall. 

Rick's idea of having a pop-up showing the big version with all controls is interesting. Would the address still be in the smaller box that's always visible on the detail page?

opentreemap/treemap/js/src/streetView.js
```diff
@@ -32,7 +31,8 @@ exports.create = function(options) {
                 if (panorama === null) {
                     panorama = new google.maps.StreetViewPanorama(div, {
                         position:pos,
-                        addressControl: showAddress
+                        addressControl: false,
+                        panControl: false,
```
jwalgran: Lint: trailing comma

maurizi:
There is a built-in fullscreen mode that I left enabled (it's the button in the upper right of the street view box), which is a very simple and easy way to have a big version of the street view.

It looks like the options apply to both the tiny version and the full screen version, so if we want to lean on the full screen button for this we'd have to leave the address info enabled all the time.

If we want something in between we'd have to roll it ourselves.

We could always just increase the _height_ of the street view box, which would give us some room to fit the address info without totally obscuring the street images, without needing to mess with the layout of the plot detail page.

dboyer:
Definitely not worth building something ourselves. The height thing might work if it doesn't make the height/width ratio look too odd.

jwalgran:
I think increasing the height looks ok.

screen shot 2015-12-07 at 10 43 19 am

```diff
diff --git a/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss b/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss
index fc7b89e..d6ac444 100644
--- a/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss
+++ b/opentreemap/treemap/css/sass/partials/pages/_treedetails.scss
@@ -92,7 +92,7 @@
 .street-view-small {
     background: $well-color;
     width: auto;
-    height: 220px;
+    height: 350px;
     min-width: 100%;
     margin-bottom: 15px;
     border-radius: 9px;
diff --git a/opentreemap/treemap/js/src/streetView.js b/opentreemap/treemap/js/src/streetView.js
index c299da2..918ddab 100644
--- a/opentreemap/treemap/js/src/streetView.js
+++ b/opentreemap/treemap/js/src/streetView.js
@@ -31,8 +31,8 @@ exports.create = function(options) {
                 if (panorama === null) {
                     panorama = new google.maps.StreetViewPanorama(div, {
                         position:pos,
-                        addressControl: false,
-                        panControl: false,
+                        addressControl: true,
+                        panControl: false
                     });
                 } else {
                     panorama.setPosition(pos);

**Comment by jwalgran** [comment]
I think we should keep the `hideAddress` option so that we can hide it while adding a tree, when we already have a text box that shows the address.

screen shot 2015-12-07 at 10 50 15 am

**Comment by dboyer** [comment]
Okay, that sounds fine.

**Comment by maurizi** [comment]
Ok, this has been fixed to restore the address information and increase the screen size on the plot detail:

plot_detail

**Comment by dboyer** [comment]
+1 with the nice side benefit of now you don't have to pan up to see large trees

File: opentreemap/treemap/css/sass/partials/pages/_treedetails.scss [hunk 2]
```diff
@@ -98,10 +98,14 @@
     border-radius: 9px;
 }
 
-// Needed because bootstrap breaks street view on firefox
-// by setting max-width: 100% on all <img> tags
-.street-view-small img {
-    max-width: none;
+.street-view-small {
+    height: 440px;
```
jwalgran: Are we using 440px for any specific reason? Looked the best in trial-and-error testing?

maurizi: Because it gave a nice 2 to 1 size ratio of the box.  I explored going larger but then you start getting black bars on the top and bottom.  We could go smaller...

jwalgran:
:+1: I see the tall street view on the detail page and the address is hidden on the small street view shown while adding a tree.

PR #2370, opened 2015-12-03
Switch from thumbnail to full-size image for user profile photo.

maurizi:
We create a compresseed 85x85 pixel thumbnail for uploaded user photos,
which is intended to be used for the small profile photo in the header.

This is too small and shows compression artifacts when used on the user
profile page in a 195x195 pixel image.

Switching to the full size image for the user profile page fixes the issue.

Connects to OpenTreeMap/otm-addons#876


steventlamb:
+1

PR #2378, opened 2015-12-08
Fix ecobenefit cacheing

RickMohr:
* Actually turn on `USE_ECO_CACHE` :-D
* Cache benefits for all map feature types, not just `Plot`, so slow LA green infrastructure benefits are also cached. (Currently they use the `universal_rev`, so we flush the cache when any feature is updated. We could start keeping a `rev` by feature type if it becomes an issue.)

Connects OpenTreeMap/otm-cloud#105


maurizi:
@RickMohr It looks like there are some import errors when running the unit tests

opentreemap/treemap/ecocache.py
```diff
@@ -50,12 +50,12 @@ def _get_or_compute(prefix, filter, compute_value):
 def _get_key(prefix, filter):
     if filter and (filter.filterstr or filter.displaystr):
```
steventlamb: this should probably now be simplified to:
```
    if prefix == 'eco/Plot':
        version = filter.instance.eco_rev
    elif prefix == 'count/plots':
        version = filter.instance.geo_rev
    else:
        version = filter.instance.universal_rev
```
assigning to the same value in the first clause and the catch-all is confusing. Is it ok to drop out this check entirely (which we're effectively doing)?
```
    if filter and (filter.filterstr or filter.displaystr):
        version = filter.instance.universal_rev
```

RickMohr: No we still need that logic you added. I'll add this comment:

```
When a filter is active, we compute ecobenefits for only a subset of map features. 
In this case we use `universal_rev` in the cache key, since any change to a map feature might cause the filter to return a different subset.
```

steventlamb: How about:
```
filter_active =  filter and (filter.filterstr or filter.displaystr)
if prefix == 'eco/Plot' and not filter_active:
    version = filter.instance.eco_rev
elif prefix == 'count/plots' and not filter_active:
    version = filter.instance.geo_rev
else:
    version = filter.instance.universal_rev
```

steventlamb:
+1, tested

In order to test this, I added the following diff
```diff
diff --git a/opentreemap/treemap/ecocache.py b/opentreemap/treemap/ecocache.py
index b2a3dba..934722d 100644
--- a/opentreemap/treemap/ecocache.py
+++ b/opentreemap/treemap/ecocache.py
@@ -41,7 +41,9 @@ def _get_or_compute(prefix, filter, compute_value):
     else:
         key = _get_key(prefix, filter)
         value = cache.get(key)
+        print("try to read cache")
         if value is None:
+            print("value not found in cache")
             value = compute_value()
             cache.set(key, value, _TIMEOUT)
     return value
```
Then I visited an instance with gsi enabled and did the following:
Loaded the page with no search
Hard reloaded with no search
Loaded the page with a search
Hard reloaded with a search

As desired, this appeared in my console:
```
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
try to read cache
try to read cache
try to read cache
try to read cache
try to read cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
value not found in cache
try to read cache
try to read cache
try to read cache
try to read cache
try to read cache
try to read cache
```

opentreemap/treemap/ecocache.py
RickMohr: Here's a more concrete comment.

Example of why `eco_rev` is insufficient when a filter is active: You filter for only Oak trees. We compute and cache eco for your 25 oak trees. Now you update the species field of a tree from `null` to Oak. That doesn't change the `eco_rev`, so we incorrectly display the cached value instead of recomputing eco for your now 26 oak trees. Your update does change the `universal_rev` however, so using that in the cache key solves the problem.

Issue #2380, opened 2015-12-09
Catch UDF date and multichoice parsing errors while validating tree imports


jwalgran:
Catch and report parsing errors rather than allow them to crash the import task.

PR #2382, opened 2015-12-09
Handle import parsing of UDF dates and multichoice

jwalgran:
When importing a tree CSV with date and/or multichoice UDF columns, data in an invalid format would trigger the throwing of an `ValueError` which crashed the celery validation task. In this commit I continue the pattern of other UDF validations by catching the `ValueError` and converting it to a `ValidationError`, which is caught and logged by thevalidation task.
 
I wanted to add some UDF validation tests, but the TreeValidationTest class was already relatively long. I split the setUp and helper function into a base class that could be shared with multiple tree validation test classes.

Connects to #2380
Connects to #2383

### DIFF
--- a/opentreemap/importer/models/trees.py
+++ b/opentreemap/importer/models/trees.py
@@ -3,6 +3,8 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
+import json
+
 from django.core.exceptions import ValidationError
 from django.contrib.gis.db import models
 from django.contrib.gis.geos import Point, Polygon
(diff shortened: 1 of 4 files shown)


jwalgran:
Importing multichoice fields is not yet supported. See https://github.com/OpenTreeMap/otm-core/issues/2383.

I mistakenly though this was supported when I added multichoice validation handling. We may want to  add a file-level validation to reject an attempt to import a multichoice field rather than allow it to appear as a row-level validation failure.

jwalgran:
Reopening to get this attached to another issue in waffle

steventlamb:
+1, tested, great work

PR #2385, opened 2015-12-11
Pull ecoservice name from Django settings.

maurizi:
This allows the name to be customized by `local_settings.py` for different
environments.

Connects to OpenTreeMap/otm-cloud#17


RickMohr:
+1

Issue #2359, opened 2015-11-25
Stewardship activities do not export


jmorrison1847:
I've tested on a demo map and on Rhody trees (which uses the custom stewardship activity "Dead" on their map). The export field "udf: Stewardship" is always empty, even though there are instances of stewardship for both plots and trees. To illustrate:
On this demo map, there is one tree that clearly matches the criteria "TREE" "HAS BEEN" "WATERED" in the advanced search filter.
![screen shot 2015-11-25 at 8 54 54 am](https://cloud.githubusercontent.com/assets/12401491/11399004/81df94d8-9352-11e5-9a4b-78a570c4145e.png)
Looking at that tree detail page, we see it was watered on 11/16.
![screen shot 2015-11-25 at 8 55 26 am](https://cloud.githubusercontent.com/assets/12401491/11399016/94ce51ba-9352-11e5-99f8-f231ac7f6f1c.png)
However, in the export, the column "udf: Stewardship" is wholly blank.
![screen shot 2015-11-25 at 8 56 12 am](https://cloud.githubusercontent.com/assets/12401491/11399030/aecb38da-9352-11e5-966c-a730f3111889.png)

dboyer:
Maybe this is a regression. We fixed this bug back in June - https://github.com/OpenTreeMap/otm-core/issues/2011

PR #2386, opened 2015-12-11
Fix search exporting of collection UDFs.

maurizi:
This was broken by bf9b9d, which switched to `collection_udf_settings`.

Connects to #2359


jwalgran:
My test exports had values in the collection UDF columns as well as all other UDF columns.

:+1:

Issue #2379, opened 2015-12-08
Regression: incorrect display filter


RickMohr:
Production has the display filter "Show empty planting sites" (searching for "EmptyPlot")  but staging has "Show planting sites" (searching for "Plot").

RickMohr:
When this is fixed, test that refreshing the page, correctly sets the display filter for planting sites.

PR #2389, opened 2015-12-14
Advanced search UI

RickMohr:
* Restructure `instance.advanced_search_fields` and `instance.DEFAULT_SEARCH_FIELDS` to match advanced search categories in UI
* Create partials in new template folder `treemap/partials/advanced_search`
* Add `active` class to dropdowns and "Advanced" button when an advanced search is in effect
* Disable a model's search fields when that model isn't in the display filters
* In stewardship search, disable dropdowns until usable
* Remove CSS for no-longer-used advanced search sliders

Note: Temporarily hard-coding `rainBarrel.capacity` and `bioswale.udf:Perennial Plants` so styling work can proceed

Connects #2372


maurizi:
If I click into a date search box, then click on something else such as the map, the main search dropdown closes but the date picker window is still present:
floater

maurizi:
If I make `Stewardship` invisible to the current user, the dropdown for `Tree Care` is still present but it has no content.

maurizi:
If I set the permissions for `Planting Site Alerts` to invisible to the current user, I can still choose to search for Planting Sites with open alerts, but the third dropdown is empty.

maurizi:
When I refresh the page, my search terms are correctly applied to the DOM elements for everything except the display filter for planting sites.
This may be caused by https://github.com/OpenTreeMap/otm-core/issues/2379

maurizi:
Code looks good in general.  I think the problems with the date picker and the empty "Tree Care" dropdown should be fixed as part of this PR, but the problem with alerts could be pushed out into another issue.

RickMohr:
@maurizi, ready for another look.

maurizi:
:+1:

Issue #2366, opened 2015-12-02
Crash on user profile page


maurizi:
Seems to be caused by certain edits having data is no longer valid.

Stacktrace:
```
Internal Server Error: /users/jmorrison/
Traceback (most recent call last):
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django_tinsel/decorators.py", line 87, in wrapper
    RequestContext(request), **kwargs)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/shortcuts.py", line 45, in render_to_response
    using=using)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/loader.py", line 115, in render_to_string
    template_name, context, context_instance, dirs, dictionary)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/engine.py", line 221, in render_to_string
    return t.render(context_instance)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/base.py", line 209, in render
    return self._render(context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/base.py", line 201, in _render
    return self.nodelist.render(context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/base.py", line 903, in render
    bit = self.render_node(node, context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/debug.py", line 79, in render_node
    return node.render(context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/loader_tags.py", line 135, in render
    return compiled_parent._render(context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/base.py", line 201, in _render
    return self.nodelist.render(context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/base.py", line 903, in render
    bit = self.render_node(node, context)
  File "/usr/local/otm/env/local/lib/python2.7/site-packages/django/template/debug.py", line 79, in render_node
    return node.render(context)
[...]

PR #2390, opened 2015-12-14
Fix bug in formatting edit values for clashing custom field names.

maurizi:
Because we were not filtering by both model and field name, but just field
name, `UserDefinedFieldDefinition.clean_value` was being given the
`datatype_dict` of Bioswale.Perennial Plants for an edit on the
Perennial Plants field of a Rain Garden.

Connects to #2366

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -1268,11 +1268,14 @@ class Audit(models.Model):
                 udfd_pk = int(self.model[4:])
                 udf_def = next((udfd for udfd in udfds if udfd.pk == udfd_pk),
                                None)
-                datatype = udf_def.datatype_by_field[field_name]
+                if udf_def is not None:
+                    datatype = udf_def.datatype_by_field[field_name]
             else:
                 udf_def = next((udfd for udfd in udfds
-                                if udfd.name == field_name), None)
-                datatype = udf_def.datatype_dict
+                                if udfd.name == field_name
+                                and udfd.model_type == self.model), None)
+                if udf_def is not None:
+                    datatype = udf_def.datatype_dict
             if udf_def is not None:
                 return udf_def.clean_value(value, datatype)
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/audit.py
```diff
@@ -1268,11 +1268,14 @@ def _deserialize_value(self, value):
...
             else:
                 udf_def = next((udfd for udfd in udfds
-                                if udfd.name == field_name), None)
-                datatype = udf_def.datatype_dict
+                                if udfd.name == field_name
+                                and udfd.model_type == self.model), None)
+                if udf_def is not None:
+                    datatype = udf_def.datatype_dict
             if udf_def is not None:
                 return udf_def.clean_value(value, datatype)
```
steventlamb: I think it is better not to have conditionally defined variables that rely on a boolean to not trigger a `NameError`. If some aspect of the spec changes such that a developer changes that boolean such that it may be `True` sometimes (but not always) even when `udf_def` is `None`, then they may commit a `NameError` that they failed to produce in development.

Safer is something like:
```
e = Exception('Cannot format a UDF audit whose definition has been'
              ' deleted! This audit should be deleted as well')
if self.model.startswith('udf:'):
    udfd_pk = int(self.model[4:])
    udf_def = next((udfd for udfd in udfds if udfd.pk == udfd_pk),
                   None)
    if udf_def is not None:
        datatype = udf_def.datatype_by_field[field_name]
    else:
        raise e
else:
    udf_def = next((udfd for udfd in udfds
                    if udfd.name == field_name
                    and udfd.model_type == self.model), None)
    if udf_def is not None:
        datatype = udf_def.datatype_dict
    else:
        raise e
return udf_def.clean_value(value, datatype)
```

steventlamb: otherwise, +1 (didn't test though)

Issue #1907, opened 2015-01-14
Plot detail: Show i-Tree region and code


RickMohr:
This would make it much easier to diagnose ecobenefits problems.

RickMohr:
The eco section of plot detail can be formatted a few different ways:

![image](https://cloud.githubusercontent.com/assets/2162993/11666802/45b425e0-9dbc-11e5-809c-8e99c7f01c18.png)
![image](https://cloud.githubusercontent.com/assets/2162993/11666854/82eb38fe-9dbc-11e5-9d9b-1ac0303d4d93.png)

![image](https://cloud.githubusercontent.com/assets/2162993/11666829/67542812-9dbc-11e5-925d-e6fad23fa54a.png)

RickMohr:
@dboyer, @designmatty, like either of these options?

dboyer:
Either option is fine with me, but I think we should only display it for admins and super admins. It will be confusing to general users.

designmatty:
I like option two. Seems a bit overkill to hide it within a modal.

jwalgran:
I like showing the 2 additional rows in the yearly ecosystems services table.

PR #2392, opened 2015-12-15
Feature/1907/add itree info to plot detail

steventlamb:
connects #1907 on github

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -19,7 +19,6 @@ from urllib import urlencode
 
 from treemap.DotDict import DotDict
 from treemap.species import SPECIES
-from treemap.species.codes import ITREE_REGIONS
 from treemap.json_field import JSONField
 from treemap.lib.object_caches import udf_defs
 from treemap.species.codes import (species_codes_for_regions,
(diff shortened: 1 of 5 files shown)


opentreemap/treemap/instance.py
```diff
@@ -573,30 +572,16 @@ def update_revs(self, *attrs):
         for attr in attrs:
             setattr(self, attr, getattr(qs[0], attr))
 
-    def itree_region_codes(self):
-        from treemap.models import ITreeRegion
+    def itree_regions(self, **extra_query):
```
jwalgran: The return "type" of this method makes me a little unsure as to whether or not we should use this OO approach. Depending on the value of `itree_region_default` you are returned either a List[ITreeRegionInMemory] or an ITreeRegion QuerySet. Are there other places in the application where we duck type this intensely? Can you expand on your commit message and explain why you think this is better than constructing a list of code/name tuples?

jwalgran:
Unit tests pass. Adding/editing trees, seeing the region only when the user is privileged, and importing species worked without any problems.

opentreemap/treemap/instance.py
steventlamb: (summary of our discussion)
The list vs queryset ducktyping is idiomatic python. While there are not other places in the app we use ducktyping in this fashion, it is still a useful solution for encapsulating the data model of itree regions. There is also not a big performance worry, because ITreeRegion data is small, and these methods don't deal with the geom field.

I have expanded the commit message to explain why I think this is better.

jwalgran:
Thanks for expanding the commit message.

:+1:

PR #2394, opened 2015-12-16
Fix species importer i-Tree override tests

RickMohr:
`ecoservice_not_running` was always reporting that the eco-service was not running, so tests using it never ran.

* Fix `ecoservice_not_running`
* Fix `ITreeCommitTest._assert_overrides` so that species with merge conflicts are committed

Connects #2395


RickMohr:
waffle

RickMohr:
@maurizi, since the tests passed on CircleCI I think it's OK not to test this.

maurizi:
:+1:

PR #2402, opened 2015-12-18
Advanced search tweaks

RickMohr:
* Highlight advanced search "Tree Care" dropdown when stewardship search active
* Update field disabler now that "empty plot" display filter works again

Connects #2400


maurizi:
:+1:

PR #2404, opened 2015-12-18
Create search_fields.py from instance.py

RickMohr:
The search field constants and code were unwieldy in `instance.py`.


steventlamb:
+1 code looks good and advanced search works

PR #2410, opened 2016-01-06
Feature/mw/advanced search style

RickMohr:


RickMohr:
@designmatty, labels are now above inputs and check boxes still work. Shall I merge this?

Issue #2393, opened 2015-12-16
Blank error if you move a plot outside the map bounds (via Quick Edit)


RickMohr:
Console says 
```
Field error returned from server, but no dom element bound from client. mapFeature.geom ["Planting Sites must be created inside the map boundaries"]
```
Screen shows
![image](https://cloud.githubusercontent.com/assets/2162993/11846661/93baf9ea-a3e7-11e5-9c83-337349b096c6.png)

PR #2414, opened 2016-01-07
Fix error handling on the tree details quick edit

maurizi:
- Empty error boxes were appearing - fixed via CSS rule
 - Errors due to out of bounds plots were not appearing, fixed by adding
   an element to display mapFeature.geom errors.

Connects to #2393


steventlamb:
+1, tested (had to rebase on `develop`)

PR #2416, opened 2016-01-07
Add rollbar exception reporting for celery tasks

steventlamb:


opentreemap/opentreemap/celery.py
```diff
@@ -16,3 +19,12 @@
...
 app.config_from_object('django.conf:settings')
 app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
+
+
+@task_failure.connect
+def handle_task_failure(**kw):
+    rollbar_settings = getattr(settings, 'ROLLBAR', {})
+    access_token = rollbar_settings.get('access_token')
+    if access_token:
+        rollbar.init(access_token, rollbar_settings['environment'])
```
jwalgran: This handler will get called for every exception. Did you run across any documentation regarding whether or not it is a good idea to call `init` multiple times?

jwalgran: This was printed to the gunicorn standard out:

```WARNING:rollbar:Rollbar already initialized. Ignoring re-init.```

Looks like rollbar handles it properly, but is telling us not to do it. I am not sure if we should bother keeping our own flag. What do you think?

jwalgran: This example has `init` called when the module is loaded https://github.com/rollbar/rollbar-celery-example/blob/e92396064c379c480f036d1e361f66b7b2d0c409/tasks.py

jwalgran:
I added a `raise` to a celery task, triggered it twice, and saw it arrive in the dashboard twice.

screen shot 2016-01-08 at 5 59 12 pm

:+1:

opentreemap/opentreemap/celery.py
steventlamb: Thanks for bringing this up, I was confused about this.

Initially when I read https://github.com/rollbar/pyrollbar/blob/master/rollbar/contrib/django/middleware.py#L92 I thought that rollbar was being initialized once per request and so I copied that pattern. I see now in https://docs.djangoproject.com/en/1.9/topics/http/middleware/#init that this is actually happening once per process/thread, so it would be good to follow suit.

I do not know if celery will load the middleware and initialize rollbar. Even if it does, it creates an implicit dependency. I think it's probably better to initialize it once in `celery.py`, and once through the middleware. It looks like rollbar keeps track of inits with internal state and doesn't expose a way to check (https://github.com/rollbar/pyrollbar/blob/master/rollbar/__init__.py#L300). This exposes us to the `re-init` warning, but is probably OK, it's better than doing it on every exception. Does that make sense?

I do not understand what you mean by "bother keeping our own flag".

jwalgran: Related to https://github.com/rollbar/pyrollbar/blob/master/rollbar/__init__.py#L300. Rollbar is keeping track of calls to init, so we should not try to rig up a similar thing in our code.

I vote yes to moving the `init()` call out to the `celery.py` module so it is called once on load.

steventlamb: fixed

maurizi:
LGTM

Issue #1180, opened 2014-03-11
Improve validation of tree and plot fields


RickMohr:
Negative numbers are allowed for trunk diameter, tree height, canopy height, plot width and plot length.

PR #2417, opened 2016-01-08
Validate nullable numeric plot and tree fields

jwalgran:
Allowing physically impossible measurement values is bad for data quality, and impossibly large diameter values cause problem with eco benefit calculations.

##### Testing

You should not be able to enter negative values for tree height, tree diameter, tree canopy height, plot width or plot length.

If no species is selected, you should not be able to enter tree height > 800 ft or tree diameter > 200 in. On a map using the metric system, you should not be able to enter a tree height > 243.84 m or a tree diameter > 508 cm.

Adjusting the max allowable height and diameter of a `Species` through the Django console will and assigning that species to a tree should override the maximum values use for validation.

Note: Empty quick edit error messages should be addressed by #2414. There are validation error formatting issues on the tree detail page that are not addressed in this PR.

-----

Connects to #1180
Connects to #1928

### DIFF
--- a/opentreemap/treemap/models.py
+++ b/opentreemap/treemap/models.py
@@ -415,6 +415,10 @@ class Species(UDFModel, PendingAuditable):
     """
     http://plants.usda.gov/adv_search.html
     """
+
+    DEFAULT_MAX_DIAMETER = 200
+    DEFAULT_MAX_HEIGHT = 800
+
     ### Base required info
     instance = models.ForeignKey(Instance)
     # ``otm_code`` is the key used to link this instance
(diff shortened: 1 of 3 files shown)


maurizi:
Interestingly, while testing this I attempted to make a tree's diameter negative and it silently removed the `-` and saved it as a positive number.  I'm not even sure if that's a bug or an unintended feature?

maurizi:
Other than the weirdness with negative diameters, everything is working great. :+1:

jwalgran:
I can confirm that the multi-trunk diameter editor is coercing negative values to positive before posting. The tree height plain text field does not.

screen shot 2016-01-08 at 3 17 44 pm

I'll open an issue to have the diameter UI match what is PUT to the server.

PR #2418, opened 2016-01-08
Importer bug fixes

maurizi:
-  Always set importer status to canceled when calling the cancel endpoint
    On some occasions we cannot get the Celery task when attempting to call
    `GroupResult.restore()`, which caused attempts to call `.revoke()` to fail,
    which caused the change to the Import status to be rolled back.

    We explicitly check the return of `.restore()` before attempting to call
    `revoke()`, which prevents the transaction rollback from reverting the changed
    status.

 - Do not show the importer view link until it has started verification …

Connects to OpenTreeMap/otm-addons#990
Connects to OpenTreeMap/otm-addons#991

### DIFF
--- a/opentreemap/importer/templates/importer/partials/import_table.html
+++ b/opentreemap/importer/templates/importer/partials/import_table.html
@@ -27,10 +27,12 @@
                 {% endif %}
             </td>
             <td>
-              <a class="js-view"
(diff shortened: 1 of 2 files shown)


maurizi:
Waffle

steventlamb:
+1, tested

PR #2422, opened 2016-01-08
Feature/update field configs

steventlamb:


opentreemap/treemap/instance.py
```diff
@@ -359,9 +359,31 @@ def boundary_thumbprint(self):
         return "%s_%s" % (self.url_name, thumbprint)
 
     @transaction.atomic
-    def remove_map_feature_types(self, class_names):
-        remaining_types = [class_name for class_name in self.map_feature_types
-                           if class_name not in class_names]
+    def remove_map_feature_types(self, remove=None, keep=None):
+        from treemap.util import to_object_name
+        if keep and remove:
+            raise Exception('Invalid use of remove_map_features API')
```
jwalgran: I'd like to be more helpful to the caller and explain, in the Exception message, that you can use either `keep` or `remove` but not both.

steventlamb: fixed

maurizi:
:+1:

Issue #2427, opened 2016-01-11
Staging: Advanced search dropdown visible on tree detail page


dboyer:
On testmap0111

1. I added a photo to a tree detail page
2. On the social media sharing popup, I clicked no thanks
3. The pop-up went away and I could then see the dropdown arrow for advanced search behind the search button (see screenshot)

This happens in both Chrome and Firefox. It looks like the arrow first shows up when the pop-up box to select the photo to upload appears.

Fix as part of trees team 36

![adv_search_detail_page](https://cloud.githubusercontent.com/assets/516764/12241244/209d1b72-b860-11e5-9cc0-f527175c4f4f.png)

PR #2434, opened 2016-01-12
Fix style issues caused when closing social media sharing modal

maurizi:
There is an open issue in the version of Bootstrap we are using that causes
a 14-17px padding to be added to the body element, in an attempt to hide
the scrollbar of the page when modals are open.
https://github.com/twbs/bootstrap/issues/14040

As a side effect of this, opening a modal while another modal is still
closing (which is documented by Bootstrap as something to avoid) causes
this padding to remain in effect in even when the modal is closed.

I fixed that issue by ensuring that we wait until the upload modal has
closed before attempting to open the social media modal.

We may have to wait for a newer version of Bootstrap to fix the
main issue of unnecessarily adding padding when modals are open, or we may
be able to fix it with CSS rules.

Connects to #2427

### DIFF
--- a/opentreemap/treemap/js/src/imageUploadPanel.js
+++ b/opentreemap/treemap/js/src/imageUploadPanel.js
@@ -66,7 +66,11 @@ module.exports.init = function(options) {
             }
 
             if (callback) {
-                callback(new Bacon.Next({event: e, data: data}));
+                // Downstream users will be opening modals, which leads to
+                // style errors if that is done before a modal closes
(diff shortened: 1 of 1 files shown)


RickMohr:
I see the styling glitch when the "Add Photo" modal is first raised. With your fix it does go away when the second modal is closed, but I wonder if there's a CSS workaround to make it not show up ever? I note that only the header has the glitch, not the subheader or the body.

RickMohr:
+1 this is better, and CSS fix is not straightforward.

PR #2435, opened 2016-01-12
Remove history shim

RickMohr:
We added `native.history.js` for IE9 support. It is causing a bug, OpenTreeMap/otm-addons#1024.
We are not supporting IE9, so it can be removed.

Connects OpenTreeMap/otm-addons#1024


RickMohr:
waffle

maurizi:
:+1: Works as before and the bug is fixed

Issue #2332, opened 2015-11-02
Upgrade libsass-python


maurizi:
We pinned `libsass-python` to a specific commit to work around a regression.  The upstream library has been updated, so we can safely upgrade now.

We should probably also upgrade `grunt-sass` in conjunction so that both are using the same version of `libsass`.

maurizi:
We are no longer pinned to a specific commit of `libsass-python`, it was upgraded when we updated to use Webpack.

PR #2444, opened 2016-01-13
Update python dependencies to the latest versions

maurizi:
The aim behind doing this is to take advantage of the requires.io service, which has been sending us pull requests whenever a new version of one of our dependencies has been updated.

I opted to defer some updates by using the special `rq.filter` comment recognized by requires.io.
 - I left updated Django to the latest 1.8 release, but used a filter to stay on 1.8 instead of the latest, which is 1.9
 - I kept us at our current versions of `pep8` and `flake8`, because the latest versions have new linting rules which our codebase does not follow.  We need to fix our code lint at the same time that we update these (see https://github.com/OpenTreeMap/otm-core/issues/2443).
 - I left `libsass` using a git version even though the release we are pinned to has been released on PyPi, because we need to update the Python `libsass` in conjunction with `node-sass` (see https://github.com/OpenTreeMap/otm-core/issues/2332).

Code changes were required for a few upgrades:
 - Django extensions wanted us to define the base directory of our project in `settings.py`, which we were already doing, but using the wrong name https://github.com/maurizi/otm-core/commit/10780ba8e9ab21c85e688bea784cd5e56702f26b
 - Pillow now requires an extra argument to the `rotate` method 703c8f4
 - There was a typo in our `.travis.yml` that was previously silently ignored by coverage.py that now is not - I fixed the typo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - flake8 --exclude migrations,opentreemap/settings/local_settings.py,*.pyc opentreemap
   - npm run check
   - testem ci -l Firefox
-  - cd opentreemap && coverage run --rcfile=../coveragerc --source=. manage.py test
+  - cd opentreemap && coverage run --rcfile=../.coveragerc --source=. manage.py test
 
 after_success:
   coveralls
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,3 @@
 django-extensions==1.6.1
 six==1.10.0
-django-debug-toolbar==1.3.2
+django-debug-toolbar==1.4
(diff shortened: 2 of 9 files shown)


maurizi:
Reopened because somehow I caused Github to close the old PR....

jwalgran:
The tests are green on CI, which is a great sign. There is a lot of movement here so it will take me a little more time to walk through all the changes.

jwalgran:
:+1:

maurizi:
@jwalgran Your link reminded me that we ought to switch from the unmaintained `django-storages` to `django-storages-redux`, which is what newer projects have been using.  I made an issue to do that: https://github.com/OpenTreeMap/otm-core/issues/2456

PR #2445, opened 2016-01-13
Make batch moderation selector more specific

jwalgran:
[This is a new version of #2440 targeted at `develop`]

The single action selector was including the anchor tag for "View Details" on the photo moderation page, which should be a plan request rather than an AJAX request that replaces the content of the page with the response.

The links on the photo moderation page already had a class attribute set correctly. I added `class="action"` to the links on the comment moderation page so that they would still trigger AJAX requests when clicked.

##### Testing

Add a bunch of photos and comments then make sure all the moderation action links work correctly. "View Detail" should be the only anchor that works like a normal link, the rest should make AJAX calls.

----

Connects to OpenTreeMap/otm-addons#1030


maurizi:
:+1:

Issue #948, opened 2013-12-17
Use an autocompleter for location searches


andrewbt:
from http://www.bosqueurbanomexico.org/registraArbol.html (need to create account, but GIF screencast pasted below)

https://drive.google.com/file/d/0B3KytMYbF2eoZ3hkVUVkZjJTSGc/edit?usp=sharing

Issue #1124, opened 2014-02-19
Normal search and address search don't work well together


steventlamb:
If you enter an address, then shift+tab back to species and enter a species and hit enter, instead of doing the species search as you might expect, it attempts to match a geocoded location to your address search.

maurizi:
I just ran into this again today - and it is not just species search.  If you have done a search for an address, any further attempts to search (species, stewardship, display searches, etc) do not work, and instead it attempts to geocode.

PR #2447, opened 2016-01-14
Use ESRI suggest for autocompleting geocoded addresses

maurizi:
- Upgrades to typeahead.js 0.11.1 (the latest).
 - Use ESRI suggest endpoint to provide autocomplete of addresses
   This removes our client-side geocoder in favor of our (unused) server-side
   geocode endpoint using python-omgeo.

   Since the suggest endpoint is using the instance's extent as a viewbox, I
   was able to remove the viewbox code from the geocode endpoint, as results
   will be matched up with the 'magicKey' provided by ESRI suggestion results.

   The autocompletion has been wired into both the location search in the
   search bar as well as the location search when adding a tree.

   The styling also had be fixed up a bit since we now have typeaheads in
   places that used to just have a normal input.

Connects to #948

### DIFF
--- a/opentreemap/geocode/views.py
+++ b/opentreemap/geocode/views.py
@@ -13,7 +13,7 @@ from django.contrib.gis.geos.point import Point
 from django_tinsel.decorators import json_api_call
 
 from omgeo import Geocoder
-from omgeo.places import Viewbox, PlaceQuery
+from omgeo.places import PlaceQuery
 
 
 geocoder = Geocoder(sources=settings.OMGEO_SETTINGS)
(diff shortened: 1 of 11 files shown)


steventlamb:
getting some `urlState` issues that seem to go away when I switch back to `develop`
screenshot from 2016-01-14 12 51 16

steventlamb:
all tests done on a clone of phillytreemap:

1) If I search for "Queen Village" and opt for the geocode result (not the geom in our db), I can no longer search for a species.
2) If I search for "Queen Village" and opt for the geom in our db, it will highlight the boundary. If I then switch to an geocode address search, it will not clear the highlighted boundary, which is somewhat confusing.

Otherwise, +1, works.

maurizi:
Neither of the issues you noted are regressions, so I'm not going to attempt to fix them.

the first issue was already captured as https://github.com/OpenTreeMap/otm-core/issues/1124, and I've reported the second as https://github.com/OpenTreeMap/otm-core/issues/2466

Issue #2449, opened 2016-01-14
Missing search section needs text explaining that missing searches conflict with detail searches


maurizi:

RickMohr:
This is a regression -- the text used to be there.

PR #2455, opened 2016-01-15
Restore missing message

RickMohr:
CSS class `field-group-message` has `display: none` because most such messages
are hidden by default. This one is always shown, so override with `display: block`.

Connects #2449


maurizi:
LGTM

PR #2457, opened 2016-01-16
[requires.io] dependency update on develop branch

maurizi:


test-requirements.txt
```diff
@@ -1,3 +1,3 @@
 PyVirtualDisplay==0.1.5
-selenium==2.48.0
+selenium==2.49.0
```
maurizi: Based on the changelog this seems to be small bug fixes (https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES).  Since the UI tests passed on CircleCI I feel comfortable upping this.

requirements.txt
```diff
@@ -9,7 +9,7 @@ wsgiref==0.1.2
 pep8==1.4.6 # rq.filter: ==1.4.6
 flake8==2.0 # rq.filter: ==2.0
 python-omgeo==1.7.2
-django-registration-redux==1.2
+django-registration-redux==1.3
```
maurizi: According to the changelog on requires.io (https://requires.io/github/OpenTreeMap/otm-core/requirements/?branch=requires-io-develop) this has a bundle of new features that can get some good use out of!
 - Adds a `resend_activation_mail` method
 - Distinguishes between account activation failure and attempted reactivation.
 - Adds support for Django 1.9

I definitely want to do some smoke testing before we upgrade though.

jwalgran:
After provisioning my app server with this branch and running migrations:

- I was able to register and confirm a new account
- The UI tests passed

Looks good to me

PR #2460, opened 2016-01-18
Reusable components for filters on "list of items" pages

RickMohr:
* Template `treemap/partials/filter_dropdown.html` creates a filter dropdown
* Helper `treemap/lib/page_of_items/make_filter_context` sets up context for the template
* Enhance `UrlParms.url()` to take optional param overrides, allowing easy creation of a URL for each filter option.
* Change the comment and photo moderation pages to use the new scheme.

### DIFF
--- a/opentreemap/treemap/lib/page_of_items.py
+++ b/opentreemap/treemap/lib/page_of_items.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from django.core.urlresolvers import reverse
+
+
+class UrlParams(object):
+    def __init__(self, url_name, *url_args, **params):
+        self._url = reverse(url_name, args=url_args) + '?'
+        self._params = params
+
+    def params(self, *keys):
+        return self._param_string(keys, self._params)
+
+    def url(self, *keys, **overrides):
+        params = self._params
+        if overrides:
+            keys += tuple(overrides.keys())
+            params = dict(params.items() + overrides.items())
+        return self._url + self._param_string(keys, params)
+
+    @staticmethod
+    def _param_string(keys, params):
+        return '&'.join(['%s=%s' % (key, params[key]) for key in keys])
+
+
+def make_filter_context(urlizer, filter_value, option_data):
+    # Set up context for the treemap/partials/filter_dropdown.html template.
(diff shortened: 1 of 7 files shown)


opentreemap/treemap/lib/page_of_items.py
```diff
@@ -0,0 +1,54 @@
...
+            keys += tuple(overrides.keys())
+            params = dict(params.items() + overrides.items())
+        return self._url + self._param_string(keys, params)
+
+    @staticmethod
+    def _param_string(keys, params):
+        return '&'.join(['%s=%s' % (key, params[key]) for key in keys])
+
+
+def make_filter_context(urlizer, filter_value, option_data):
```
maurizi: I think this function can be cleaned up and made a little more pythonic by moving this from a free floating function to another method of `UrlParams`, giving it access to the private `_params` dict.  Consider changing it to this:

```python
class UrlParams(object):
    ...
    def filter_context(self, filters, options):
        filter_value = {self._params[key] for key in filters}
        # Same as before
```

RickMohr: I agree that your suggestion would simplify the code somewhat. But currently `UrlParams` is a fairly generic class and I hesitate to add such specific view-context details.

opentreemap/treemap/templates/treemap/partials/filter_dropdown.html
```diff
@@ -0,0 +1,14 @@
+<div class="btn-group">
+    <button type="button" class="btn btn-sm btn-info dropdown-toggle" data-toggle="dropdown">
+        {{ filter.selected_option.label }} <span class="caret"></span>
+    </button>
+    <ul {{ filter.container_attr }} class="dropdown-menu" role="menu">
```
maurizi: Can you add a class of `dropdown-menu-right` to this UL?  It was on _one_ of the filters that got removed in favor of this partial, and fixes issues with the dropdown menu getting cut off by the edge of the screen.

opentreemap/treemap/lib/page_of_items.py
maurizi: Makes sense

maurizi:
Other than my comment about the `dropdown-menu-right` class, this looks good!  Nice to see some repetitive code/HTML abstracted away. :+1: once that class is added in.

PR #2461, opened 2016-01-18
Feature/importer status page checks

steventlamb:


steventlamb:
test fixes coming soon

opentreemap/importer/models/base.py
```diff
@@ -30,6 +31,15 @@ class Meta:
     FAILED_FILE_VERIFICATION = 6
     CANCELED = 8
     VERIFICATION_ERROR = 10
+    UNKNOWN_ERROR = 12
+
+    @classproperty
+    def INACTIVE_CODES(self):
```
RickMohr: How about `STATUS_INACTIVE`? (My first thought was "These codes aren't used any more")

opentreemap/importer/models/base.py
```diff
@@ -30,6 +31,15 @@ class Meta:
     FAILED_FILE_VERIFICATION = 6
     CANCELED = 8
     VERIFICATION_ERROR = 10
+    UNKNOWN_ERROR = 12
```
RickMohr: Since this represents the specific case where the celery task disappears without completing I think it would be clearer to call it something like `TASK_VANISHED`. People reading the code would have an idea what it's about, while `UNKNOWN_ERROR` could mean many things.

opentreemap/importer/views.py
```diff
@@ -110,6 +113,18 @@ def get_import_table(request, instance, table_name):
 _EVENT_TABLE_PAGE_SIZE = 5
 
 
+def _cleanup_tables(instance):
+    ievents = (list(TreeImportEvent.objects.filter(instance=instance)) +
+               list(SpeciesImportEvent.objects.filter(instance=instance)))
+    for ie in ievents:
+        if ie.task_id != '':
```
maurizi: Since we previously did not unset `task_id` after an import finished, should we exclude imports that are in the `FINISHED_CREATING` state, so that we do not move finished imports into an `UNKNOWN_ERROR` incorrectly?

opentreemap/importer/views.py
```diff
@@ -110,6 +113,18 @@ def get_import_table(request, instance, table_name):
...
+            result = AsyncResult(ie.task_id)
+            if result.failed():
```
maurizi: Do we also want to mark the import as failed if we could not retrieve the `AsyncResult`?  I'm not sure how long Celery/Redis holds onto these, and I doubt they'd still be there after switching Redis backends, like we've recently done.

opentreemap/importer/views.py
steventlamb: Maybe not necessary because they would not produce a truthy value for `AsyncResult(ie.task_id).failed()`, but can't hurt, so I'll add a check to only look at import events that are not in finished states.

opentreemap/importer/tasks.py
```diff
@@ -144,7 +144,10 @@ def commit_import_event(import_type, import_event_id):
 
     finalize_task = _finalize_commit.si(import_type, import_event_id)
 
-    chord(commit_tasks, finalize_task).delay()
+    async_result = chord(commit_tasks, finalize_task).delay()
```
maurizi: Looking at the [Celery code](https://github.com/celery/celery/blob/master/celery/result.py#L777-L787) suggests that we may need to call `async_result.save()` here (and in other places where we store a GroupResult) in order to later be able to restore the result, though I'm not sure how that works if restoring it via the `AsyncResult` base class `__init__` constructor as opposed the subclass static method `GroupResult.restore`.

opentreemap/importer/models/base.py
steventlamb: This method produces a list of codes that are used by predicate-style named methods (eg. `is_finished`), of which this class has several. For that reason, I think giving it a non-predicate-style name is a good idea, which is why I chose the `_CODES` suffix.

I do agree with your point. How about `INACTIVE_STATUS_CODES`?

RickMohr:
We never discussed what user-visible changes this work should aim to support. Please see proposal at https://github.com/OpenTreeMap/otm-addons/issues/1045

opentreemap/importer/models/base.py
RickMohr: If we adopt OpenTreeMap/otm-addons#1045 I think this variable will no longer have two uses, and we should:
* add the new statuses to `is_finished` (possibly renaming it...)
* change the query for what goes in the "Finished" table to include only `FINISHED_CREATING` and `FAILED_FILE_VERIFICATION`. (`VERIFICATION_ERROR` indicates an internal error, so the job should be resumable.)

opentreemap/importer/models/base.py
steventlamb: agreed.

opentreemap/importer/tasks.py
steventlamb: This worked in testing without saving. I will do some tests that include a celery outage/restart and see how saving affects the situation. If needed, will add.

opentreemap/importer/views.py
steventlamb: Probably better to be conservative about assigning this error state in fear of accidentally applying it to a perfectly valid job. If being unable to retrieve the `AsyncResult` turns out to be a problem in the wild, we can address it then. Does that make sense?

RickMohr: I think we should mark it as failed if there's no `AsyncResult` *and* the job is in a "running" state. Reasons:
* It will allow LA to restart their stalled imports
* Seems like if the celery task is actually running then `AsyncResult(ie.task_id)` should always work.

opentreemap/importer/models/base.py
steventlamb: agreed.

opentreemap/importer/tasks.py
steventlamb: `AsyncResult` does not have a save method. I originally tried to use a `GroupResult` for this feature and was not able to make it work.

opentreemap/importer/models/base.py
steventlamb: fixed.

steventlamb:
@RickMohr this is ready for another look

opentreemap/importer/models/base.py
```diff
@@ -65,12 +66,10 @@ def status_summary(self):
             self.FAILED_FILE_VERIFICATION: "Invalid File Structure",
             self.CANCELED: "Canceled",
             self.VERIFICATION_ERROR: "Verification Error",
+            self.TASK_VANISHED: "Unknown Error",
```
RickMohr: To be able to resume jobs I think we need:
```
self.VERIFICATION_TASK_VANISHED: "Failed During Verification",
self.CREATION_TASK_VANISHED: "Failed During Tree Creation",
```

... and to make the message right for species see https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/importer/models/species.py#L50

RickMohr:
If I start an import and then restart celery my job is stuck in "Verifying" with e.g. 982/999 rows completed. The `task_id` has an `AsyncResult`, but `failed()` returns false. Its status is `PENDING`, and it doesn't appear to have any useful information.

Do we have to find the child tasks and examine them??

steventlamb:
I ran into a similar issue, but only if I restarted celery without waiting for workers to cleanup. It's not clear what we can or should do in this case. From the docs:
```
It’s important to note that the worker is not known to crash, and if it does it is usually an unrecoverable error that requires human intervention (bug in the worker, or task code).
```
http://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late

I tried to force a random out-of-memory situation in my VM and was unable. Do we know if celery has actually ever crashed in those cases, or just that workers died on a task?

jwalgran:
The OOM Killer is something we need to protect ourselves from. After some googling I am not 100% sure whether it sends SIGTERM (`kill -15`) or SIGKILL (`kill -9`) to the process, but we can definitely use `kill` to test the effects of the celery process being shut down violently.

steventlamb:
@rcheetham @jwalgran @maurizi I think this is ready for another review, and barring the TODO questions, ready to merge. The `kill -9` case is covered.

opentreemap/importer/tasks.py
```diff
@@ -86,13 +86,13 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
...
 
     try:
-        row_set = ie.rows()
-        validation_tasks = (
-            _validate_rows.subtask(row_set[i:(i+settings.IMPORT_BATCH_SIZE)])
-            for i in xrange(0, ie.row_count, settings.IMPORT_BATCH_SIZE))
+        validation_tasks = []
+        for i in xrange(0, ie.row_count, settings.IMPORT_BATCH_SIZE):
+            args = (import_type, ie.id, i)
+            validation_tasks.append(_validate_rows.subtask(args))
```
RickMohr: Simpler:
```python
validation_tasks.append(_validate_rows.subtask(import_type, ie.id, i))
```
```python
def _validate_rows(import_type, import_event_id, start_row_id):
```

opentreemap/importer/models/base.py
```diff
@@ -47,6 +48,9 @@ class Meta:
 
     status = models.IntegerField(default=PENDING_VERIFICATION)
 
+    last_processed_at = models.DateTimeField(null=True, blank=True)
+    is_lost = models.BooleanField(default=False)
```
RickMohr: Like all new fields these days this should be null=True with a TODO (or an unmerged PR) to remove null=True.

opentreemap/importer/models/species.py
```diff
@@ -263,6 +263,7 @@ def validate_itree_code_field(self):
         else:
             self.cleaned[fields.species.ITREE_PAIRS] = pairs
 
+    @transaction.atomic
```
RickMohr: Since this function is not fast, perhaps the transaction should just go around the `save`s at the end.

RickMohr: Or maybe the transaction is for a different reason? Seems OK to have those two `save`s without a transaction.

RickMohr: Same questions for the other transactions you added.

opentreemap/importer/views.py
```diff
@@ -117,23 +140,26 @@ def _get_table_context(instance, table_name, page_number):
     species = SpeciesImportEvent.objects\
         .filter(instance=instance)\
         .order_by('-created')
-    finished = GenericImportEvent.FINISHED_CREATING
+    inactive_q = Q(is_lost=True) | Q(
```
RickMohr: Lost tasks are resumeable so shouldn't they stay in the "Active" table?

opentreemap/importer/tasks.py
steventlamb: I tried it this way and it did not work. I think it has something to do with the way celery handles arg passing to subtasks.

opentreemap/importer/views.py
steventlamb: I think it makes sense to leave them in the 'finished' table until the resumption feature is implemented.

opentreemap/importer/models/species.py
steventlamb: Interesting, I just assumed that the usual rule applies - use transactions to avoid inconsistent states arising from relationships that are not enforced at the schema level. Is your point that a transaction is not necessary because, while incorrect, it's not the worst thing ever if one succeeds and one fails?

Do you have some reason to believe that just wrapping the saves in a transaction is substantively faster? It comes at the cost of readability. It might be the case that the transaction is initiated at the first db call, which should result in the same performance.

opentreemap/importer/models/base.py
steventlamb: Thanks. Fix on the way.

opentreemap/importer/models/species.py
RickMohr: Right, a failure doesn't really matter -- there's no integrity problem if we correctly save the heartbeat and then fail to save the row or if we correctly save the row and then fail to save the heartbeat.

So I think we don't need the transactions. As to the speed question, my point was that those methods do a lot of time-consuming things before getting to the saves, which were the point of adding the transaction.

opentreemap/importer/tasks.py
RickMohr: Here's a multiple-argument task:
https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/exporter/tasks.py#L230

opentreemap/importer/models/species.py
steventlamb: I will remove the (newly added) transaction decorators.

opentreemap/importer/tasks.py
steventlamb: this worked with `.s` instead of `.subtask`. Fix pushed using this modified suggestion.

opentreemap/importer/models/base.py
steventlamb: fix pushed

steventlamb:
@RickMohr all comments addressed

opentreemap/importer/models/base.py
```diff
@@ -55,6 +60,10 @@ def row_count(self):
         return self.rows().count()
 
     def status_summary(self):
+        t = "Unknown Error During %s" if self.is_lost else "%s"
```
RickMohr: If you change "During" to "While" the status message will read better:

Unknown Error During Not Yet Started
Unknown Error During Loading
Unknown Error During Preparing Verification
Unknown Error During Verifying
Unknown Error During Creating Trees

Unknown Error While Not Yet Started
Unknown Error While Loading
Unknown Error While Preparing Verification
Unknown Error While Verifying
Unknown Error While Creating Trees

RickMohr:
+1 After changed my timeout to 3 minutes I was able to see imports with these states (by stopping celery or changing the code to not proceed):

Unknown Error During Loading
Unknown Error During Preparing Verification
Unknown Error During Verifying
Unknown Error During Creating Trees

PR #2462, opened 2016-01-19
Move universal_rev_hash from APIv4 to APIv5

maurizi:
We are adding support to the mobile apps for multichoice fields, but are not
yet adding support for universal_rev_hash, so I am leaving multichoice fields
available in v4 but moving universal_rev_hash to only be supported in v5 and
above.

### DIFF
--- a/opentreemap/api/instance.py
+++ b/opentreemap/api/instance.py
@@ -34,6 +34,7 @@ def transform_instance_info_response(instance_view_fn):
 
     v3 - Collection UDFs added
     v4 - Multiselect fields added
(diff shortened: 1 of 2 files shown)


jwalgran:
I created an instance with a multichoice field and added it to the the multichoice field to the search configuration. 

- When I set the iOS app to API v3, the multichoice field is not returned and the response contains `geoRevHash` instead of `universalRevHash`
- When I set the iOS app to API v4, the multichoice field details are returned and the response still contains `geoRevHash` instead of `universalRevHash`

:+1:

Issue #2458, opened 2016-01-18
UserDefinedFieldDefinition.reverse_clean does not properly handle unicode values


jwalgran:
The ``reverse_clean`` method, used when setting the value of a UDF, is passing unicode strings to the `str()` function, which fails.

The `str()` call is here:
https://github.com/OpenTreeMap/otm-core/blob/8423bff2c89badf234085f898fa066d9cf9a0f0f/opentreemap/treemap/udf.py#L740

Here is the one place where `reverse_clean` is invoked:
https://github.com/OpenTreeMap/otm-core/blob/8423bff2c89badf234085f898fa066d9cf9a0f0f/opentreemap/treemap/udf.py#L959

This was discovered while importing field values with accented characters into a Plot UDF.

jwalgran:
This may be connected to #2021

PR #2464, opened 2016-01-21
Handle non-ASCII characters in text and choice UDF fields

jwalgran:
An exception was being thrown when setting the value of a text or choice UDF if the value contained non-ASCII characters. This was caused by indiscriminately using `str()` to coerce values to strings.

Advanced searches that included non-ASCII characters in text or choice fields were also raising exceptions. The changes in this PR ensure that values are always encoded and decoded properly.

##### Testing

Use a test instance with the customization module enabled.

Add a text UDF and a choice UDF to the test instance. Ensure that the choice UDF has choices that contain non-ASCII characters like ≥. Add these fields to the search configuration so that they are available for advanced searching.

Add trees to instance and set UDF values that contain non-ASCII characters. Verify that the non-ASCII characters are saves and display correctly on the tree detail page.

Verify that advanced searches with non-ASCII values correctly return results.

---

Connects to #2458

### DIFF
--- a/opentreemap/treemap/ecobackend.py
+++ b/opentreemap/treemap/ecobackend.py
@@ -73,7 +73,7 @@ def json_benefits_call(endpoint, params, post=False, convert_params=True):
 
                     v = "ST_GeomFromEWKT('%s')" % GEOSGeometry(hexstring).ewkt
                 else:
-                    v = str(v)
+                    v = v if isinstance(v, unicode) else str(v)
 
                 if k == "param":
                     if k in paramdata:
--- a/opentreemap/treemap/ecobackend.py
+++ b/opentreemap/treemap/ecobackend.py
@@ -72,8 +72,8 @@ def json_benefits_call(endpoint, params, post=False, convert_params=True):
                     hexstring = ''.join('%02X' % ord(x) for x in bytestring)
 
                     v = "ST_GeomFromEWKT('%s')" % GEOSGeometry(hexstring).ewkt
-                else:
-                    v = v if isinstance(v, unicode) else str(v)
+                elif not isinstance(v, unicode):
+                    v = str(v)
 
                 if k == "param":
                     if k in paramdata:
--- a/opentreemap/treemap/ecocache.py
+++ b/opentreemap/treemap/ecocache.py
@@ -67,7 +67,10 @@ def _get_key(prefix, filter):
         version = filter.instance.universal_rev
 
     filter_key = '%s/%s' % (filter.filterstr, filter.displaystr)
-    filter_hash = hashlib.md5(filter_key).hexdigest()
+    # Explecitly calling `encode()` ensures that the presence of a
+    # unicode symbol in the filter string will not raise a
+    # UnicodeEncodeError exception when calling `md5()`
+    filter_hash = hashlib.md5(filter_key.encode('utf-8')).hexdigest()
     key = "%s/%s/%s/%s" % (prefix,
                            filter.instance.url_name,
                            version,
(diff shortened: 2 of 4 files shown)


opentreemap/treemap/ecobackend.py
```diff
@@ -73,7 +73,7 @@ def json_benefits_call(endpoint, params, post=False, convert_params=True):
 
                     v = "ST_GeomFromEWKT('%s')" % GEOSGeometry(hexstring).ewkt
                 else:
-                    v = str(v)
+                    v = v if isinstance(v, unicode) else str(v)
```
RickMohr: slightly simpler:
``` python
elif not isinstance(v, unicode):
    v = str(v)
```

opentreemap/treemap/ecocache.py
```diff
@@ -67,7 +67,10 @@ def _get_key(prefix, filter):
         version = filter.instance.universal_rev
 
     filter_key = '%s/%s' % (filter.filterstr, filter.displaystr)
-    filter_hash = hashlib.md5(filter_key).hexdigest()
+    # Explecitly calling `encode()` ensures that the presence of a
```
RickMohr: Explecitly > Explicitly

maurizi:
Looking at this now.

opentreemap/treemap/ecobackend.py
jwalgran: Implemented in 84406d8

opentreemap/treemap/ecocache.py
jwalgran: Fixed in 84406d8

maurizi:
I added a text field, choice field, and multichoice field.  I was able to save values of `☃` on all three fields, and I was able to correctly retrieve search results for the textfield and the single choice field.

However, while the multi choice search didn't error out, it did not return results when it should have.

jwalgran:
I was able to reproduce the error, but I think it is outside the scope of the this PR/issue.  I see that an advanced search on a multichoice field turns into a "LIKE" filter. My hypothesis is that "LIKE" will not match unicode values and we need to change our filtering to do so.

```
http://localhost:6060/jw-2016-01-28/map/?z=11/34.0353/-118.1532&q={"tree.udf:Color":{"LIKE":"\"Bröwn\""}}
```

maurizi:
It looks like the problem is actually that due to our encoding multichoices into JSON before storing them, they are not properly encoding as unicode in the DB, but rather are pre-escaped.

I did a `SELECT udfs FROM treemap_mapfeature WHERE pk=<my mapfeature>` for a plot that had a unicode value in a single choice, text field, and a multi choice.
```
"Multi"=>"[\"A \\u2603\", \"B\"]", "Single"=>"A ☃", "Text field"=>"☃"
```
Notice how the multichoice values are stored as `\\u2603` but the other values are stored as `☃`.  I suspect we could fix this by ensuring that multichoice values containing unicode characters are stored as UTF-8 and not with the unicode stored using escape characters (with a possible data migration? I'm not sure if it was possible to save a unicode mutlichoice value prior to this PR)

maurizi:
And if we store the unicode values correctly, we should be able to continue to use a `LIKE` search.

jwalgran:
Saving the multichoice JSON as Unicode is implemented in 5187407

Before

```
otm=# SELECT udfs->'Parking' FROM treemap_tree WHERE plot_id=1562026;
             ?column?
----------------------------------
 ["\ud83d\ude93", "\ud83d\ude96"]
(1 row)
```

After

```
otm=# SELECT udfs->'Parking' FROM treemap_tree WHERE plot_id=1562026;
  ?column?
------------
 ["🚓", "🚖"]
(1 row)
```

jwalgran:
`reverse_clean` was never working for unicode values, so I do not think a data migration is required.

maurizi:
:+1:

jwalgran:
c97b606 is a rebase that fixes commit message grammar

PR #2465, opened 2016-01-22
Extend support for configurable mobile search fields.

maurizi:
Removes the need to set anything other than the field identifiers
when setting custom mobile search fields, similar to setting custom search
fields for the website.  Instead, the search type, label, and other required
properties are generated when making API requests.

Still needs some additional work:
 - ~~Need to remove fields from the configuration when their UDFDs are deleted~~
 - ~~Needs some styling work~~
 - ~~Could use some unit tests~~

Connects to OpenTreeMap/otm-addons#1038


maurizi:
@jwalgran This is ready for another look.

maurizi:
@jwalgran I added the API response transformations to remove unsupported earch types from older API versions as we discussed, and I added some unit tests.

This PR and all associated PRs should be totally done now and ready for review.

jwalgran:
The labels generated for the alert fields are capitalized but the labels for other fields are lower case

screen shot 2016-01-28 at 12 03 18 am

jwalgran:
I have erroring tests and I can't create instances on this branch. I had similar problems with another PR and the solution was to rebase onto develop.

maurizi:
@jwalgran I removed bounds from the migration and title-cased missing data searches.  I also rebased on develop, which unfortunately removed your comments from the PR, but I replied to each one.

jwalgran:
```
CommandError: Conflicting migrations detected (0018_merge, 0018_add_species_field_names in treemap).
To fix them run 'python manage.py makemigrations --merge'

maurizi:
@jwalgran rebased _again_ and added a merge migration.

jwalgran:
2f333b9 introduced some small test failures

maurizi:
Just noticed that too, and pushed a commit to fix them.

jwalgran:
:+1:  after appeasing the unit test gods

Issue #2399, opened 2015-12-17
Address slowness when instance bounds are complex


RickMohr:
Proposed solution is to store instance bounds in a separate `InstanceBounds` table rather than directly on `Instance`.

We considered using the Django ORM `defer` when loading an instance. But there's no way to say "defer `bounds` whenever an instance is loaded" -- we'd have to use it all over the place. And there's no way to use it when dereferencing a related object (e.g. `plot.instance`).

PR #2467, opened 2016-01-22
Feature/move instance bounds to separate model

steventlamb:
connects #2399 on github

This changeset detaches bounds, which may be large, complicated geometries, from instances. This means that bounds objects will only be loaded when actually used. This results in a large importer speedup (~60%) for instances with complicated bounds. The speedup should affect most pages on the site, but may only be noticeable on large instances, if at all.

In order to test this PR, attempt to create a try and attempt the normal bounds operations for it: placing a tree out of bounds, finding nearby instances, etc.

These migrations are meant to gracefully support blue/green deployments. There will be a companion PR, one per repo, that makes another change, cumulatively completing the non-backward compatible change across two releases. When this PR is merged the other PR will be opened. When this PR is released, the other PR will then be merged to develop.

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -158,7 +158,6 @@ class Instance(models.Model):
         'BenefitCurrencyConversion', null=True, blank=True)
 
     """ Center of the map when loading the instance """
-    bounds = models.MultiPolygonField(srid=3857)
     bounds_obj = models.OneToOneField(InstanceBounds,
                                       on_delete=models.CASCADE,
                                       null=True, blank=True)
@@ -272,7 +271,7 @@ class Instance(models.Model):
 
     @property
     def extent_as_json(self):
-        boundary = self.bounds.boundary
+        boundary = self.bounds_obj.geom.boundary
         xmin, ymin, xmax, ymax = boundary.extent
 
         return json.dumps({'xmin': xmin, 'ymin': ymin,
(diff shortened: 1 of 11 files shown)


maurizi:
> When this PR is merged the other PR will be opened. When this PR is released, the other PR will then be merged to develop.

I think you can and should open the other PR now.  Just make sure to note in the PR title that it should not be merged.

opentreemap/treemap/instance.py
```diff
@@ -149,7 +158,10 @@ class Instance(models.Model):
         'BenefitCurrencyConversion', null=True, blank=True)
 
     """ Center of the map when loading the instance """
-    bounds = models.MultiPolygonField(srid=3857)
+    bounds = models.MultiPolygonField(srid=3857, null=True, blank=True)
```
maurizi: Based on what we discussed as the plan for making this safe for a blue/green deployment, I think we need to *remove* this field (without checking in the accompanying migrations)

maurizi: And we can delete the migration that makes the field nullable :)

steventlamb:
I understand that it is possible to open the PR now, but I would rather wait until review and edits are done. There's some overhead in managing a total of six open PRs at a time.

maurizi:
I think the order of migrations 16 + 17 needs to be flipped, so that we can  null out the original `bounds` field when copying the data over to the new `InstanceBounds` model as part of the data migration.

steventlamb:
@maurizi migrations swapped, running forward and backwards works.

steventlamb:
pushed another commit to drop `instance.bounds`

jwalgran:
The mobile API is responding correctly.

Requests to the `locations/x,y/instances` API from the iOS simulator to my dev vm took ~50ms on `develop` and ~150ms on this branch. While slower, this is an acceptable response time for an API that is only called once when a mobile app is launched. There is a maximum distance filter built into the API as well, so an increasing number of public instances should not cause the performance of this to degrade.

I grepped for usages of "bounds" and did not find any locations where we did not switch to `bounds_obj`.

Instance bounds validations while adding trees still work, and I did see the import validation performance increase about 3x on my dev VM.

jwalgran:
:+1:

Issue #2468, opened 2016-01-26
Add an index to ImportRow.idx


maurizi:
We sort `ImportRow` models by `idx` in [multiple places in the importer](https://github.com/OpenTreeMap/otm-core/search?utf8=%E2%9C%93&q=order_by%28%27idx%27%29), and without an index on the column it is putting a large strain on the DB.

PR #2469, opened 2016-01-26
Add a database index to ImportRow.idx for improved performance.

maurizi:
Connects to #2468


maurizi:
It occurs to me that we may also/only need a combined index of `idx` and `importevent_id`

hectcastro:
I don't think I'd be prepared to merge this pull request without more evidence to show that this change:

1. Has a positive impact on importer performance
2. Doesn't have a negative impact in other areas of the application

When I was troubleshooting https://github.com/OpenTreeMap/otm-addons/issues/1057, replacing the `ORDER BY` with indexed and unindexed columns didn't seem to make much of an impact. Also worth noting that those results are just one data point and not an average of multiple tests.

PR #2471, opened 2016-01-26
Support for super admin instance details

RickMohr:
* Add and maintain `InstanceUser.last_seen` date
* Add template for showing read-only fields in an inline edit form
* Simplify code in `form_extras.py`
* Use "Yes/No" instead of "True/False" for default boolean display value
  This matches the server-side display values used in `form_extras.py`
* Make the display value of a choice field the choice's display value

Connects OpenTreeMap/otm-addons#1035

### DIFF
--- a/opentreemap/treemap/models.py
+++ b/opentreemap/treemap/models.py
@@ -507,6 +507,11 @@ class InstanceUser(Auditable, models.Model):
     role = models.ForeignKey(Role)
     reputation = models.IntegerField(default=0)
     admin = models.BooleanField(default=False)
+    last_seen = models.DateField(null=True, blank=True)
+
+    def __init__(self, *args, **kwargs):
+        super(InstanceUser, self).__init__(*args, **kwargs)
+        self._do_not_track.add('last_seen')
 
     class Meta:
         unique_together = ('instance', 'user',)
--- a/opentreemap/treemap/templatetags/form_extras.py
+++ b/opentreemap/treemap/templatetags/form_extras.py
@@ -393,6 +393,10 @@ class AbstractNode(template.Node):
             # there's no meaningful intermediate value to send
             # without rendering the same markup server-side.
             display_val = None
+        elif choices:
+            display_vals = [choice['display_value'] for choice in choices
+                            if choice['value'] == field_value]
+            display_val = display_vals[0] if display_vals else field_value
         else:
             display_val = unicode(field_value)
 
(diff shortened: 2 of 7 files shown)


opentreemap/treemap/models.py
```diff
@@ -507,6 +507,7 @@ class InstanceUser(Auditable, models.Model):
     role = models.ForeignKey(Role)
     reputation = models.IntegerField(default=0)
     admin = models.BooleanField(default=False)
+    last_seen = models.DateField(null=True, blank=True)
```
jwalgran: `InstanceUser` is `Auditable` so I think we need to explicitly exclude this field from audit tracking to avoid creating an audit row on every authenticated page request.

jwalgran:
There is another open PR with treemap migrations 0015, 0016, and 0017 

https://github.com/OpenTreeMap/otm-core/pull/2467/files

We will need to coordinate merging the migrations.

opentreemap/treemap/models.py
jwalgran: I saw that we are using `now().date()`, so we would only generate an audit row once per day. While this is a lot better, I think those rows are mostly just adding noise.

opentreemap/treemap/templatetags/form_extras.py
```diff
@@ -377,12 +377,10 @@ def _field_value(model, field_name, data_type):
 
         if field_value is None:
             display_val = None
-        elif data_type == 'date' and model.instance:
-            display_val = dateformat.format(field_value,
-                                            model.instance.short_date_format)
-        elif data_type == 'date':
-            display_val = dateformat.format(field_value,
-                                            settings.SHORT_DATE_FORMAT)
+        elif data_type.startswith('date'):
```
jwalgran: Is the change from `==` to `startswith` meant to catch both "date" and "datetime"?

opentreemap/treemap/models.py
RickMohr: Good point. Added `self._do_not_track.add('last_seen')`

opentreemap/treemap/templatetags/form_extras.py
RickMohr: Changed to `data_type in ['date', 'datetime']`

jwalgran:
:+1:

RickMohr:
waffle

PR #2473, opened 2016-01-27
Rename instance.bounds_obj

steventlamb:
As part of a multiple commit, blue-green safe migration, instance.bounds
has been replaced with a one-to-one foreign key, bounds_obj, which is
now being renamed to bounds. Effectively, this "moves" instance.bounds
to another table.

See the commits on
https://github.com/OpenTreeMap/otm-core/commit/600f163eb0963e45cf0513515bb0d98e45077082
for more details.

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -158,9 +158,9 @@ class Instance(models.Model):
         'BenefitCurrencyConversion', null=True, blank=True)
 
     """ Center of the map when loading the instance """
-    bounds_obj = models.OneToOneField(InstanceBounds,
-                                      on_delete=models.CASCADE,
-                                      null=True, blank=True)
+    bounds = models.OneToOneField(InstanceBounds,
+                                  on_delete=models.CASCADE,
+                                  null=True, blank=True)
 
     """
     Override the center location (which is, by default,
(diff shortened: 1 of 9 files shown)


opentreemap/treemap/instance.py
```diff
@@ -158,9 +158,9 @@ class Instance(models.Model):
         'BenefitCurrencyConversion', null=True, blank=True)
 
     """ Center of the map when loading the instance """
-    bounds_obj = models.OneToOneField(InstanceBounds,
-                                      on_delete=models.CASCADE,
-                                      null=True, blank=True)
+    bounds = models.OneToOneField(InstanceBounds,
```
steventlamb: I think we'll want another migration to make this no longer nullable.

steventlamb: I think this still needs another migration that removes the old bounds field first.

jwalgran:
grep does not find any lingering instances of `bounds_obj`

:+1:

steventlamb:
this has been redone to have the correct migrations, ready for one last look and then to merge

jwalgran:
I migrated back to treemap 0014 and got the bounds for an instance:

```sql
SELECT ST_AsText(bounds) 
FROM treemap_instance 
WHERE id = 22;
```

I then migrated up through treemap 0021 and got the same bounds geometry back with this query:

```sql
SELECT ST_AsText(geom)
FROM treemap_instancebounds 
WHERE id = (
    SELECT bounds_id 
    FROM treemap_instance
    WHERE id = 22
);
```

:+1:

Issue #2459, opened 2016-01-18
Unhandled exception is thrown if the request passed to the start_import view does not have FILES


jwalgran:
We expect this view to always be invoked with an uploaded file, but we have logged and unhandled ` IndexError: list index out of range` exception when trying to get the filename of the first uploaded file.

```python
filename = files.keys()[0]
```

https://github.com/OpenTreeMap/otm-core/blob/666ba48c09049439b8a1ec764e43da505852ac38/opentreemap/importer/views.py#L654-L655

PR #2475, opened 2016-01-27
Fix 500 when import started with no attachment

steventlamb:
connects #2459 on github

### DIFF
--- a/opentreemap/importer/tests.py
+++ b/opentreemap/importer/tests.py
@@ -17,7 +17,7 @@ from django.conf import settings
 from django.db import connection
 from django.test import TestCase
 from django.test.utils import override_settings
-from django.http import HttpRequest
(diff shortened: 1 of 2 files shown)


RickMohr:
Consider simplifying this by putting the test before `process_csv` instead of inside it, and changing your test to call `start_import` and make its request with
```python
request = make_request({'type': tree})
```

RickMohr:
+1 I didn't come up with a good way to test this, but I think the unit test is satisfactory.

steventlamb:
I tested this by modifying the JS UI so that the import button is always available, and sent requests without an attachment. The `400` came back in the network tab with the appropriate message.

RickMohr:
I'll try that.

RickMohr:
I saw the 400 response with "No Attachment Received"

PR #2479, opened 2016-01-28
Make Street View show what you want to see

RickMohr:
Using camera location of Street View image along with tree location, set panorama heading so you're looking at the tree.

To test:
* Find a place you're familiar with, that has trees, on some tree map
* Select a tree and "View All Details" -- Street View should be pointing at the tree
* Click another nearby tree -- Street View should point to it

Or:
* Find a place you're familiar with, that has trees, on some tree map
* Click "Add a Tree"
* Place the marker (Street View appears)
* Move the marker to different places on the street -- Street View should point to it

### DIFF
--- a/opentreemap/treemap/js/src/streetView.js
+++ b/opentreemap/treemap/js/src/streetView.js
@@ -31,15 +31,23 @@ exports.create = function(options) {
             if (status == google.maps.StreetViewStatus.OK) {
                 if (panorama === null) {
                     panorama = new google.maps.StreetViewPanorama(div, {
-                        position:pos,
+                        position: pos,
                         addressControl: showAddress,
                         panControl: false,
                     });
                 } else {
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/js/src/streetView.js
```diff
@@ -31,15 +31,21 @@ exports.create = function(options) {
...
                         addressControl: showAddress,
                         panControl: false,
                     });
                 } else {
                     panorama.setPosition(pos);
                 }
-            }
-            else {
+                if (data !== null) {
+                    var cameraPos = data.location.latLng,
```
steventlamb: it took a verbal explanation to understand how this works, implies some knowledge about the google street (the concept of "camera position"). please consider adding a comment to the source code explaining at a high level what this does.

steventlamb:
+1 tested

PR #2484, opened 2016-02-01
Clear typeaheads when switching out of add tree mode

maurizi:
Fixes an issue where when selecting a species or address after having
already added a tree, the only results shown would be for the tree
previously added.

By explicitly clearing the typeahead when deactivating add tree mode or
clearing edit controls, the problem is fixed.

Connects to OpenTreeMap/otm-addons#1025


RickMohr:
+1, tested.

It might be a good idea to wait until 2.2 changes are merged to `develop` before merging this, so the person doing the release isn't faced with a tricky merge.

(...and thanks for essentially offering to merge it instead of me.)

maurizi:
Closing to reopen against the release branch

PR #2486, opened 2016-02-01
[DO NOT MERGE] Revert "Revert "Add coarse values to importer rate limiting""

steventlamb:
Reverts OpenTreeMap/otm-core#2485

This is the official feature branch for importer rate limit tuning. It doubled reverts because it was merged into develop and then removed. It should not be merged until a release branch has been created and it has been tested against a production-like stack.


steventlamb:
This should probably be closed, but it has some valuable data in the commit messages. These should be noted elsewhere.

PR #2487, opened 2016-02-02
Improve geocoder + typeahead experience

RickMohr:
This discussion focuses on the OTM top search box, but mostly applies as well to the address search boxes for adding a tree and creating a new instance.

##### Front End #####
Front end handles geocode search with and without an autocomplete suggestion. Distinguish 3 cases when user initiates a search:

1. User picked a geocoder suggestion. Call geocode endpoint, passing magic key and address text.
2. User picked a non-geocoder choice (e.g. an OTM boundary). Don't call geocode endpoint.
3. User didn't pick an autocomplete suggestion, or did but then edited it. Call geocode endpoint, passing just address text.

Also, when calling geocoder pass instance bounding box so it can be used for the non-magic-key case.

##### Back End #####
Back end handles geocode search with and without a magic key. When geocoding with a magic key (i.e. user chose an autocomplete suggestion)
* Use `settings.OMGEO_SETTINGS_FOR_MAGIC_KEY`
* Return only the top hit (except for "point of interest" searches)

When geocoding without a magic key (i.e. user did not choose an autocomplete suggestion)
* Use `settings.OMGEO_SETTINGS`
* Create geocoder with `postprocessors=[]`, and use fewer source-specific postprocessors (otherwise only non-specific results are returned)
* Restore bounding box code removed in 45ddc6d

##### Testing #####
* Type nothing and hit Enter. The geocoder is not called.
* Type "340 12th St, Philadelphia
  * Hit enter. You get bold 3 choices because you chose no suggestion and your query isn't specific enough for the geocoder.
  * Click "340 N 12th St, Philadelphia, Pennsylvania, USA" and "Search". You zoom to Azavea.
* Type "Brewery"
  * Hit enter. You get an error because you chose no suggestion and your query isn't specific enough for the geocoder.
  * Click "Brewerytown / Philadelphia Neighborhood" and "Search". You zoom to the Brewerytown neighborhood because the suggestion you chose is an OTM boundary, not an ESRI suggestion.
  * Click "Brewery" and "Search". You get a list of breweries.
  * Click "Brewerytown, Pennsylvania, United States" and "Search" You zoom to the center of Brewerytown because the suggestion you chose is an ESRI suggestion, and we always zoom to the same level with those.

Also test the address search boxes for adding a tree and creating a new instance.

Related to OpenTreeMap/otm-addons#1069
Connects OpenTreeMap/otm-addons#1062


maurizi:
Can you reopen this against `release/2.2.0`?

RickMohr:
Already working on it. Closing to target the release branch.

Issue #1025, opened 2014-01-16
OTM1 migrator should sanitize values that came from otm1


steventlamb:
Right now, there are bits on both side that handle sanitizing or transforming individual values. This should be standardized and checked for completeness.

PR #2494, opened 2016-02-03
Clear typeaheads when switching out of add tree mode

maurizi:
Fixes an issue where when selecting a species or address after having
already added a tree, the only results shown would be for the tree
previously added.

By explicitly clearing the typeahead when deactivating add tree mode or
clearing edit controls, the problem is fixed.

Connects to #1025


RickMohr:
+1, tested

PR #2503, opened 2016-02-11
Feature/edit warnings

steventlamb:


RickMohr:
We discussed using `options.shouldUpdate` instead of `options.update`.

Also having the modal template start with appropriate customization warnings in case the user removes customization. The warning would be absent if the treemap doesn't have customization, or if removing it would lose no data.

opentreemap/treemap/js/src/inlineEditForm.js
```diff
@@ -174,7 +164,36 @@ exports.init = function(options) {
...
+                //          of the data object
+                // * an eventStream - for blocking side effects,
+                //                    failure cases and early exits.
+                //
+                // when providing an eventStream as the return value for
+                // onSaveBefore, it should meet the following criteria:
+                // * it should be a stream of data objects. If you are using
+                //   eventStreams to block on IO, map them to the `data`
+                //   object provided.
+                // * it should fire an event when it's ok to proceed and
```
jwalgran: The phrase "fire an event" might be confusing here. If I correctly understand this new API, "It should push a new stream value..." may be a better description of the requirement.

jwalgran:
:+1:

Issue #2497, opened 2016-02-04
Stewardship search should not allow date searches until the drop downs are filled in


maurizi:
Searching for stewardship but only filling in the date range and not choosing values for the drop downs causes an invalid search that causes the ecobackend to errror.

We can fix this by disabling setting a date until the drop downs are filled (and disable them again if they are cleared).

PR #2508, opened 2016-02-12
Conditionally disable collection UDF date search fields

jwalgran:
The advanced search UI was allowing the selection of a date range for stewardship/alerts without selecting a type of either stewardship or alert. We don't support searching across collection UDF types, so this resulted in the code generating invalid filter query strings.

This commit addresses the issue by following the existing patterns in `udfcSearch.js` and disabling the date fields until both a model and a CUDF type have been selected.

##### Testing

The date fields at the end of the "Tree Care" section of the advanced search bar should be disabled until values have been selected from the first two dropdowns.

screen shot 2016-02-12 at 11 38 28 am

screen shot 2016-02-12 at 11 38 39 am

---

Connects to #2497


jwalgran:
I originally assigned this to Steve because he has the most ownership of `udfcSearch.js` but I just noticed has a lot of other pulls assigned to him.

RickMohr:
+1, works well.

PR #2511, opened 2016-02-17
Support ESRI basemaps

RickMohr:
`esri-leaflet` makes it easy.

But note that including `esri-leaflet` increases our JS download by 56K (6%). Once we switch to Leaflet 1.0 we can reduce that greatly using https://github.com/esri/esri-leaflet-bundler to grab just the basemap layer code.

For the default basemap I chose ESRI "Topographic" because i thought it looked better for a treemap than "Streets".

Google:
image

ESRI Streets:
image

ESRI Topographic:
image

#### Testing

Change your favorite instance to use ESRI basemaps via the Django shell:
```
manage.py shell_plus
i = Instance.objects.get(url_name='...')
i.basemap_type = 'esri'
i.save()
```

Connects OpenTreeMap/otm-addons#1079

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -125,6 +125,7 @@ class Instance(models.Model):
     basemap_type = models.CharField(max_length=255,
                                     choices=(("google", "Google"),
                                              ("bing", "Bing"),
+                                             ("esri", "ESRI"),
                                              ("tms", "Tile Map Service")),
                                     default="google")
     basemap_data = models.CharField(max_length=255, null=True, blank=True)
--- a/opentreemap/treemap/js/src/MapManager.js
+++ b/opentreemap/treemap/js/src/MapManager.js
@@ -14,12 +14,14 @@ var $ = require('jquery'),
     layersLib = require('treemap/layers'),
 
     MIN_ZOOM_OPTION = layersLib.MIN_ZOOM_OPTION,
-    MAX_ZOOM_OPTION = layersLib.MAX_ZOOM_OPTION;
+    MAX_ZOOM_OPTION = layersLib.MAX_ZOOM_OPTION,
+    BASE_LAYER_OPTION = layersLib.BASE_LAYER_OPTION;
 
 // Leaflet extensions
 require('utfgrid');
 require('leafletbing');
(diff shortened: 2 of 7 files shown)


jwalgran:
jshint failed the CI builds

```
Running "jshint:treemap" (jshint) task
Linting treemap/js/src/MapManager.js ...ERROR
[L218:C21] W015: Expected 'L' to have an indentation at 13 instead at 21.
                    L.esri.basemapLayer("Imagery", MAX_ZOOM_OPTION),
Linting treemap/js/src/MapManager.js ...ERROR
[L219:C21] W015: Expected 'L' to have an indentation at 13 instead at 21.
                    L.esri.basemapLayer("ImageryTransportation", MAX_ZOOM_OPTION)]),
Linting treemap/js/src/MapManager.js ...ERROR
[L219:C82] W015: Expected ']' to have an indentation at 9 instead at 82.
                    L.esri.basemapLayer("ImageryTransportation", MAX_ZOOM_OPTION)]),
Warning: Task "jshint:treemap" failed. Use --force to continue.
```

jwalgran:
It looks like the tree layer is below the ESRI layers.

screen shot 2016-02-18 at 9 33 21 am

RickMohr:
@jwalgran, comments addressed.

jwalgran:
I am still running into two problems with layer ordering.

- When I load the initial map, I do not see the tree layer. It only appears after I zoom in or out one level.
- If I switch to another layer type (i.e. Hybrid) the tree layer is no longer on top, and I cannot see it back on top until I refresh the page.

RickMohr:
@jwalgran, sorry for the multiple iterations. I think this one works.

jwalgran:
The map feature layer is now consistently above the basemaps, but If you select a zip code from "Search by Location" the boundary layer briefly appears but is then covered up by the basemap tiles.

I like the clean way in which you introduced the z-index option.

CI builds are failing due to a migration number collision.

RickMohr:
All z indexes now live in `layersLib.js` ... migrations merged ... this time for sure.

jwalgran:
Works perfectly and I don't see any regressions in the traditional Google basemap. Thanks for slogging through it.

:+1:

Issue #2500, opened 2016-02-09
Queries to updating TreeImportEvent during an import can take over 10 seconds to complete


jwalgran:
During isolated, single-user testing of tree imports we logged database update queries taking over 10 seconds. A normalized example:

```
UPDATE "importer_treeimportevent"
SET
    "file_name" = '',
    "errors" = '',
    "field_order" = '',
    "owner_id" = 0,
    "instance_id" = 0,
    "created" = '' ::timestamptz,
    "completed" = NULL,
    "status" = 0,
    "last_processed_at" = '' ::timestamptz,
    "is_lost" = FALSE,
    "task_id" = '',
    "plot_length_conversion_factor" = 0.0,
    "plot_width_conversion_factor" = 0.0,
    "diameter_conversion_factor" = 0.0,
    "tree_height_conversion_factor" = 0.0,
    "canopy_height_conversion_factor" = 0.0
WHERE
    "importer_treeimportevent"."id" = 0;
```

We suspect long running transactions may be blocking updates, possibly related to the high frequency at which we are updating `last_processed_at` while an import is in progress.

hectcastro:
Raw PostgreSQL logs, along with the pgBadger processed report:

- [postgresql.log.2016-02-09-03.zip](https://github.com/OpenTreeMap/otm-core/files/123904/postgresql.log.2016-02-09-03.zip)
- [postgresql.log.2016-02-09-04.zip](https://github.com/OpenTreeMap/otm-core/files/123905/postgresql.log.2016-02-09-04.zip)
- [pgbadger.html.zip](https://github.com/OpenTreeMap/otm-core/files/123906/pgbadger.html.zip)

steventlamb:
Our best guess so far is that `commit_rows` should not be wrapped with `transaction.atomic`, only `commit_row` should be.

RickMohr:
Also, we update `last_processed_at` for every row, for crash detection. It would be just as effective to update it only at the end of a chunk of rows.

PR #2515, opened 2016-02-18
Reduce transaction scope and heartbeat frequency

steventlamb:
connects #2500 on github

These two features together lead to a large increase in task runtime in
cases where multiple tasks were running in parallel. This is because
each task would hold a transaction with edits to one import_event record
for the entire life of the task. The result is that each task would have
to wait until a sibling task finishes its transaction. We were seeing
queries in excess of ten seconds as a result. By scoping the transaction to just the
length of the row edits, queries from other tasks need only block as long as it takes a
task to edit each row.

Additionally, the heartbeat was being run once per row (85 times per
task) when once per task is perfectly adequate.

It is suspected, but not confirmed, that the heartbeat is the particular
aspect of the row edit that was causing the lock on import_event that
was blocking other tasks. It is possible that performing either one of
these two changes would have fixed this issue. However, they are both
improvements so there is little benefit to further investigation.

To test:
1) run an import job with rate limiting at its current setting of
1/m. Note the time per job.
2) change rate limiting value to something like 15/m.
3) run an import job. If this fix works, the tasks run more frequently
but with the same average runtime.

### DIFF
--- a/opentreemap/importer/models/species.py
+++ b/opentreemap/importer/models/species.py
@@ -312,7 +312,6 @@ class SpeciesImportRow(GenericImportRow):
             self.status = SpeciesImportRow.VERIFIED
 
         self.save()
-        self.import_event.update_progress_timestamp_and_save()
         return not fatal
 
     def _prepare_merge_data(self):
@@ -419,5 +418,4 @@ class SpeciesImportRow(GenericImportRow):
 
         self.species = species
         self.status = SpeciesImportRow.SUCCESS
-        self.import_event.update_progress_timestamp_and_save()
         self.save()
--- a/opentreemap/importer/models/trees.py
+++ b/opentreemap/importer/models/trees.py
@@ -101,7 +101,6 @@ class TreeImportRow(GenericImportRow):
     def model_fields(self):
         return fields.trees
 
-    @transaction.atomic
     def commit_row(self):
         is_valid = self.validate_row()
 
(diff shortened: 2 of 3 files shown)


RickMohr:
Changes look good. I'd like to understand how you measured job time before I test.

steventlamb:
I measured job time by:
1) running `scripts/debugcelery.sh`, which produces log lines with job times. This is the main thing to watch when analyzing job times (as opposed to system resources, for which I used Vector)
2) used `manage.py random_csv` to create a dataset of about 10k trees
3) Adjusted my vagrantfile to add more CPU cores to my app VM. Two should be fine to replicate the transaction locking.
3) Adjusted the values in `default_settings.py` for rate limiting. Depending on your system resources, you should find some key values where the situation changes. For example, at the current setting of 1/m, jobs finish fast enough and with enough sparsity that two workers don't ever block. At a setting of 100/m (effectively no rate limiting), you should see job times increase (until you checkout this branch). When I was tuning I noticed interesting things happen across the 5/m-15/m range, but that shouldn't be necessary for this test.

Please let me know if you need more info.

RickMohr:
For the record, we observed that with 1 cpu, `debugcelery.sh` creates two celery worker processes, so changing the number of cpus shouldn't be necessary.

RickMohr:
+1 At `15/m` I saw 9-10 sec/task without these changes and 6-7 sec/task with these changes.

I did need to use 2 cpus -- with 1 cpu the same process got all the tasks.

PR #2517, opened 2016-02-23
Lazily cache itree codes

steventlamb:
Improves all cases in which these values aren't used


RickMohr:
+1, good sleuthing. Did not test.

Issue #2506, opened 2016-02-11
Do not send the empty choice/multichoice value from the API response


jwalgran:
The mobile applications have a dedicated "Clear" button, so we do not need to display an empty choice.

PR #2522, opened 2016-02-24
Mobile apps shouldn't show a blank choice for search fields

RickMohr:
We now have 3 cases for whether a choice/multichoice field should offer a blank choice:
* Always -- search fields on the web
* Never -- search fields on mobile
* Choice fields but not multichoice fields -- detail page fields

Connects #2506

Testing: verify that the 3 cases above work as expected

### DIFF
--- a/opentreemap/treemap/search_fields.py
+++ b/opentreemap/treemap/search_fields.py
@@ -171,8 +171,8 @@ def advanced_search_fields(instance, user):
 
 
 def mobile_search_fields(instance):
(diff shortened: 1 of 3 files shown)


RickMohr:
...Fixing failing tests...

RickMohr:
...tests fixed.

jwalgran:
The empty choices are gone from the iOS filter page, but I still see choices empty choices on the web app.

:+1:

PR #2524, opened 2016-03-01
Confirm onSaveBefore returns correct value

steventlamb:
This changeset does two things:
1) makes onSaveBefore return type checking more explicit and throws with a
helpful message.
2) make the default onSaveBefore behavior return the correct type

This was necessary because for versions of the inlineEditForm that do
not override onSaveBefore, the identity function was used. This is not
correct because onSaveBefore must now return a stream, or null.


jwalgran:
Works. The "type" checking should help catch future accidental misuse of `onSaveBefore` and nudge the developer in the right direction.

:+1:

Issue #2504, opened 2016-02-11
Create proposal for importer, exporter and presentation field names


RickMohr:
The result of this task should be a table with proposed field names for:
* csv exports
* bulk uploader import templates
* management site columns

Some investigation should go into:
* whether case sensitivity will be an issue in processing input
* conditional exclusion of model prefix based on the context
* fields that should be removed (tree ID)
* fields that should be renamed (plot -> planting site)

steventlamb:
Case sensitivity will not be an issue. The importer is very lenient. I made a udf called 'TheNum', and the following csv imports correctly and makes a plot with a val for "TheNum":
```
poiNt x,poInt y,plot: THeNum
-101.665420332,33.6443490445,3
```

General Strategies:
As discussed, there is no benefit to having database-style column headers in any CSV. Given that the importer handles spaces and case well enough, it makes sense to use similar, nicely formatted names everywhere. The exporter is flexible in how it handles changes to column headers, so this will not be difficult.
In general, Plot should be renamed to planting site in all cases.
Readability is lost if you remove model prefixes. For example 'Planting Site Length' makes more sense than 'Length'. The exporter and importer should have identical column headers, but it's fine for mgmt pages to remove the prefix when it is part of a model-specific table.

| tree export                  | import template     | mgmt           | plan                                                                                     |
| ------------------------------|---------------------|----------------|------------------------------------------------------------------------------------------|
| plot__geom__x                | Point X             | Geom           | Don't merge, go with 'Point X'                                                           |
| plot__geom__y                | Point Y             | Geom           | Don't merge, go with 'Point Y'                                                           |
| date_removed                 |                     | Date Removed   | Date Removed, add to import template                                                     |
| date_planted                 | Date Planted        | Date Planted   | Date Planted                                                                             |
| diameter                     | Diameter            | Diameter       | Diameter                                                                                 |
| plot__width                  | Plot Width          | Width          | Planting Site Width                                                                      |
| plot__id                     |                     | Plot (also Id) | Don't fully merge                                                                        |
| plot__length                 | Plot Length         | Length         | Planting Site Length                                                                     |
| species__cultivar            | Cultivar            |                | Cultivar                                                                                 |
| plot__readonly               |                     | Readonly       | ? This is a tough one, hard to differentiate model without banging into UDF recognition. |
| readonly                     |                     | Readonly       | ? This is a tough one, hard to differentiate model without banging into UDF recognition. |
[...]

PR #2528, opened 2016-03-04
Feature/unify csv columns

steventlamb:
Connects #2504 
Connects #2505

### DIFF
--- a/opentreemap/exporter/tasks.py
+++ b/opentreemap/exporter/tasks.py
@@ -176,11 +176,11 @@ def async_csv_export(job, model, query, display_filters):
                           + sorted(values_sp))
 
         if ordered_fields:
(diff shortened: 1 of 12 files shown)


steventlamb:
This has at least one known outstanding issue:
- `tree__plot` is the only export field that was observed to reliably contains a plot ID, so it was promoted to the primary plot ID field. This doesn't work for plots that don't have trees. This must be fixed.

it probably also has the following:
- will not work for instances with alerts

it's been suggested to add the following:
- add unit tests that exports and then imports

RickMohr:
waffle

RickMohr:
waffle

RickMohr:
closing in favor of #2533

Issue #1445, opened 2014-06-25
Update robots.txt to direct indexers to content that should be indexed


jwalgran:
We are seeing errors when the googlebot hits filter urls that appear on the instance/edits page. These should not really be errors, but it raises the larger issue of what should we tell robots to index. I think there are only three groups of pages that robots should be indexing:

* instance pages (community, faq, about)
* map feature detail pages
* public user profile pages

PR #2529, opened 2016-03-08
Update robots.txt

jwalgran:
When robots.txt is empty, search crawlers request every linked page on the site. Only a small subset of pages on the site have content that would be useful in a general web search context:

- Map feature detail pages
- Public user profile pages
- Instance custom pages

It is important that the /instance/edits/ page is disallowed because it is a paginated list of every edit and can result in millions of crawler requests if the instance is large.

These URLs should be allowed:
/
/map/
/engage/
/analyze/
/mobile/
/pricing/
/instance/features/12345/
/instance/page/About/
/user/person/

These should be disallowed:
/instance/
/instance/map/
/instance/edits/
/instance/management/
/instance/management/a-management-sub-page/

--- 

##### Testing

The Googlebot testing tool allows you to paste in the content of robots.txt and verify individual addresses
https://www.google.com/webmasters/tools/robots-testing-tool

Note: You need to log in with the Google account that controls analytics for the domain to use the test tool.

screen shot 2016-03-08 at 10 37 06 am

I also ran the Yandex tester, which allows testing in bulk
https://webmaster.yandex.com/robots.xml

screen shot 2016-03-08 at 9 18 38 am

---

Connects to #1445

### DIFF
--- a/opentreemap/treemap/static/robots.txt
+++ b/opentreemap/treemap/static/robots.txt
@@ -1,1 +1,27 @@
+User-agent: *
 
+# Allow the home page to be indexed.
+Allow: /$
+
+# Allow marketing pages to be indexed.
+Allow: /map/$
+Allow: /engage/$
+Allow: /analyze/$
+Allow: /mobile/$
+Allow: /pricing/$
+
+# Allow user profile pages to be indexed. If the user makes their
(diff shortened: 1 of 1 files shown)


jwalgran:
It makes more sense to keep this in a different repository.

PR #2530, opened 2016-03-08
Skip proximity validation on imports with IDs

jwalgran:
Proximity validation can prevent instance admins from correcting the locations previously uploaded trees in bulk.

---

##### Testing

1. Create an instance centered on Reno, NV. Activate it so that exports are available.
1. Import [mgpoints.csv](https://github.com/OpenTreeMap/otm-core/files/164087/mgpoints.csv.zip)
1. Export all the planting sites and delete all the columns from the exported csv except for the `x` , `y`, and `id` columns. Change the column headers to `point x`, `point y`, and `opentreemap plot id`.
1. Shuffle the values in the `opentreemap plot id` column. (I used "Sort" in LibreOffice Calc to reverse them)
1. Import the edited file.

This fails on the `develop` branch, but should succeed on this branch.

---

Connects to https://github.com/OpenTreeMap/otm-clients/issues/183
Connects to https://github.com/OpenTreeMap/otm-clients/issues/187

### DIFF
--- a/opentreemap/importer/models/trees.py
+++ b/opentreemap/importer/models/trees.py
@@ -269,6 +269,15 @@ class TreeImportRow(GenericImportRow):
         return True
 
     def validate_proximity(self, point):
+        # This block must stay at the top of the function and
+        # effectively disables proximity validation when the import
+        # row includes an OTM plot id. Proximity validation can
+        # prevent instance admins from correcting the locations of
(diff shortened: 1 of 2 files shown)


RickMohr:
I imported that block of trees right over itself without errors. And when I deleted the `opentreemap plot id` on two rows I got two errors. Nice to see a simple fix and a test.
:+1:

PR #2534, opened 2016-03-14
Push URL state to app when state is pushed or replaced, not just when popped

RickMohr:
When we updated `urlState.HistoryApi` to use the browser `History` API instead of `native.history.js` the changes were incomplete (https://github.com/OpenTreeMap/otm-core/pull/2435). The latter called its `onStateChange` callback for any change to the URL -- push, replace, or pop. The former has only a hook for pop.

The fix is to call the `onStateChange` callback ourselves when pushing or replacing URL state.

Connects OpenTreeMap/otm-addons#1114


jwalgran:
:+1:

Issue #189, opened 2013-08-14
fontello.css references font files that do not exist


jwalgran:
Test in 9d8901b528

The chrome console is showing 404 requests for http://ec2-54-224-212-206.compute-1.amazonaws.com/static/font/fontello.svg

PR #2540, opened 2016-03-30
Imports respect instance units

RickMohr:
* Fetch and use instance's units for tree imports
* Refactor nearby code
* Respect unit conversions when validating against species max diameter/height
* Make tests actually test these things

Also tweak units warning:
* Show tree units first
* "Plot" -> "Planting Site"
* Tweak spacing

To test:
1) Make an instance centered on Minneapolis
2) Change units to metric for tree and plot fields (cm for diameter, m for the rest)
3) Import `//fileshare/users/rmohr/notes/otm/importer/minneapolis1.csv`
4) Verify values are correct by looking at tree detail

Connects #189


maurizi:
:+1: Yay tests!

Issue #2532, opened 2016-03-10
Trailing white space in import CSV headers prevents viewing a verified import


jwalgran:
To reproduce, import a file with the header "point y " and, after verification, click the "View" link
```
ERROR:django.request:Internal Server Error: /jw-2016-03-10/importer/status/tree/575/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/decorators.py", line 45, in wrapper
    return view_fn(request, instance, *args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/decorators.py", line 69, in f
    return view_fn(request, instance, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/http.py", line 45, in inner
    return func(request, *args, **kwargs)
  File "/opt/app/core/opentreemap/importer/views.py", line 296, in show_import_status
    return render_to_response(template, ctx, RequestContext(request))
  File "/usr/local/lib/python2.7/dist-packages/django/shortcuts.py", line 45, in render_to_response
    using=using)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader.py", line 115, in render_to_string
    template_name, context, context_instance, dirs, dictionary)
  File "/usr/local/lib/python2.7/dist-packages/django/template/engine.py", line 221, in render_to_string
    return t.render(context_instance)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 210, in render
    return self._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 202, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 905, in render
    bit = self.render_node(node, context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/debug.py", line 79, in render_node
    return node.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader_tags.py", line 135, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 202, in _render
[...]

PR #2541, opened 2016-03-31
On import, strip leading and trailing blanks from headers

RickMohr:
Also strip when reading headers from DB since unstripped ones were saved in the past

Connects #2532


jwalgran:
On the `develop` branch I imported a file with extra whitespace in the headers and reproduced the error. On this branch I was able to see and commit the previously uploaded file, and upload and commit another file with whitespace in the headers

:+1:

Issue #2149, opened 2015-07-14
Address fields should not be nullable


RickMohr:
`MapFeature` fields `address_street`, `address_city`, and `address_zip` are currently defined as `null=True` but they should not be. Django text fields should be `blank=True` but not `null=True`.

PR #2544, opened 2016-03-31
Remove null=True from MapFeature address fields.

maurizi:
Connects to #2149


maurizi:
I'm investigating the unit test failures now, will re-assign this once they're fixed.

maurizi:
I did some digging and the source of the unit test failures lies in the audit and pending audit system.  I'm inclined to punt on these changes for now.

PR #2545, opened 2016-03-31
Fix crashes on user page for users whose username is all numbers

maurizi:
Connects to #2521


opentreemap/treemap/templatetags/auth_extras.py
```diff
@@ -197,7 +202,7 @@ def render(self, context):
             user_identifier = self.user_identifier
 
         user_content = self.nodelist.render(context)
-        if str(user_identifier).isdigit():
+        if isinstance(user_identifier, (int, long)):
             if req_user.pk == int(user_identifier):
```
jwalgran: It looks like the `int()` casting here is now redundant.

jwalgran:
:+1:

Issue #2463, opened 2016-01-19
Long field names in the missing data search dropdown do not wrap


maurizi:
<img width="515" alt="nowrap" src="https://cloud.githubusercontent.com/assets/4432106/12433048/e39bef32-becc-11e5-93fc-514ad9bc01d4.PNG">

maurizi:
This appears to be browser specific - it wraps in Chrome but not in firefox

PR #2546, opened 2016-03-31
Fix the styling of long field names in advanced search dropdowns

maurizi:
Firefox was not respecting the 0 width style, which caused the checkboxes
to have a padding on the left for the "invisible" checkboxes, *unless*
they had text long enough to wrap.

By using fixed positioning to place the inputs off the screen, we get
around browser inconsistencies with width, while still keeping the inputs
functionally invisible, but still enabled.

Connects to #2463

### DIFF
--- a/opentreemap/treemap/css/sass/modules/_checkboxes.scss
+++ b/opentreemap/treemap/css/sass/modules/_checkboxes.scss
@@ -1,8 +1,8 @@
 @mixin checkboxes {
-	input[type='radio'],
+    input[type='radio'],
     input[type='checkbox'] {
         opacity: 0;
-        float: left;
+        position: absolute;
         width: 0;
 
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/css/sass/modules/_checkboxes.scss
```diff
@@ -1,8 +1,9 @@
 @mixin checkboxes {
-	input[type='radio'],
+    input[type='radio'],
     input[type='checkbox'] {
         opacity: 0;
-        float: left;
+        position: fixed;
+        left: -100px;
         width: 0;
```
designmatty: I'm fairly certain that if instead of doing the changes you made here, you add `position: absolute;` the problem will be solved.

Issue #2519, opened 2016-02-23
Editing empty plot creates tree if instance has a multichoice custom tree field


RickMohr:
1) Add a multichoice tree field to an instance
2) Go to plot detail for an empty plot
3) Edit, and save

Expected: no tree is created
Actual: a tree is created

RickMohr:
Diagnosis:
* Even though no tree fields are visible the client request includes a value of `[]` for the custom tree field
* The back end creates a tree at https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/views/map_feature.py#L254

PR #2547, opened 2016-03-31
Only PUT multichoice field value if it has changed

RickMohr:
Previously when hitting "Save" on a detail page we PUT values for all multichoice fields even if those values had not changed.

That caused us to incorrectly create a tree when saving an empty planting site.

Now we correctly detect changes to multichoice fields and only PUT changed values.

Connects #2519

Testing:

1. Create a custom multichoice tree field
2. Open your developer tools network panel
2. On a tree detail page repeatedly Edit, change the value, and Save. Verify in the network panel that the correct
values are sent.
3. Delete the tree
4. Edit and Save. No tree should be created.


maurizi:
:+1:

Issue #2420, opened 2016-01-08
Illegal searches possible that crash results


steventlamb:
With the inclusion of "missing data" searches, it is possible to search for a field value but also search for mapfeatures that are missing a value for that field. This produces a mixture of a `LIKE` and `ISNULL` search, which is disallowed on the server and tiler and leads to a crash without feedback.

To fix, we should either:
1) Allow `ISNULL` and `LIKE`. Why not? Just because it doesn't make sense doesn't mean it should be illegal. (easy)
2) Add search validation to the client to prevent submitting this earch (hard)

At a glance (1) works with a simple change to the config, but we should investigate if there is a regression and/or a reason that this was not supported in the first place.

Exceptions:

appserver:
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app/core/opentreemap/treemap/decorators.py", line 45, in wrapper
    return view_fn(request, instance, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/http.py", line 158, in inner
    response = func(request, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 79, in wrapper
    params = callable_or_dict(request, *args, **wrapper_kwargs)
  File "/opt/app/core/opentreemap/treemap/views/tree.py", line 67, in search_tree_benefits
    total_plots = get_cached_plot_count(filter)
  File "/opt/app/core/opentreemap/treemap/ecocache.py", line 35, in get_cached_plot_count
    return _get_or_compute(prefix, filter, compute_value)
  File "/opt/app/core/opentreemap/treemap/ecocache.py", line 45, in _get_or_compute
    value = compute_value()
  File "/opt/app/core/opentreemap/treemap/ecocache.py", line 33, in <lambda>
    compute_value = lambda: filter.get_object_count(Plot)
  File "/opt/app/core/opentreemap/treemap/search.py", line 81, in get_object_count
    return self.get_objects(ModelClass).count()
  File "/opt/app/core/opentreemap/treemap/search.py", line 65, in get_objects
    q = create_filter(self.instance, self.filterstr, DEFAULT_MAPPING)
  File "/opt/app/core/opentreemap/treemap/search.py", line 161, in create_filter
[...]

PR #2549, opened 2016-04-01
Disable conflicting search fields when fields are changed

maurizi:
This prevents searches that don't make much sense, such as searching for
a specific species and a tree missing a species at the same time.

Connects to #2420

screenshot from 2016-03-31 22-49-44
screenshot from 2016-03-31 22-50-11
screenshot from 2016-03-31 23-00-14


RickMohr:
Hey, now we can remove the (lame) message "Checking a field's box below overrides any other filter for that field"!

RickMohr:
For fields not in the "missing" category I think a message like this would be better:

`Disabled while a "Missing" search for this field is active.`

opentreemap/treemap/js/src/searchBar.js
```diff
@@ -224,18 +238,63 @@ function updateDisabledFieldGroups(search) {
...
+    // Then disable all fields which are not filled in but which have the same
+    // key as another filled in field
+    _.each(search.filter, function(predicate, field) {
+        var searchTypes = _.keys(predicate),
+            minOrMax = _.contains(searchTypes, 'MIN') || _.contains(searchTypes, 'MAX');
+        $(dom.searchFields).filter('[name="' + field + '"]').each(function(i, elem) {
+            var $elem = $(elem),
+                searchType = $elem.attr('data-search-type');
+
+            // Min/Max fields don't affect other min/Max fields
```
RickMohr: I didn't get this on first reading. Maybe something like "e.g. when user fills in the min field, don't disable the corresponding max field."

RickMohr:
And since there can only be one search for a field I think the other should say:

`Disabled while a search for this field is active`

opentreemap/treemap/js/src/searchBar.js
```diff
@@ -224,18 +238,63 @@ function updateDisabledFieldGroups(search) {
...
+            if (minOrMax && _.contains(['MIN', 'MAX'], searchType)) {
+                return;
+            }
+
+            if (($elem.is(':checkbox') && !$elem.is(':checked')) || $elem.val().length === 0) {
```
RickMohr: If I check the "Missing" box for a multiselect field I get a JS error `"Cannot read property 'length' of null"`

RickMohr:
The "disabled" message is indented (and thus wraps) when the dropdown has a scrollbar:
image

RickMohr:
"Disabled" message is indented (and thus wraps) on text inputs:
image

RickMohr:
Otherwise works great.

maurizi:
@RickMohr all fixed

RickMohr:
To me, this is confusing to parse:
`Disabled while a missing search for this field is active`
(it sounds like the search is missing) which is why I'd prefer:
`Disabled while a "Missing" search for this field is active`

RickMohr:
+1 with consideration of quote fix. Yay!

PR #2551, opened 2016-04-04
Fix random_trees command delete option and add delete_data command

maurizi:
The --delete option to random_trees was incorrectly attempting to delete
all trees from all maps, not just the map associated with the command.

I switched the option to delete via SQL for better performance and so that
it could delete audits and other related objects as well.

I also added a delete_data command, which deletes all trees, plots, map
features and other related data from the given map, leaving mainly only
the import history and customizations in place.

Connects to #1139

### DIFF
--- a/opentreemap/treemap/management/commands/delete_data.py
+++ b/opentreemap/treemap/management/commands/delete_data.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from treemap.management.util import InstanceDataCommand
+
+
+class Command(InstanceDataCommand):
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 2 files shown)


RickMohr:
This worked well to delete the 1672 trees in my `miami-beach-demo` instance. 

I verified that photos are also deleted.

It took several minutes, meaning it could take several hours for a larger map.

Seems to me the goal here is to start fresh, so we don't care about creating audits for the deletes, and actually want to **remove** audits for the deleted objects so they don't show up in the Edits list. If we decide we do want to leave the audits in place (not my preference) we should make sure they don't break the Edits UI.

I think we should consider a transaction with 4 SQL deletes -- tree/plot/mapfeature/audit.

RickMohr:
+1, works great, code looks good.

So painful to do bulk deletes in Django!

Issue #2555, opened 2016-04-13
`cast_to_subtype` may fail if `self` is not a `MapFeaure`


RickMohr:
In this rollbar issue https://rollbar.com/Azavea/OpenTreeMap/items/916/
we've instantiated a `MapFeature` whose `feature_type` is `"Plot"`, and used its `plot` attribute to get a `Plot` object. The `Plot` object's `feature_type` is also `"Plot"`, but it has no `plot` attribute.
That should not be possible. Django seems to be constructing the object incorrectly. 

So far we have only reproduced this in production (release 2.5.0) when a user is logged in.

hectcastro:
Related details:

- All cases of this exception appear to have an authenticated user associated
- All cases are also related to either the Greenprint Maps or Carson City map instances
- Django and its dependencies were all at the same versions across 2.4.0 and 2.5.0
- A similar exception has occurred in testing, staging, and production

PR #2556, opened 2016-04-13
Address issue with incorrect calls to MapFeature.cast_to_subtype

maurizi:
For unknown reasons, sometimes when `MapFeature.cast_to_subtype` is called
on Plot objects, they lack the `plot` attribute pointing from the
MapFeature superclass to the Plot subclass, causing `cast_to_subtype` to
incorrectly trest those objects as `PolygonalMapFeature` classes.

This works around the issue by having `cast_to_subtype` to just return
`self` if not called on the `MapFeature` superclass.

Connects to #2555


RickMohr:
+1

PR #2559, opened 2016-04-21
Fix styling regression in by Chrome version 50

maurizi:
Connects to OpenTreeMap/otm-clients#211

### DIFF
--- a/opentreemap/treemap/css/sass/partials/_subheader.scss
+++ b/opentreemap/treemap/css/sass/partials/_subheader.scss
@@ -31,14 +31,6 @@
         transition: .2s ease-in-out -webkit-transform;
         transition: .2s ease-in-out transform;
 
-        + .stats-bar {
-            position: absolute;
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/css/sass/partials/_subheader.scss
```diff
@@ -31,14 +31,6 @@
         transition: .2s ease-in-out -webkit-transform;
         transition: .2s ease-in-out transform;
 
-        + .stats-bar {
```
designmatty: Whats the reasoning behind this removal?

maurizi: It was using a more specific selector and overriding the `> .stats-bar` selector below, so I merged the two sections together, keeping everything but the `position: absolute`

maurizi:
Got a verbal :+1: from @designmatty

Issue #2558, opened 2016-04-18
When zooming or resizing the map page quickly the z= url parameter cycles and the browser tab becomes unresponsive


jmorrison1847:
**Macbook Pro running OSX Yosemite 10.10.3 in Chrome version 49.0.2623.112**

I was only able to duplicate this behavior when turning satellite imagery on, but the basics are:
1. Aggressively zoom in/out and move the visible part of the screen
2. The screen freezes and the URL starts going quickly back-and-forth between two states (the part that denotes x/y)
3. Chrome becomes unresponsive.

A screen recording of the behavior can be viewed here:https://drive.google.com/a/azavea.com/file/d/0B_E16nocP-BpYUM3WE1tZTRnVTg/view?usp=sharing

jwalgran:
I was able to reproduce this.

hadleystein1:
+1

jmorrison1847:
+1 from yegTreeMap

jwalgran:
I am going to adjust the title of this issue because "crash" made me think of a different behavior than the one we are seeing.

jmorrison1847:
+1 TreeMapLa; in this case, simply resizing the browser causes the behavior

PR #2560, opened 2016-04-22
Fix infinite loop of browser map state

maurizi:
When panning or zooming the map, we store the map state in the URL.
Conversely, when changing the URL, we pan or zoom the map to match the new URL.

Because we used a fixed precision when storing lat/lng in the URL, but did not
do so for the state object that stored the current zoom/lat/lng, it was
possible for changes to the map state to cause an infinite loop, where the URL
and map state would constantly change but never get in sync with each other.

By serializing and deserializing our internal state object whenever we change
the URL, we ensure that when we later compare the current state object to the
new state to try and detect changes we'll never encounter these loops.

Connects to #2558

### DIFF
--- a/opentreemap/treemap/js/test/urlState.js
+++ b/opentreemap/treemap/js/test/urlState.js
@@ -130,15 +130,15 @@ var urlStateTests = {
         var context = createContext();
         urlState.init(context);
         urlState.stateChangeStream.take(1).onValue(function(state) {
-            assert.deepEqual(state, {bar: 2});
+            assert.deepEqual(state, {bar: '2'});
             done();
         });
         // This should NOT trigger an update on stateChangeStream.
-        urlState.set('foo', 1, {
+        urlState.set('foo', '1', {
             silent: true
         });
         // This SHOULD trigger an update on stateChangeStream.
-        urlState.set('bar', 2, {
+        urlState.set('bar', '2', {
             silent: false
         });
     }
--- a/opentreemap/treemap/js/src/urlState.js
+++ b/opentreemap/treemap/js/src/urlState.js
@@ -126,6 +126,10 @@ function set(key, value, options) {
         var newState = _.extend({}, _state);
         newState[key] = value;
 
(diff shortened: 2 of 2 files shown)


opentreemap/treemap/js/test/urlState.js
```diff
@@ -130,15 +130,15 @@ var urlStateTests = {
         var context = createContext();
         urlState.init(context);
         urlState.stateChangeStream.take(1).onValue(function(state) {
-            assert.deepEqual(state, {bar: 2});
+            assert.deepEqual(state, {bar: '2'});
```
jwalgran: Why did we switch the numbers to strings in these tests?

maurizi: Adding in the line `newState = getStateFromUrl(getUrlFromState(newState));` means that we always serialize our URL state to text and then deserialize it back to a JS object.  For special-cased items that have a deserialize function builtin, like `zoomLatLng` we can convert it to a number, but for unrecognized items in the URL they'll always be text.

This was the case before when we loaded the page with those items in the URL, so this isn't changing behaviour too drastically.

jwalgran:
Did you explore adding a `toFixed` calls to the deserialization method as an alternative to round tripping?  https://github.com/maurizi/otm-core/blob/78a1213308cd534f5f2daa82b2263ced58b5e93c/opentreemap/treemap/js/src/urlState.js#L94-L95

jwalgran:
I could not reproduce the issue on this branch.

👍

maurizi:
> Did you explore adding a toFixed calls to the deserialization method as an alternative to round tripping? https://github.com/maurizi/otm-core/blob/78a1213308cd534f5f2daa82b2263ced58b5e93c/opentreemap/treemap/js/src/urlState.js#L94-L95

Yeah, I looked into it but ultimately decided this was the better way to deal with the situation.  Using `.toFixed` in the deserialization seemed to make the issue harder to trigger, but I was still able to do so with a lot of trying.

PR #2573, opened 2016-05-03
fixes issue for otm-addon repo with header buttons overflowing

designmatty:
this goes with https://github.com/OpenTreeMap/otm-addons/pull/1164


RickMohr:
+1

PR #2578, opened 2016-05-10
Fix broken comment moderation UI tests.

maurizi:

### DIFF
--- a/opentreemap/otm_comments/templates/otm_comments/partials/moderation.html
+++ b/opentreemap/otm_comments/templates/otm_comments/partials/moderation.html
@@ -8,7 +8,7 @@
   <div class="page-header-title">
       <h1>{% trans "Comments" %}</h1>
   </div>
-  <div class="page-header-actions">
+  <div class="page-header-actions pull-right">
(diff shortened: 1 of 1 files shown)


jwalgran:
For reference, this is the failed test I see in latest build of the `develop` branch

```
======================================================================
FAIL: test_filtering (otm_comments.uitests.CommentReviewUITest)
----------------------------------------------------------------------
  File "/opt/app/core/opentreemap/otm_comments/uitests.py", line 98, in test_filtering
    self.assert_num_rows(1, 'Only the hidden comment should be shown')
  File "/opt/app/core/opentreemap/otm_comments/uitests.py", line 58, in assert_num_rows
    self.assertEqual(num, len(rows), msg)
AssertionError: Only the hidden comment should be shown
```

maurizi:
@azavea-bot rebuild

jwalgran:
@azavea-bot rebuild

jwalgran:
The UI tests pass on my machine and I verified that this does not affect the usability of buttons in the upper right of the comment moderation page.

👍

RickMohr:
@maurizi, want to merge this so the build will work again?

PR #2579, opened 2016-05-10
Only delete map-feature related audits in the 'delete_data' command

maurizi:
Connects to OpenTreeMap/otm-addons#1158


jwalgran:
I double-checked this whitelist by querying all the distinct model names in the database and verifying that there were no additional ones we would want to keep.

👍

Issue #2569, opened 2016-04-27
Cannot select tree dot after searching by geographic area


dboyer:
Tested on Mac in Chrome and Safari. Same behavior.

1. Search by a geographic area (neighborhood, zip code)
2. Geographic area is shaded on the map with tree dots visible within the polygon
3. Click on a tree dot that falls within the polygon
4. Nothing happens

This is on production as well as staging so was not introduced in the newest release.

RickMohr:
Likely related to the changes in #2511 to support ESRI basemaps.

RickMohr:
@jmorrison1847 writes:

Noted by LA Tree Map--when filtering down to a neighborhood (or presumably any set of boundary files), the trees within that neighborhood are correctly summarized but are not clickable.

Try Beverly Hills in LA: https://www.opentreemap.org/latreemap/map/?z=12/34.0825/-118.3997&q=%7B%22mapFeature.geom%22%3A%7B%22IN_BOUNDARY%22%3A%2230798%22%7D%7D

Or China Town North in Philly: https://www.opentreemap.org/phillytreemap/map/?z=15/39.9588/-75.1507&q=%7B%22mapFeature.geom%22%3A%7B%22IN_BOUNDARY%22%3A%2231411%22%7D%7D

jmorrison1847:
Oop, did not see this had already been noted, thanks @RickMohr!

PR #2584, opened 2016-05-16
Make tile layers display above boundary polygon

RickMohr:
See discussion in code comments.

Connects #2569

### DIFF
--- a/opentreemap/treemap/js/src/boundarySelect.js
+++ b/opentreemap/treemap/js/src/boundarySelect.js
@@ -21,12 +21,25 @@ function showBoundaryGeomOnMapLayerAndZoom(map, boundaryGeom) {
     // We want to put the boundary polygon below the plot tiles
     // so that plots will display and select normally.
     // But in Leaflet < 0.8 all polygon layers go in the "overlay pane",
-    // so our only choice is to set the z-index of the entire overlay pane.
+    // which is above the tile pane:
+    //     Map pane
+    //         Tile pane (position absolute, z-index 2)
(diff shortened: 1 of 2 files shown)


RickMohr:
@azavea-bot rebuild

jwalgran:
On this branch the boundary layer is behind the ESRI basemap

screen shot 2016-05-17 at 1 32 28 pm

screen shot 2016-05-17 at 1 33 57 pm

RickMohr:
The Google basemap's tiles are in a `div` separate from the tree dot tiles:
```html
<div class="leaflet-google-layer" ...> ... </div>
<div class="leaflet-map-pane" ...> ...tree dot tiles... </div>
<div class="leaflet-objects-pane" ...> ...boundary polygon... </div>
```
That allows us to put the boundary polygon between the basemap and the tree dot tiles.

But the ESRI basemap's tiles are in the same `div` as the tree dot tiles:
```html
<div class="leaflet-map-pane" ...> ...ESRI basemap tiles AND tree dot tiles... </div>
<div class="leaflet-objects-pane" ...> ...boundary polygon... </div>
```
Because the `leaflet-map-pane` has position absolute it establishes a separate stacking context and there's no way to put the boundary polygon and the tree dot tiles.

Since before too long we will likely switch to Leaflet 1.0 (which allows much finer control over the layer stacking order) and since ESRI basemaps aren't being used yet, I think we should address the issue when we switch to Leaflet 1.0. Entered as #2586.

jwalgran:
👍

Issue #2576, opened 2016-05-05
Improve Google Analytics integration


maurizi:
We use Google Tag Manager to include our Google Analytics snippet.  We may need to upgrade to `analytics.js` (see https://developers.google.com/analytics/devguides/collection/upgrade/).

Things to do:
 - Include the user ID in hits if the user is logged in.
 - Created named trackers to differentiate `otm-core` from other websites sharing the same URL (https://developers.google.com/analytics/devguides/collection/analyticsjs/creating-trackers#the_create_method).
   - ~~One tracker that is `Instance`-specific, one that is `Instance` agnostic (removing `url_name` from the URL for page view hits, see https://developers.google.com/analytics/devguides/collection/analyticsjs/pages#modifying_page_urls)~~ After further investigation, I'm not sure it is possible to have 1 tracker per instance
 - Use https://github.com/googleanalytics/autotrack to add event tracking for specific buttons/links

maurizi:
Can test https://support.google.com/analytics/answer/1008083?hl=en

PR #2587, opened 2016-05-19
Update Google Analytics tracking code

maurizi:
- Switches from using Google Tag Manager to using `analytics.js` directly
 - Uses two trackers with different IDs for tracking events by map and in aggregate
 - Removes URLs from instance and user specific page hits to allow for aggregate analysis
 - Brings in the `autotrack` analytics plugin, and uses it to enable tracking of some of search and the add tree process.

Connects to #2576

### DIFF
--- a/opentreemap/treemap/templates/instance_base.html
+++ b/opentreemap/treemap/templates/instance_base.html
@@ -14,6 +14,17 @@
 {% endif %}
 {% endblock application_css %}
 
+{% block google_analytics_actions %}
+{# for instance-specific pages, remove the url_name from the url for better tracking #}
+var pathname = document.location.pathname;
+pathname = pathname.replace(/\/[a-zA-Z]+[a-zA-Z0-9\-]*\//, '/<instance_url_name>/');
+
+// Note: we're only removing the url_name on the app tracker, not the global tracker
+// This way we can still see which instances get more traffic and look at usage differences between them
+ga('set', 'page', pathname);
+{{ block.super }}
+{% endblock google_analytics_actions %}
+
 {% block instance_title %} | {{ request.instance.name }}{% endblock %}
 
 {% block instancetopnav %}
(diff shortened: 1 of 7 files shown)


opentreemap/treemap/templates/instance_base.html
```diff
@@ -14,6 +14,17 @@
 {% endif %}
 {% endblock application_css %}
 
+{% block google_analytics_actions %}
+{# for instance-specific pages, remove the url_name from the url for better tracking #}
+var pathname = document.location.pathname;
+pathname = pathname.replace(/\/[a-zA-Z]+[a-zA-Z0-9\-]*\//, '/<instance_url_name>/');
```
RickMohr: Perhaps clearer and safer:
```
pathname = pathname.replace(/^\/{{request.instance.url_name}}\//, '/<instance_url_name>/');
```

opentreemap/treemap/templates/instance_base.html
```diff
@@ -145,6 +156,7 @@
       <a class="btn btn-primary addBtn"
          data-action='addtree'
          data-feature="add_plot"
+         data-event-category="add-tree" data-event-action="start-add-tree"
```
RickMohr: Probably we also want `data-event` atts on the top-level menu buttons (just below `instancetopnav`).

opentreemap/treemap/templates/treemap/map-add-tree.html
```diff
@@ -35,22 +35,26 @@ <h4 class="summaryAddress"></h4>
...
+                    <input type="radio" name="addFeatureOptions" value="edit" id="addtree-viewdetails"
+                           data-event-category="add-tree" data-event-action="done-edit" /><label for="addtree-viewdetails">
                         {% trans "Continue editing this tree" %}
                     </label>
                 </div>
                 <div>
-                    <input type="radio" name="addFeatureOptions" value="close" id="addtree-done" checked /><label for="addtree-done">
+                    <input type="radio" name="addFeatureOptions" value="close" id="addtree-done" checked
+                           data-event-category="add-tree" data-event-action="done" /><label for="addtree-done">
                         {% trans "I'm done!" %}
```
RickMohr: These only add events if the user changes "After I add this tree". Seems like it would be better to have an event on the "Done" button.

PR #2591, opened 2016-05-20
Make Boundary.updated_at not nullable

jwalgran:
Removes a TODO from the code. When this field was added it was marked as nullable so that the migration would be backwards compatible. The the current production database does not have any null values in this column.


maurizi:
:+1:

Issue #2594, opened 2016-05-23
If plot.id is set to invisible the tree detail page raises an exception while rendering


jwalgran:
Example exception:

```
ERROR:django.request:Internal Server Error: /INSTANCE/features/2120589/
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app/core/opentreemap/treemap/decorators.py", line 45, in wrapper
    return view_fn(request, instance, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django_tinsel/decorators.py", line 33, in routed
    return req_method(request, *args2, **kwargs2)
  File "/opt/app/core/opentreemap/treemap/views/map_feature.py", line 108, in render_map_feature_detail
    **kwargs)
  File "/opt/app/core/opentreemap/treemap/views/map_feature.py", line 101, in map_feature_detail
    RequestContext(request))
  File "/usr/local/lib/python2.7/dist-packages/django/shortcuts.py", line 45, in render_to_response
    using=using)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader.py", line 115, in render_to_string
    template_name, context, context_instance, dirs, dictionary)
  File "/usr/local/lib/python2.7/dist-packages/django/template/engine.py", line 221, in render_to_string
    return t.render(context_instance)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 210, in render
    return self._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 202, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 905, in render
    bit = self.render_node(node, context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/debug.py", line 79, in render_node
    return node.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/loader_tags.py", line 135, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 202, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python2.7/dist-packages/django/template/base.py", line 905, in render
    bit = self.render_node(node, context)
[...]

PR #2595, opened 2016-05-23
Handle invisible plot ID field

jwalgran:
Instance admins can set the permissions on the Plot.id field to "Invisible." This is not a useful feature, and we intend to replace it with a better permission system.

For now, referencing `pk` rather than `id` prevents template rendering errors and we can fix the problem of a map feature not showing up as favorited on the detail page by performing the favorite check before the fields are masked.

---

##### Testing

On the `develop` branch, on the roles management page for an instance you own, set Planting Site Id to Invisible on the Administrator role. Attempting to view a tree detail page will raise an exception. Switch to this branch and detail pages should load without error and the favoriting feature should work.

---

Connects to https://github.com/OpenTreeMap/otm-core/issues/2594

### DIFF
--- a/opentreemap/treemap/lib/map_feature.py
+++ b/opentreemap/treemap/lib/map_feature.py
@@ -331,11 +331,17 @@ def context_dict_for_map_feature(request, feature, edit=False):
 
     user = request.user
     if user and user.is_authenticated():
-        feature.mask_unauthorized_fields(user)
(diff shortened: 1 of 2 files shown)


hectcastro:
@azavea-bot rebuild

RickMohr:
+1 Good solution until we have better permissions. Works as advertised.

Issue #2589, opened 2016-05-19
Add canopy columns


jwalgran:
Add a `canopy_percent` numeric column to the boundaries table
Add a `canopy_enabled` boolean column to treemap_instance table. False by default.
Add a nullable `canopy_boundary_category ` to treemap_instance table. This is the boundary category that will be used for displaying a choropleth layer

Write and run a script to set these values on the PA instance.

PR #2596, opened 2016-05-24
Make instance canopy fields non-null

jwalgran:
This cleans up the backwards compatible 0026 migration.

---

Connects to https://github.com/OpenTreeMap/otm-core/issues/2589


jwalgran:
This can be merged now that the 2.8 release has been started.

RickMohr:
+1 
Django shell refused to save an instance where `canopy_enabled` or `canopy_boundary_category` was set to `None`. Existing instances have `canopy_enabled == False` and `canopy_boundary_category == ''`.

Issue #2598, opened 2016-05-25
Add field to toggle boundary searching


kdeloach:
We'll need to add a field to the `treemap_boundary` table to prevent some canopy coverage boundaries from appearing as searchable locations on the frontend.

Otherwise, many of the smaller boundaries which are not meaningful in a search context will appear in the locations dropdown.

![screenshot 2016-05-27 at 16 47 54](https://cloud.githubusercontent.com/assets/43062/15620848/d39b997a-242a-11e6-97d9-b6d486995f41.png)

PR #2605, opened 2016-05-27
Add searchable field to Bounds model

kdeloach:
The purpose of this field is to disable boundaries from appearing in the
frontend "Search by Location" autocomplete popup.

**Test:**

1. Run migrations
    ```sh
    ./scripts/manage.sh migrate
    ```

2. Disable searching on all boundaries (if you dare)
    ```sh
    vagrant ssh services -c 'sudo -u postgres psql otm -c "update treemap_boundary set searchable=false"'
    ```

3. Verify that no boundaries appear in the "Search by Location" autocomplete

**Note:** You may have to clear localStorage to see this change.

```
localStorage.clear()
```

Connects #2598

### DIFF
--- a/opentreemap/treemap/models.py
+++ b/opentreemap/treemap/models.py
@@ -1230,6 +1230,7 @@ class Boundary(models.Model):
                                       db_index=True)
 
     canopy_percent = models.FloatField(null=True)
+    searchable = models.NullBooleanField(default=True)
 
     objects = models.GeoManager()
 
--- a/opentreemap/treemap/migrations/0027_boundary_searchable.py
+++ b/opentreemap/treemap/migrations/0027_boundary_searchable.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('treemap', '0026_add_canopy_fields'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='boundary',
+            name='searchable',
+            field=models.NullBooleanField(default=True),
+        ),
+    ]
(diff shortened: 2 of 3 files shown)


opentreemap/treemap/models.py
```diff
@@ -1230,6 +1230,7 @@ class Boundary(models.Model):
                                       db_index=True)
 
     canopy_percent = models.FloatField(null=True)
+    searchable = models.BooleanField(default=True)
```
jwalgran: We like to keep our migrations backwards compatible by using `NullBooleanField` and then making it non-nullable in a follow up migration.

kdeloach: This should populate existing rows with a default value for backwards compatibility. Do you prefer making this a `NullBooleanField` instead?

jwalgran: Sorry. Backwards compatible from a blue/green standpoint. During a deploy we run the migrations and test the "dark" stack. During that time, the old code is running with the new schema and inserts would fail.

It is a little academic in this situation, because there is no user-facing way to insert into the boundaries table. I have been bitten by migration problems during deploys enough times that I like to err on the side of safety and keep all new fields nullable when they are first introduced.

kdeloach: Understood.

kdeloach:
Changed `BooleanField` to `NullBooleanField` and updated the autocomplete endpoint to exclude boundaries which have `searchable` set to `False`.

opentreemap/treemap/models.py
```diff
@@ -1230,6 +1230,7 @@ class Boundary(models.Model):
                                       db_index=True)
 
     canopy_percent = models.FloatField(null=True)
+    searchable = models.NullBooleanField()
```
jwalgran: I think we still want a `default=True` here so that the migration will fill the column with `'t'`s for us.

jwalgran:
I used

```sql
update treemap_boundary set updated_at = now() where id = (select max(id) from treemap_boundary);
```

to "touch" the table in development and verified that the boundaries were no in the autocomplete list.

opentreemap/treemap/models.py
kdeloach: Updated.

jwalgran:
👍

PR #2606, opened 2016-05-31
Make boundary searchable field non-null

kdeloach:
This should be included in the next release after #2605 is released to production.


jwalgran:
I removed the "blocked" label because the 2.8.0 release, which includes #2605, is is progress

kdeloach:
This is good to merge, right?

jwalgran:
👍

Issue #2590, opened 2016-05-19
Show canopy percentage at point


jwalgran:
NOTE: Reuse the click handling for green infrastructure

- Add function to retrieve boundary names, categories, and canopy percent at a specified lat/lng.

- Include boundary names, categories, and canopy percentage in the tree detail popup.

- AJAX endpoint returns boundary names, categories, and canopy percentages at a specified lat/lng.

- When clicking the map away from a tree point, show the boundary names, categories, and canopy percentage

RickMohr:
Once this is implemented we'll start seeing canopy info on any treemap which happens to intersect a boundary with canopy data. For example, if somebody made a Pittsburgh tree map they would see the canopy percentages but not the canopy layer. Is that OK? 

If not, we'll want a configuration entry to explicitly enable showing canopy data.

jwalgran:
https://github.com/OpenTreeMap/otm-core/pull/2593 added a `canopy_enabled` field to `Instance` so we can control who sees the canopy features.

PR #2607, opened 2016-05-31
Show canopy information when clicking canopy-enabled maps

RickMohr:
For canopy-enabled instances:
* If click hits a feature, include canopy coverage in popup
* If click doesn't hit a feature, fetch HTML showing canopy coverage
* Display canopy coverage as a table of boundaries at click point and their canopy percentages

Also:
* Increase `maxWidth` of popup
* Remove option to pass size of popup (we aren't using it)

Note that double-clicking to zoom a canopy-enabled map shows a canopy popup you probably didn't want to see.
The easy fix is to use `BU.leafletSingleClickStream` instead of `BU.leafletEventStream` to handle
clicks. But the downside is that normal popups are delayed by e.g. 200ms, which seems noticable.
There's probably a way to incur the delay only for clicks that don't hit a feature, but I didn't see
a simple way to implement that.

Connects #2590


jwalgran:
Our script to load canopy percentages used 0.0 to 1.0. We either need multiply by 100 as part of the formatting in the popup, or update the data loading scripts to use values from 0 to 100.

screen shot 2016-06-01 at 11 18 05 am

I lean toward storing them in a more "raw" form as 0.0 to 1.0, since they can be used in calculations without first dividing them by 100.

jwalgran:
I agree with not using `leafletSingleClickStream` to avoid the delay and addressing the popup being shown on double-click if it is reported as a problem.

Nice work and nice additional cleanup.

👍

Issue #2588, opened 2016-05-19
After searching by address show a dot on the map


jwalgran:
- The dot should be clearly distinguishable from a tree or stormwater feature.
- The dot should not obscure a map feature and block click events.
- The dot should be removed when searching for a different location

Related to [Address search zooms in but does not highlight tree](https://github.com/OpenTreeMap/otm-core/issues/2564) but different.

jmorrison1847:
+1 Greening of Detroit

jmorrison1847:
+1 Wingfield Springs

PR #2611, opened 2016-06-06
Show a map marker when a location search succeeds

RickMohr:
* Marker is removed when user starts a new search or clicks the "Reset" button.
* Icon uses otherwise-unused `marker.png`.
* Marker displays on top of other layers, but is not clickable.

Connects #2588

image


jwalgran:
Works nicely.

I would like to ask Scott/Matt to develop an original position marker icon.

👍

Issue #2609, opened 2016-06-02
Add a thematic canopy layer to the map, with double-ended slider to control it


jwalgran:
Depends on https://github.com/OpenTreeMap/otm-tiler/issues/108

- When canopy is enabled, make choropleth a selectable layer.
- Add a double ended slider to set the minimum and maximum canopy percentage arguments. Refresh the canopy layer when they change.

PR #2613, opened 2016-06-08
Add canopy percent choropleth layer

kdeloach:
**Test:**
* Checkout https://github.com/OpenTreeMap/otm-tiler/pull/109 and restart the tiler
* Rebuild static assets `./scripts/grunt.sh`
* Load "PA Tree Map" (or any instance with canopy data)
  * Set the canopy category `update treemap_instance set canopy_boundary_category ='Municipality' where url_name='patreemap';`
* Ensure that the "Regional Canopy Percentages" layer exists
* Ensure that the filter control appears when "Regional Canopy Percentages" layer is visible
* Ensure that altering the filter control updates the canopy layer
 
screenshot 2016-06-08 at 17 03 08

Connects #2609

Depends on https://github.com/OpenTreeMap/otm-tiler/pull/109

### DIFF
--- a/opentreemap/treemap/js/src/CanopyFilterControl.js
+++ b/opentreemap/treemap/js/src/CanopyFilterControl.js
@@ -1,0 +1,54 @@
+"use strict";
+
+var $ = require('jquery'),
+    _ = require('lodash'),
+    L = require('leaflet'),
+    Bacon = require('baconjs');
+
+require('ionRangeSlider');
+
+var canopyFilterTmpl = _.template($('#canopy-filter-tmpl').html());
+
+var CanopyFilterControl = L.Control.extend({
+    options: {
+        position: 'bottomleft'
+    },
+
+    initialize: function() {
+        this._changeBus = new Bacon.Bus();
+        this.tilerArgsProp = this._changeBus.toProperty({
+            canopyMin: 0,
+            canopyMax: 1
+        });
+    },
+
+    onAdd: function(map) {
+        var self = this,
+            $el = $(canopyFilterTmpl());
+
+        $el.find('[name="canopy-slider"]').ionRangeSlider({
+            type: 'double',
+            min: 0,
+            max: 100,
+            from: 0,
+            to: 100,
+            step: 1,
+            drag_interval: true,
+            hide_min_max: true,
+            grid: false,
+            postfix: '%',
+            onChange: function (data) {
+                self._changeBus.push({
+                    canopyMin: data.from / 100,
+                    canopyMax: data.to / 100
+                });
+            },
+        });
+
+        var el = $el.get(0);
+        L.DomEvent.disableClickPropagation(el);
+        return el;
+    }
+});
+
+module.exports = CanopyFilterControl;
--- a/opentreemap/treemap/js/src/MapManager.js
+++ b/opentreemap/treemap/js/src/MapManager.js
@@ -11,6 +11,7 @@ var $ = require('jquery'),
     urlState = require('treemap/urlState'),
 
     layersLib = require('treemap/layers'),
+    CanopyFilterControl = require('treemap/CanopyFilterControl'),
 
     MIN_ZOOM_OPTION = layersLib.MIN_ZOOM_OPTION,
     MAX_ZOOM_OPTION = layersLib.MAX_ZOOM_OPTION,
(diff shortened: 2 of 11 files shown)


kdeloach:
Refactored this to use a slightly more elegant solution to filter the canopy layer.

kdeloach:
Added test instructions.

jwalgran:
@designmatty You helped us with choosing a color for tree canopy tiles and I would like to get your opinion on the color ramp for this choropleth map. 

The ramp is defined here: https://github.com/OpenTreeMap/otm-tiler/pull/109/files#diff-7d93bf1c5b9fa1d6ea25a6137f2047efR5

opentreemap/treemap/js/src/CanopyFilterControl.js
```diff
@@ -0,0 +1,53 @@
...
+            min: 0,
+            max: 1,
+            from: 0,
+            to: 1,
+            step: 0.01,
+            drag_interval: true,
+            hide_min_max: true,
+            grid: false,
+            prettify: function(n) {
+                return Math.floor(n * 100) + '%';
```
jwalgran: The slider jumps over certain values, but the values that are not selectable are not consistent. This may be frustrating for a user who is trying to set the sliders to "20% - 40%" but can only see "19% - 41%," for instance.

We may need to reverse the slider behavior so that the min and max are 0 to 100 and then divide by 100 in `onChange`.

designmatty:
@jwalgran looks pretty good to me right now.

opentreemap/treemap/js/src/CanopyFilterControl.js
kdeloach: I wasn't able to reproduce this but making the slider intervals whole numbers does seem like a better approach. Let me know if this problem still happens.

opentreemap/treemap/js/src/CanopyFilterControl.js
```diff
@@ -0,0 +1,51 @@
...
+        position: 'bottomleft'
+    },
+
+    initialize: function() {
+        this._changeBus = new Bacon.Bus();
+        this.changeStream = this._changeBus.toEventStream();
+    },
+
+    onAdd: function(map) {
+        var self = this,
```
kdeloach: TODO: `self` is no longer required.

opentreemap/treemap/js/src/MapManager.js
```diff
@@ -104,7 +105,32 @@ MapManager.prototype = {
 
         if (config.instance.canopyEnabled) {
             var canopyLayer = layersLib.createCanopyBoundariesTileLayer(config);
-            map.addLayer(canopyLayer);
+
+            var filterControl = new CanopyFilterControl();
+            filterControl.changeStream.debounce(1000).onValue(function(data) {
+                _.extend(config.instance, {
```
kdeloach: TODO: XXX

maurizi:
Most of our third-party library files use the unminified version, since we minify our entire JS bundle as part of our build process, and using the unminified version allows for easier debugging in development.

Can we change out `ion.rangeSlider.min.js` for the unminified file?

kdeloach:
Added the unminified version of the range slider plugin.

kdeloach:
Found a slightly more elegant solution to defining canopy tile layer arguments. By replacing the Bacon stream with a property, I was able to isolate all the canopy related tiler arguments within the filter control.

opentreemap/treemap/js/src/CanopyFilterControl.js
```diff
@@ -0,0 +1,56 @@
...
+
+var CanopyFilterControl = L.Control.extend({
+    options: {
+        position: 'bottomleft'
+    },
+
+    initialize: function() {
+        this._changeBus = new Bacon.Bus();
+        this.tilerArgsProp = this._changeBus.toProperty({
+            category: 'Municipality',
```
jwalgran: Instead of hard coding a value, the `category` should come from the instance config. The canopy control behavior is not affected by the value of `category`, so we may want to separate our concerns and have the stream consumer append it.

```javascript
filterControl.tilerArgsProp
             .debounce(1000)
             .map(function(args) {
               args.category = config.category;
               return args;
             })
             .onValue(...)

Comment (kdeloach) [inline opentreemap/treemap/js/src/CanopyFilterControl.js]: My thoughts are that we could reasonably extend the canopy filter to allow users to select which category to display. In that case, I think it's best to leave the code as-is. On the other hand, if we will only ever have the option to display a single category per instance than your solution makes more sense. 

Another alternative is to dynamically adjust which features are displayed based on zoom level. In which case, neither of these solutions matter. I decided that it might make more sense to just have some hard-coded assumptions for now, and wait to see how the requirements for this feature evolve.

If you have a strong preference for pulling this value out into the instance settings, I can do that. But it will still be hard-coded, just at a slightly higher level.

Comment (jwalgran) [inline opentreemap/treemap/js/src/CanopyFilterControl.js]: If we opt to hard code a value, we will need to remove this configuration field from the model, to avoid confusion.

```python
    """
    The boundary category to be used for showing a choropleth canopy
    layer. max_length=255 matches Boundary.category
    """
    canopy_boundary_category = models.CharField(max_length=255, default='',
                                                blank=True)
```

https://github.com/OpenTreeMap/otm-core/blob/9f49ddf638a0ce8ecac0ace2a02f5a0b641edd71/opentreemap/treemap/instance.py#L225-L230

I am ok with removing it.

kdeloach: I wasn't aware that column existed. In that case, I'll update the code based on your suggestions.

kdeloach: Thanks for showing me this.

kdeloach:
Updated to remove hard-coded canopy boundary category.

jwalgran:
👍

Issue #2525, opened 2016-03-01
Show a message when a photo upload fails with a 413 request entity too large response


jwalgran:
Nginx blocks requests larger than a certain size, returning a 413 response.

PR #2614, opened 2016-06-09
Show error message when uploading files larger than 1MB

kdeloach:
Prevent image uploads for file sizes equal to or greater than 1MB.

**Test:**
* Rebuild static assets `./scripts/grunt.sh`
* Try to upload a file 1MB or larger
* Ensure you get an error toast if the file is too big

Connects #2525

### DIFF
--- a/opentreemap/treemap/js/src/imageUploadPanel.js
+++ b/opentreemap/treemap/js/src/imageUploadPanel.js
@@ -2,6 +2,8 @@
 
 "use strict";
 
+var toastr = require('toastr');
+
 // For modal dialog on jquery
 require('bootstrap');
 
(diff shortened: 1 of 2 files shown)


kdeloach:
Updated to address comments.

opentreemap/treemap/js/src/imageUploadPanel.js
```diff
@@ -93,6 +101,20 @@ module.exports.init = function(options) {
         }
     });
 
+    fileupload.bind('fileuploadadd', function(e, data) {
+        data.process(function() {
+            var defer = $.Deferred();
+            _.each(data.files, function(file) {
+                if (file.size >= MAX_FILE_SIZE_BYTES) {
+                    toastr.error(file.name + ' exceeds the maximum file size of 1MB');
```
jwalgran: We should pass this error text in through the init options so that we can translate it.

I would also like to keep the human readable version of the max file in a constant right below the numeric version, so they can easily be changed together if we decide to increase the file size limit.

kdeloach: Done.

RickMohr:
+1 correctly rejects big files and permits small files.

jwalgran:
@azavea-bot rebuild

PR #2618, opened 2016-06-09
Ensure checkbox value is always True or False

kdeloach:
This reverts a change that was made to set a custom display value in the
super admin. The super admin addon has been updated to use data
attributes to set a custom display value instead of hard-coding them
here.

The value field for checkboxes must always be True or False for parity
with the Django backend. Otherwise, checkbox values will become corrupt after
saving the form multiple times via ajax.

Ref https://github.com/OpenTreeMap/otm-addons/issues/1173


RickMohr:
+1

Issue #2437, opened 2016-01-13
After quick editing a tree advanced search parameters are no longer active when new tiles are requested


jwalgran:
1. Add two plots to the map with no trees
1. Advanced search for "Show Missing Tree diameter." Both plot dots should stay full green.
1. Select one plot, quick edit, add a diameter, save

Expected: Tiles reload and only one of the dots is full green
Actual: Both dots are still full green. If you refresh the page, the advanced search is applied and only one dot is full green.

PR #2623, opened 2016-06-13
Reapply search filters after quick edit

kdeloach:
Instead of setting the layer URL to `noSearchUrl` after doing a
quick edit, we now update the rev hash on the layer URL to request new
tiles, while also preserving the current advanced search filters.

Test instructions are in #2437.

Connects #2437


maurizi:
:+1:

Issue #2608, opened 2016-06-02
List boundaries in alphabetical order for locations dropdown


kdeloach:
![screenshot 2016-06-02 at 14 27 42](https://cloud.githubusercontent.com/assets/43062/15756197/316a18d6-28ce-11e6-9135-0f0f3ababab6.png)

PR #2625, opened 2016-06-14
Fix boundaries dropdown sort order

kdeloach:
* Update boundaries endpoint to order results by sort_order then name
* Ensure client-side results are sorted when displaying "all" results
  (Bloodhound orders "all" results by id)

Connects #2608

### DIFF
--- a/opentreemap/treemap/js/src/otmTypeahead.js
+++ b/opentreemap/treemap/js/src/otmTypeahead.js
@@ -64,7 +64,7 @@ var create = exports.create = function(options) {
         $hidden_input = $(options.hidden),
         $openButton = $(options.button),
         reverse = options.reverse,
-        sorter = _.isArray(options.sortKeys) ? getSortFunction(options.sortKeys) : undefined,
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/js/src/otmTypeahead.js
```diff
@@ -94,7 +94,9 @@ var create = exports.create = function(options) {
             limit: 3000,
             source: function(query, sync, async) {
                 if (query === '') {
-                    sync(prefetchEngine.all());
+                    var result = prefetchEngine.all();
+                    result.sort(sorter);
+                    sync(result);
```
maurizi: Since we now require `sorter`, we need a reasonable default if `sortKeys` isn't passed in.

I think changing
`sorter = _.isArray(options.sortKeys) ? getSortFunction(options.sortKeys) : undefined,`
to
`sorter = _.isArray(options.sortKeys) ? getSortFunction(options.sortKeys) : getSortFunction(['value']),`

would work well.

kdeloach: Good point, updated.

maurizi:
:+1:

PR #2628, opened 2016-06-16
Increase max upload filesize

kdeloach:
Read `MAXIMUM_IMAGE_SIZE` from an environmental variable and pass that
value to the client-side JS instead of hard-coding it.

Depends on https://github.com/OpenTreeMap/otm-cloud/pull/233

### DIFF
--- a/opentreemap/opentreemap/settings/default_settings.py
+++ b/opentreemap/opentreemap/settings/default_settings.py
@@ -45,6 +45,7 @@ AWS_HEADERS = {
 
 # Size is in bytes (20 MB)
 MAXIMUM_IMAGE_SIZE = int(os.environ.get('DJANGO_MAXIMUM_IMAGE_SIZE', 20971520))
+MAXIMUM_IMAGE_SIZE_MB = MAXIMUM_IMAGE_SIZE / 1024 / 1024
 
 # API distance check, in meters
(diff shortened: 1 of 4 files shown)


opentreemap/opentreemap/settings/default_settings.py
```diff
@@ -44,7 +44,8 @@
 }
 
 # Size is in bytes
-MAXIMUM_IMAGE_SIZE = 20971520  # 20mb
+MAXIMUM_IMAGE_SIZE = int(os.environ.get('DJANGO_MAXIMUM_IMAGE_SIZE', 20971520)) # 20 MB
```
jwalgran: Lint:

Lint 

```
opentreemap/opentreemap/settings/default_settings.py:47:80: E261 at least two spaces before inline comment
opentreemap/opentreemap/settings/default_settings.py:47:80: E501 line too long (87 > 79 characters)

kdeloach:
This is ready to be reviewed.

jwalgran:
To test I used images just under 20MB and just over 20MB. I saw the correct behaviour uploading photos for both trees and map features.

I commented out the client-side size validation and verified that Nginx correctly blocked an upload over 20MB and allowed an upload under 20MB

👍

PR #2631, opened 2016-06-20
Backfill canopy_boundary_category column before making non null

kdeloach:
Fixes an error received applying migrations to staging.

Ref: OpenTreeMap/otm-cloud#234


jwalgran:
What is the best way to recreate the problem in my dev database?

kdeloach:
For reasons not entirely clear to me, I was not able to run migrations against the production database backup without these changes.

kdeloach:
I could only reproduce this error by running migrations against a production database backup. It's possible that restoring from a backup causes this error, and that this is not necessary at all. It still might be a good idea to include this just to be on the safe side.

kdeloach:
I also found that doing a reverse migration to drop the canopy columns and then re-running migrations also fixes this problem... Which is another indication that the backup I restored from may have been dumped in a weird state.

jwalgran:
Migrations applied cleanly. After this migration, my `canopy_boundary_category` does not have nulls.

```
select url_name, canopy_boundary_category from treemap_instance where canopy_boundary_category is null;
 url_name | canopy_boundary_category
----------+--------------------------
```
👍

PR #2633, opened 2016-06-22
Initialize eco factors from instance bounds, not point

RickMohr:
Sometimes the center of a treemap is not in any i-Tree region, but the treemaps bounds do overlap an i-Tree region.

Depends on OpenTreeMap/otm-addons#1194


kdeloach:
+1 Tested

Issue #2629, opened 2016-06-16
Add color legend to Regional Canopy Filter slider


jwalgran:
Show that light green means less canopy, dark green means more canopy.

This could be implemented as a gradient inside the slider bar.

PR #2637, opened 2016-06-27
Add regional canopy filter legend

kdeloach:
This alters the style of the regional canopy filter control somewhat and adds a legend.

screenshot 2016-06-27 at 16 20 16

Connects #2629


RickMohr:
Looks good. I would remove "Less" and "More" for two reasons:
1. The gradient values would line up with the slider values
2. I think the concept is clear without the words

kdeloach:
Very true. Removed.

RickMohr:
+1

PR #2639, opened 2016-06-29
Improve default map extent to be based on feature locations

maurizi:
We previously always centered the map on the map center point with a fixed
zoom.
With this change, we instead position the map so that all features added
to the map are visible. If a map has no trees yet, we position the map so
that all of the map's bounds are visible.

Connects to OpenTreeMap/otm-addons#1155

### DIFF
--- a/opentreemap/opentreemap/util.py
+++ b/opentreemap/opentreemap/util.py
@@ -119,3 +119,15 @@ def add_rollbar_handler(logger, level=logging.WARNING):
         rollbar_handler = RollbarHandler()
         rollbar_handler.setLevel(level)
         logger.addHandler(rollbar_handler)
+
+
+def extent_as_json(extent):
+    xmin, ymin, xmax, ymax = extent
+
+    return json.dumps({'xmin': xmin, 'ymin': ymin, 'xmax': xmax, 'ymax': ymax})
+
+
+def extent_intersection(*extents):
+    extents = zip(*extents)
+    xmins, ymins, xmaxes, ymaxes = extents
+    return (max(xmins), max(ymins), min(xmaxes), min(ymaxes))
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.contrib.gis.db import models
+from django.contrib.gis.db.models import Extent
 from django.contrib.gis.geos import MultiPolygon, Polygon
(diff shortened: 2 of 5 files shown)


kdeloach:
The feature works perfectly. 

However, some instances now have worse initial extents in cases where they have added trees that far exceed their bounds. We may want to preview this change on several instances from production before merging this.

**phillytreemap**

Before:
screenshot 2016-06-29 at 15 39 31

After:
screenshot 2016-06-29 at 15 41 31

**latreemap**

Before:
screenshot 2016-06-29 at 15 40 26

After:
screenshot 2016-06-29 at 15 40 49

maurizi:
@kdeloach in general its not supposed to be possible to have trees that are outside of the map bounds.

For PhillyTreeMap it goes out that much further because the map boundaries actually do go out that far.  PhillyTreeMap includes all of Delaware, and there are some trees in south Delaware.

For LA, it looks like there is some bad data, as there is a tree in Taiwan :astonished: .

We may want a data migration that deletes trees that are outside their map's boundaries.

@RickMohr What do you think?

kdeloach:
I may have picked the 2 weirdest instances to test.

RickMohr:
I kind of like the data migration. `phillytreemap` and `latreemap` would start zoomed further out.

But a safer option (to avoid surprising our users) would be to say if the zoom level is lower than our current default zoom of 11, do what we do now (zoom 11, centered on map center). We'd get the benefits of this PR but `phillytreemap` and `latreemap` would look the same as before.

maurizi:
@RickMohr I don't think that's a good option, because part of the point of this was for statewide maps with custom boundaries to be more zoomed out.

RickMohr:
Good point.

Better idea -- clip the extent of the features to the extent of the bounds.

kdeloach:
We may want to consider caching this potentially expensive operation instead of calculating it for each page request.

RickMohr:
```
i = Instance.objects.get(url_name='latreemap')
time(i.mapfeature_set.aggregate(Extent('geom'))['geom__extent'])
CPU times: user 42 µs, sys: 4.26 ms, total: 4.3 ms
Wall time: 352 ms
```

RickMohr:
Rather than putting it in `settings.js` (computing it on every page request) we could make it a data attribute on the map element (computing it only for the map page).

maurizi:
@kdeloach @RickMohr This has been updated significantly, please take another look.

RickMohr:
LGTM

kdeloach:
+1 Looks good.

PR #2641, opened 2016-06-30
Add support for Google Maps keys provided as environment variables

maurizi:
Depends on https://github.com/OpenTreeMap/otm-cloud/pull/238

Connects to OpenTreeMap/otm-cloud#220


opentreemap/treemap/templates/treemap/map.html
```diff
@@ -7,7 +7,7 @@
 {% endblock head_extra %}
 
 {% block searchscripts %}
-<script src="//maps.google.com/maps/api/js?v=3&sensor=false"></script>
+<script src="https://maps.googleapis.com/maps/api/js?key={{ settings.GOOGLE_MAPS_API_KEY }}"></script>
```
RickMohr: We load the Google Maps API in two other places:
* https://github.com/OpenTreeMap/otm-core/blob/79701aa167f060726d6a6c665e8de86d7791e27c/opentreemap/treemap/templates/treemap/map_feature_detail.html#L264
* https://github.com/OpenTreeMap/otm-addons/blob/ae686022c0027557403f4e590753dff970cb1edc/modeling/templates/modeling/modeling.html#L98

maurizi: Fixed this, in a rebased commit for this PR (5928556) and with a new PR https://github.com/OpenTreeMap/otm-addons/pull/1202

RickMohr:
+1

PR #2643, opened 2016-07-06
Switch from Grunt+Browserify to Webpack

maurizi:
Work-in-progress branch for switching from Grunt and Browserify to Webpack for our front-end asset toolchain.

What has been done:
 - Files were moved around.  Shared assets are now in `assets/` instead of `opentreemap/treemap/`, and the `lib/` directory was renamed to `vendor/`.
 - JS modules were rewritten.
   - Non-entry modules were moved from `<app>/js/src/` to `<app>/js/src/lib/` and are aliased as `<app>/lib/<filename>.js`.
   - Entry modules were rewritten to remove their exported `init()` functions.  Instead, the modules are executed when included in the page. I made the templated `settings.js` file accessible as a module, and also brought in the `django-js-reverse` library to make this easier.
   - On some pages, options are still being passed from the template to the JS module, but this is now done via a JS global variable.
 - CSS files were rewritten to reference static assets from a `$staticUrl` variable, which defaults to `../../` for webpack, but is changed to `/static/` for map-branded CSS builds.
 - `webpack-dev-server` support was added.  If the `WEBPACK_DEV_SERVER` environment variable is present and `settings.DEBUG` is false, all requests for static assets will go through `webpack-dev-server`.
 - The webpack configuration files are present and should work for one-time builds for both development and production

What remains to be done:
- [x] JS tests are totally broken.
- [x] Travis CI and Circle CI files need to be updated
- [x] Performance could be better.  ~~I suspect a vendor JS bundle may help.~~ After switching source maps to use `eval` incremental rebuilds are very speedy
- [x] Documentation needs updating.
- [x] Source map support and hot reload support is a bit spotty and needs further investigation.
- [x] Fix UI test failures.
- [x] Integration with Django to support long term caching.

Depends on https://github.com/OpenTreeMap/otm-cloud/pull/241
Depends on https://github.com/OpenTreeMap/otm-addons/pull/1204

Connects to OpenTreeMap/otm-addons#1198


maurizi:
@azavea-bot rebuild

opentreemap/treemap/templates/treemap/map_feature_detail.html
```diff
@@ -267,58 +267,19 @@ <h4 class="address" id="map-feature-address">{{ feature.address_full }}</h4>
 {% localize off %}
 // Define options for mapFeature.init(), which will be called in the
 // type-specific template inheriting from this one (e.g. plot_detail.html)
-var mapFeatureOptions = {
-    config: otm.settings,
+window.mapFeature = {
```
jwalgran: We should consider "namespacing" all of our globals under `window.otm` so segregate them from the hundreds of globals already under `window`.

It may also be a good idea to explain, in a comment, why we are using globals and that we are only using them for static data.

opentreemap/treemap/js/src/lib/config.js
```diff
@@ -0,0 +1,3 @@
+"use strict";
+
+module.exports = Object.freeze(window.otm.settings);
```
jwalgran: I like the idea of containing the global access behind this module.

opentreemap/opentreemap/settings/default_settings.py
```diff
@@ -192,10 +192,12 @@
...
 # Additional locations of static files
 STATICFILES_DIRS = (
+    # Webpack compiled files are output here
     os.path.join(PROJECT_ROOT, 'static/'),
-    # Put strings here, like "/home/html/static" or "C:/www/django/static".
-    # Always use forward slashes, even on Windows.
-    # Don't forget to use absolute paths, not relative paths.
+    # This is the directory where webpack get's its source data from
+    # We use this as a static directory so that images can be referenced in CSS
+    # but also be collected as by collectstatic
```
jwalgran: Comment typo: `collected as by collectstatic`

webpack.common.config.js
```diff
@@ -3,7 +3,9 @@
 var Webpack = require('webpack'),
     glob = require('glob'),
     path = require('path'),
-    _ = require('lodash');
+    _ = require('lodash'),
+    _ = require('lodash'),
```
jwalgran: Duplicated require line.

webpack.common.config.js
```diff
@@ -55,16 +57,20 @@ module.exports = {
         path: d('static'),
         sourceMapFilename: '[file].map'
     },
+    devtool: "source-map",
     module: {
         loaders: [{
             include: [shimmed["bootstrap-datepicker"], shimmed["bootstrap-multiselect"], shimmed["bootstrap-slider"]],
             loader: "imports?bootstrap"
+        }, , {
```
jwalgran: Looks like there is an extra comma here.

webpack.test.config.js
```diff
@@ -0,0 +1,18 @@
...
+var webpack = require('webpack'),
+    config = require('./webpack.common.config.js');
+
+// Only use the test entry file
+config.entry = {
+    'js/testRunner': __dirname + '/assets/tests/testRunner.js'
+};
+// Put a dummy value in place of the django reverse module in test mode
+config.externals = {
+    reverse: "undefined"
```
jwalgran: Is it important that this is the string `"undefined"` and not the literal `undefined`?

webpack.test.config.js
```diff
@@ -0,0 +1,18 @@
...
+};
+
+// We only want the ProvidePlugin for tests
+config.plugins = [config.plugins[0]];
```
jwalgran: Are we guaranteed that the `ProvidePlugin` will always be at index 0?

webpack.test.config.js
maurizi: Yes. The values in the `externals` hash are the _names_ of the global variables that are returned when requesting that module.  So in this case, when a module requires `reverse` in the test bundle, it will get the global variable _named_ `undefined`.

webpack.test.config.js
maurizi: The order is determined by what we put in the plugins array in `webpack.common.config.js`.  I'll throw a comment in there that the test config depends on `PluginProvider` being first,

opentreemap/opentreemap/settings/default_settings.py
```diff
@@ -177,16 +180,24 @@
...
 SITE_ROOT = '/'
 
 # Additional locations of static files
 STATICFILES_DIRS = (
-    # Put strings here, like "/home/html/static" or "C:/www/django/static".
-    # Always use forward slashes, even on Windows.
-    # Don't forget to use absolute paths, not relative paths.
+    # Webpack compiled files are output here
+    os.path.join(PROJECT_ROOT, 'static/'),
+    # This is the directory where webpack get's its source data from
```
RickMohr: get's -> gets

opentreemap/treemap/js/src/base.js
```diff
@@ -0,0 +1,6 @@
+"use strict";
```
RickMohr: Consider a comment on this file to explain what it's for. Does this define what's in the common bundle?

RickMohr:
Nice move to put `config` in a module. Hallelujah not having to pass it around everywhere!

RickMohr:
Since so much code changed we'll want to be extra careful when testing for the next release...

opentreemap/treemap/js/src/base.js
maurizi: Comment added in c3edfae.

opentreemap/opentreemap/settings/default_settings.py
maurizi: Fixed in c3edfae

opentreemap/treemap/js/src/lib/mapFeatureDelete.js
```diff
@@ -20,7 +27,7 @@ exports.init = function(options) {
 
         getUrls = options.getUrls || function () {
```
RickMohr: `options.getUrls` is not defined

maurizi:
@azavea-bot rebuild

opentreemap/treemap/js/src/lib/mapFeatureDelete.js
maurizi: Fixed in 07cd27d

RickMohr:
+1

PR #2644, opened 2016-07-07
Loosen the selenium package version spec

jwalgran:
On several occasions we have run into unit test failures that were resolved by updating the selenium package. Because this package is only installed in dev and test environments loosening the version specification will not jeopardize the repeatability of production builds.

2.53.6 was the latest version at the time of this commit.


maurizi:
:+1: I've wanted to do this for a while now.

jwalgran:
412a90d is a rebase on top of the latest `develop`

Issue #1003, opened 2014-01-08
Users should be able to delete map features they entered


dboyer:
Even if they don't have the ability to edit all fields associated with that tree.

jmorrison1847:
+1 LA Tree Map, MyTreeBR

PR #2646, opened 2016-07-08
Let users delete trees and map features they created, regardless of permissions

RickMohr:
* Declare on models `MapFeature` and `Tree` that users should be able to delete objects they create (via `users_can_delete_own_creations = True`)
* Update `Authorizable.user_can_delete()`, which checks whether a user is allowed to delete an object:
  * If user lacks delete permission and model has `users_can_delete_own_creations == True`, check Audit table to see if user created object.
  * Change administrator test to check `admin` field rather than role name

Related updates in the 3 places we check delete permissions:

1. Delete from a mobile app (`remove_plot` in `api.views`) -- relies on `user_can_delete()` to raise an exception if user not allowed to delete
   * Changed to catch `AuthorizeException` instead of `Exception`, and use its message instead of making a new one

2. Click the "Delete" button on the map feature detail page (`delete_map_feature` in `treemap.views.map_feature`)
   * Changed to work like deleting from mobile app -- remove "deletable" check and rely on `AuthorizeException`.

3. Enable/disable "Delete" button on map feature detail page
   * Changed to use `is_deletable` filter, which calls `user_can_delete()` (the predicate used when actually deleting).
   * Remove now-unused `map_feature_is_deletable` filter.

Connects #1003

Testing:

You will need two users with role "Editor" and one user with role "Administrator".
* As the admin user, make tree height and plot length "Read Only" for both Editors and Administrators.
* As the first editor user, create two trees and an empty planting site.
* As the first editor user, delete the first tree and its planting site (allowed because you created them).
* As the second editor user, you shouldn't be able to delete the second tree or the empty planting site (because you don't have permission to write all fields and you didn't create them).
* As the admin user, you should be able to delete the second tree and the empty planting site (because admins can always delete).

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -748,7 +748,20 @@ class Authorizable(UserTrackable):
             return True
         else:
             writeable_perms = self._get_writeable_perms_set(user)
-            return writeable_perms >= set(self.tracked_fields)
+            can_delete = writeable_perms >= set(self.tracked_fields)
+            if not can_delete:
+                if getattr(self, 'users_can_delete_own_creations', False):
+                    can_delete = self._was_created_by(user)
+            return can_delete
+
+    def _was_created_by(self, user):
+        fields_created_by_user = Audit.objects.filter(
+            model=self.__class__.__name__,
+            model_id=self.id,
+            action=Audit.Type.Insert,
+            user=user,
+        )
+        return fields_created_by_user.exists()
 
     def user_can_create(self, user, direct_only=False):
         """
@@ -871,8 +884,8 @@ class Authorizable(UserTrackable):
             super(Authorizable, self).delete_with_user(user, *args, **kwargs)
         else:
             raise AuthorizeException("%s does not have permission to "
-                                     "delete %s objects." %
-                                     (user, self._model_name))
+                                     "delete %s %s." %
+                                     (user, self._model_name, self.id))
 
 
 class AuditException(Exception):
--- a/opentreemap/api/views.py
+++ b/opentreemap/api/views.py
@@ -32,7 +32,8 @@ from treemap.decorators import api_admin_instance_request as \
     admin_instance_request
 from treemap.decorators import creates_instance_user
 from django_tinsel.exceptions import HttpBadRequestException
-from treemap.audit import Audit, approve_or_reject_audit_and_apply
+from treemap.audit import (Audit, approve_or_reject_audit_and_apply,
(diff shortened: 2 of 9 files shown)


jwalgran:
I am assigning this to myself so that we can get this merged while Mike is working on another issue.

opentreemap/treemap/audit.py
```diff
@@ -743,12 +743,25 @@ def user_can_delete(self, user):
         field permissions on a model.
         """
 
-        is_admin = user.get_role(self.instance).name == Role.ADMINISTRATOR
+        is_admin = user.get_instance_user(self.instance).admin
```
jwalgran: It has never been great that we match in the name of the Administrator role, but the `admin` flag on `InstanceUser` is used to control access to the management pages. If we couple these together, we cannot create a user that can delete map feature data without also letting them delete custom fields, etc. This may be ok, but there should be some discussion about this change if there hasn't been already.

jwalgran:
What is the purpose of this testing step?

>As the admin user, make tree height and plot length "Read Only" for both Editors and Administrators.

RickMohr:
Making tree height and plot length "Read Only" for Editors and Administrators should mean they can't be deleted by an Editor who didn't create them. An Administrator who didn't create them should still be able to delete them because administrators can always delete.

opentreemap/treemap/audit.py
RickMohr: My mistake. This should be reverted. Better to always allow Administrators to delete than to always allow people with management access to delete.

jwalgran:
There is an edge case when adding a tree to a planting site created by another user:

- As admin user
  - Create an empty planting site
- As editor user
  - Open the empty planting site detail page
  - Add a tree to the planting site

Expected: The "Delete" button becomes enabled, since the editor is allowed to delete the tree that they created.
Actual: The "Delete" button is disabled. If you refresh the page, the "Delete" button is correctly enabled.

RickMohr:
@jwalgran, that edge case is unrelated to this PR. Entered as #2664.

RickMohr:
@jwalgran, ready for another look.

jwalgran:
You mentioned the change at https://github.com/OpenTreeMap/otm-core/pull/2646/files#diff-d7d3eb4919444b4c006848eaa05ea901R746 should be reverted, but I don't see a reversion.

RickMohr:
Sorry, now pushed.

jwalgran:
👍

PR #2651, opened 2016-07-11
Explicitly set publicPath for production webpack configuration

maurizi:
`django-webpack-loader` defers to `publicPath` if it is available when
creating `<script>` tags, and otherwise constructs the path using the
bundle name and the value of the `BUNDLE_DIR_NAME` setting.

In practice, I was unable to find a value for `BUNDLE_DIR_NAME` that worked
for both locating the bundles on the local filesystem and creating URLs.

Setting `publicPath` explicitly (as we already do for development) works
around the issue.


RickMohr:
+1

Issue #2557, opened 2016-04-15
Cannot rename a UDF choice field choice if it is a number


jwalgran:
- Add a custom field to an instance and add choices that are numbers

<img width="898" alt="screen shot 2016-04-15 at 3 07 34 pm" src="https://cloud.githubusercontent.com/assets/17363/14576193/fb7040ca-031b-11e6-8ac0-9b5fd1045919.png">

- Attempt to rename the choice

Expected: The choice is renamed
Actual: Two different errors have appeared

- A validation error is returned that results in a confusing popup bubble

<img width="320" alt="screen shot 2016-04-15 at 3 12 02 pm" src="https://cloud.githubusercontent.com/assets/17363/14576281/7a09f962-031c-11e6-9f96-fcdee449558b.png">

```json
{"globalErrors": ["One or more of the specified values are invalid."], "fieldErrors": {"datatype": ["Choice '1' not found"]}}
```

- An exception is raised at https://github.com/OpenTreeMap/otm-core/blob/655d1f1eba79308f365be9d3087791c33af5a486/opentreemap/treemap/udf.py#L369

`DataError: invalid input syntax for integer: "M"`

PR #2653, opened 2016-07-12
Fix renaming numeric UDF field choices

kdeloach:
Numeric choice values are parsed as numbers on the frontend, which
causes a validation failure on the backend when renaming numeric
fields, because `1 != "1"`.

Fixed by ensuring that the validation logic compares choice values as
strings.

Connects #2557


maurizi:
:+1:

PR #2655, opened 2016-07-12
Make management sidebar click targets full height of row

lederer:
Addressing https://github.com/OpenTreeMap/otm-addons/issues/1144

There are still some gaps in the click target, to the left of the label. They are minimal, and due to the use of a `table` to do the layout of the sidebar.


jwalgran:
👍

Issue #2599, opened 2016-05-26
Tree identified as part of polygonal map feature when green infrastructure not turned on


dboyer:
On staging, see /testmap0523/features/1339114/

I added this tree via the normal add process. It indicates the tree is part of a polygonal map feature, but when I click to go to that map feature, I get an error page.

The Google Street View indicates the tree is an area associated with the "Independence National Historical Park" area. Perhaps this is related to search boundaries?

dboyer:
Replicated at /testmap0523/features/1330493/ but not at /testmap0523/features/1339113/

RickMohr:
A different map has a "Fairy Garden" at the same location -- see staging /testmap0316/map/?z=17/39.94740/-75.15163

PR #2656, opened 2016-07-12
Filter polygonal map features by current instance

kdeloach:
Fixes a bug where plots were being associated with polygonal map
features from other instances.

Test by creating a rain garden on Map A, then adding a tree at the same spot on Map B. The tree on Map B should *not* be associated with a polygonal map feature.

Connects #2599


RickMohr:
+1

Issue #2660, opened 2016-07-12
Search filters do not change if a geocoded address is active


kdeloach:
Steps:

1. Select a geocoded address from the locations autocomplete dropdown

2. Toggle between `Show trees` and `Show empty planting sites` under `Display` in the advanced search filters

Expect:
The map and the URL in the address bar should update to reflect the selected filters.

Actual:
Nothing happens.

kdeloach:
Verified on staging.

jwalgran:
Verified on production as well.

Issue #2466, opened 2016-01-22
Boundary searches are not cleared when clearing out the field and doing an address search


maurizi:
Noted by @steventlamb:

> all tests done on a clone of phillytreemap:
> If I search for "Queen Village" and opt for the geom in our db, it will highlight the boundary. If I then switch to an geocode address search, it will not clear the highlighted boundary, which is somewhat confusing.

PR #2661, opened 2016-07-13
Apply search filter changes when geocoded address is active

kdeloach:
This fixes a bug where advanced search filters would get stuck after
selecting a geocoded address in the locations autocomplete dropdown.

The Bacon stream responsible for triggering map and URL state changes
was only firing for non-geocode searches. I fixed this by merging in the
geocoding event stream so that map and URL state changes happen for both
geocode and non-geocode searches.

One unintended, but appreciated side-effect, is that selecting a
geocoded address will now remove the active boundary polygon from the
map. The previous behavior was a side-effect made possible only because we
were *not* applying search filters after selecting a geocoded address.

**Testing**

1. Verify that switching between a boundary and an address, clears the boundary polygon.

2. Verify that search filters work while a geocoded address is selected.

Connects #2466
Connects #2660

### DIFF
--- a/opentreemap/treemap/js/src/lib/mapPage.js
+++ b/opentreemap/treemap/js/src/lib/mapPage.js
@@ -35,9 +35,13 @@ module.exports.init = function (options) {
             searchBar.resetStream.map({})
         );
 
+    var geocodeEvents = searchBar.searchFiltersProp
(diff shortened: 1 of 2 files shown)


maurizi:
:+1:

Issue #2650, opened 2016-07-11
Show polygonal map feature area on the sidebar


jwalgran:
LA roof areas should be excluded.

PR #2665, opened 2016-07-13
Show polygonal feature area

jwalgran:
The new `calculate_area` method on `PolygonalMapFeature` executes a query using the accurate anywhere in the world PostGIS geography type.

`PolygonalMapFeature`s are rendered by the tiler like point feature, so we do not have access to the polygon for client-side area calculation. I opted to create a instance method that makes a separate area query rather than try annotating PolygonalMapFeature query sets because we never need the area of more than one feature at a time.

I have temporarily hard coded a square meter label into the templates until we have implemented unit configuration for stormwater map features.

---

##### Testing

Create a map with all three stormwater features and add one of each to the map. The Rain Garden and Bioswale should show a "Area" in square meters on both the map page sidebar and the detail page.

---

Connects to #2650

### DIFF
--- a/opentreemap/stormwater/models.py
+++ b/opentreemap/stormwater/models.py
@@ -33,6 +33,20 @@ class PolygonalMapFeature(MapFeature):
     def benefits(cls):
         return CountOnlyBenefitCalculator(cls)
 
+    def calculate_area(self):
+        """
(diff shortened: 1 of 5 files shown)


maurizi:
:+1:

Issue #2663, opened 2016-07-13
UTF grid does not load when only a single tree is present on the map


maurizi:
Seems to have been introduced by https://github.com/OpenTreeMap/otm-core/pull/2639.

Based on trial-and-error, it seems that if the page loads with a zoom level greater than 18, the UTF grid will not load (requests are not even sent), which will now occur if there is only one plot or if all plots on the map are in the same location.  This is causing several of our UI tests to fail.

Ideally we could fix this so that UTF grid tiles load no matter what the starting zoom level is.  Alternatively, adjusting the starting map extent so that it is no more than zoom level 18 would presumably also fix the issue.

PR #2671, opened 2016-07-15
Patch leaflet to respect maxZoom in Map.fitBounds

jwalgran:
We need this limiting when zooming the map to the extent of all features, but we are waiting for the Leaflet 1.0 release to upgrade the entire library.

---

##### Testing

Create an instance. Add a single plot, and refresh the map page. On `develop` the map feature will not be clickable because the map tells the UTF grid plugin that it is showing a zoom level > 1000 and the UTF grid plugin does not make tile requests. On this branch, the feature will be clickable.

---

Connects to #2663


maurizi:
:+1:

Issue #2648, opened 2016-07-11
Add boolean for show/hide ecobenefits for individual resource types


jwalgran:
Not the UI, just adding the boolean field for each resource type and updating the calculation code to respect it.

We decided there should be a boolean for each resource type.

Issue #2647, opened 2016-07-11
Add rain garden / bioswale eco backend


jwalgran:
Annual rainfall x area of polygon x constant number = Annual stormwater diverted

Rainfall and constant are instance manager supplied values. Hard code until the UI is ready.

RickMohr:
We have a growing collection of fields attached to a map feature type. Current fields:
* class name
* singular display name
* plural display name

New fields:
* should display ecobenefits
* percent stormwater diverted

Currently the fields are stored under two separate keys of `instance.config`:
* `map_feature_types`
* `terminology`

I propose that we combine these fields in a single dict. Defaults would be stored on the `MapFeature` subclass (as today with [`_terminology`](https://github.com/OpenTreeMap/otm-core/blob/8de1867f0af9be3e6570c6a48d19cda4f6cfde45/opentreemap/stormwater/models.py#L78)) with overrides stored on `instance.config.map_feature_types`.

For example, today the `Bioswale` class has a `_terminology` property:
```py
    _terminology = {
        'singular': 'Rain garden',
        'plural': 'Rain gardens'
    }
```
That would be replaced by a `default_config` property:
```py
    default_config = {
        'singular': 'Bioswale',
        'plural': 'Bioswales',
        'should_show_eco': True,
        'diversion_rate': 0.95
    }
```
`instance.config.map_feature_types` would become a dict instead of a list. For example, if you enable Bioswales for an instance and decide not to display ecobenefits, `instance.config.map_feature_types` would be:
```json
{
    "Bioswale": {
        "should_show_eco": false,
    }
}
```

jwalgran:
This looks like a good reorganization, and it is better to do this now, while there are not a lot of features to migrate. To keep it simple I would assume that if I turn off eco

```json
{
    "Bioswale": {
        "should_show_eco": false
    }
}
```

then turn it back on, I would end up with

```json
{
    "Bioswale": {
        "should_show_eco": true
    }
}
```

Is that what you were thinking?

RickMohr:
That is certainly simplest. It's not as nice for the UI because removing a map feature loses your overrides. But that's probably OK since map feature types are rarely if ever removed.

RickMohr:
[...]

PR #2673, opened 2016-07-16
Add back end support for green infrastructure ecobenefits

RickMohr:
* Add `PolygonalBasinBenefitCalculator` to compute `runoff_reduced` ecobenefit for both `RainGarden` and `Bioswale`
* Don't have to loop over features to calculate runoff reduced -- just sum the areas
* Refactor existing area computation code so it can be used for eco calculations also
* Rename the GSI eco category from `la_eco` to `green_infrastructure_eco`

Note: You must reprovision your app because of a `local_settings.j2` change in `otm-cloud`

Also, in `MapFeature` add `get_config` and `set_config_property` to manage configuration of a map feature type for an instance. For example, the `Bioswale` class declares its default configuration this way:
```py
    default_config = {
        'should_show_eco': True,
        'diversion_rate': 0.95
    }
```
Overrides for an instance are stored on `instance.config.map_feature_config`. For example, if a user decides not to display ecobenefits for Bioswales, `instance.config.map_feature_config` would be:
```json
{
    "Bioswale": {
        "should_show_eco": false
    }
}
```
Code referencing these properties should use `MapFeature.get_config`, and code setting these properties should use `set_config_property`.

Depends on OpenTreeMap/otm-cloud#256
Depends on OpenTreeMap/otm-addons#1213

Connects #2647
Connects #2648

### DIFF
--- a/opentreemap/stormwater/benefits.py
+++ b/opentreemap/stormwater/benefits.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from django.utils.translation import ugettext_lazy as _
+
+from treemap.ecobenefits import BenefitCalculator
+from map_features.benefits import (FEET_SQ_PER_METER_SQ, FEET_PER_INCH,
+                                   GALLONS_PER_CUBIC_FT)
+
+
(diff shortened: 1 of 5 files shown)


jwalgran:
👍

jwalgran:
Merging for Rick since he is out.

Issue #2638, opened 2016-06-27
Bulk uploader error message re: tree count when updating trees


hadleystein1:
Attempted to bulk *update* ID fields for 13,000 trees on Miami Beach Map: https://www.opentreemap.org/miamibeachtrees.

 Bulk uploader thinks the trees are being added and provides error message that prevents bulk update from being possible: "Your tree map currently contains 31,348 trees. This data upload will cause your tree map to exceed the 40,000 number of trees included in your plan. Please add more trees to your plan or upload 8,652 or fewer trees." 

![screen shot 2016-06-27 at 5 20 07 pm](https://cloud.githubusercontent.com/assets/15219245/16396160/68619a82-3c8b-11e6-9539-29d7ae7ddadf.png)

Because they are billed via Stripe, there is no way to manually (temporarily) update the tree count in the Super Admin page.

PR #2675, opened 2016-07-18
Feature/import update

RobinIsTheBird:
Bulk importer should not raise a tree limit exception if the imports are updates to existing
planting sites.

---

Changes:

1. importer/views.py - the basic fix, to `_get_tree_limit_context`, to only count csv rows without planting site ids toward the tree limit test
2. importer/tasks.py - fix race conditions in import event status update, which affect both tests and celery eager mode
3. importer/base.py - add a method to `GenericImportEvent` to support the race condition fixes
4. importer/tests.py - add a class `TreeLimitTests` to test for importing with row count exactly equal to limit (exercises race fixes), for importing too many new plots (regression test), and for importing updates that would fail the limit test if they were new plots

---

Testing:
* Automation: scripts/manage.sh test importer.tests.TreeLimitTests
* Manual:
   1. As super admin, set the tree count limit to something very small, like 3, and make instance a paid account
   2. As the instance admin, create trees up to the limit
   3. Export a csv of the instance
   4. Remove stewardship columns, if any (which the importer regards as invalid)
   5. Change a field in every row, e.g. diameter
   6. Save the changes as a csv, and import it
   7. When the import shows up in verified, click View, and scroll to the bottom to verify that there is no complaint about tree limit being exceeded
   8. Click Commit
   9. Verify that the import shows up in Creating and Finished, with no tree limit complaint

---

Connects to #2638

### DIFF
--- a/opentreemap/importer/models/base.py
+++ b/opentreemap/importer/models/base.py
@@ -70,6 +70,18 @@ class GenericImportEvent(models.Model):
         t = "Unknown Error While %s" if self.is_lost else "%s"
         return t % self.status_description()
 
+    def is_past_verifying_stage(self):
+        # This method is used by tasks for concurrency control
+        self.refresh_from_db()
+        return self.status in {
+            GenericImportEvent.FINISHED_VERIFICATION,
+            GenericImportEvent.CREATING,
+            GenericImportEvent.FINISHED_CREATING,
+            GenericImportEvent.FAILED_FILE_VERIFICATION,
+            GenericImportEvent.CANCELED,
+            GenericImportEvent.VERIFICATION_ERROR
+        }
+
     def status_description(self):
         summaries = {
             self.PENDING_VERIFICATION: "Not Yet Started",
--- a/opentreemap/importer/models/base.py
+++ b/opentreemap/importer/models/base.py
@@ -71,8 +71,6 @@ class GenericImportEvent(models.Model):
         return t % self.status_description()
 
     def is_past_verifying_stage(self):
-        # This method is used by tasks for concurrency control
-        self.refresh_from_db()
         return self.status in {
             GenericImportEvent.FINISHED_VERIFICATION,
             GenericImportEvent.CREATING,
--- a/opentreemap/importer/tasks.py
+++ b/opentreemap/importer/tasks.py
@@ -122,6 +122,7 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
 @transaction.atomic
 def _update_ie_status(ie):
     # Protect against race condition between task completion and main task
+    ie.refresh_from_db()
     if not ie.is_past_verifying_stage():
         ie.status = GenericImportEvent.VERIFIYING
         ie.update_progress_timestamp_and_save()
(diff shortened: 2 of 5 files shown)


opentreemap/importer/models/base.py
```diff
@@ -70,6 +70,18 @@ def status_summary(self):
         t = "Unknown Error While %s" if self.is_lost else "%s"
         return t % self.status_description()
 
+    def is_past_verifying_stage(self):
+        # This method is used by tasks for concurrency control
+        self.refresh_from_db()
```
jwalgran: I would not expect an `is_*` method that returns a boolean to have the side effect of querying the DB and possibly overriding properties I had edited on the import event model. We should consider moving the refresh call to [`tasks.py:103`](https://github.com/OpenTreeMap/otm-core/pull/2675/files#diff-52bab5e6c25271c634db1a91d4ae5792R104), just before we call `_update_ie_status`

jwalgran:
While we are in here I would like to take the opportunity to remove the awkwardness from the "over the limit" text. @dboyer are you ok with this update?

Before

> Your tree map currently contains 0 trees. This data upload will cause your tree map to exceed the 3 number of trees included in your plan. Please add more trees to your plan or upload 3 or fewer trees.

After

> Your tree map currently contains 0 trees. This data upload will cause your tree map to exceed the number of trees included in your plan (3). Please add more trees to your plan or upload 3 or fewer trees.

Diff

```diff
diff --git a/opentreemap/importer/templates/importer/partials/row_status.html b/opentreemap/importer/templates/importer/partials/row_status.html
index 53fa347..2b5d8f0 100644
--- a/opentreemap/importer/templates/importer/partials/row_status.html
+++ b/opentreemap/importer/templates/importer/partials/row_status.html
@@ -29,7 +29,9 @@
         <div class="alert alert-danger">
         {# TODO: Link to the plan page #}
         {% blocktrans %}
-        Your tree map currently contains {{ tree_count }} trees. This data upload will cause your tree map to exceed the {{ tree_limit }} number of trees included in your plan. Please add more trees to your plan or upload {{ remaining_tree_limit }} or fewer trees.
+        Your tree map currently contains {{ tree_count }} trees.
+        This data upload will cause your tree map to exceed the number of trees included in your plan ({{ tree_limit }}).
+        Please add more trees to your plan or upload {{ remaining_tree_limit }} or fewer trees.
         {% endblocktrans %}
         </div>
     {% elif ie.can_add_to_map %}
```

opentreemap/importer/tasks.py
```diff
@@ -118,6 +119,15 @@ def run_import_event_validation(import_type, import_event_id, file_obj):
         return
 
 
+@transaction.atomic
+def _update_ie_status(ie):
```
jwalgran: The name `_update_ie_status` may be too general for such a specific action. Since we are conditionally updating the status to a specific value, `_update_status_to_verifying_if_not_past` is a little wordy, but more self-describing.

jwalgran:
This works well. Nice, clean fix.

👍

dboyer:
@jwalgran Just saw this text change. Fine with me.

PR #2676, opened 2016-07-19
Fix UI tests

maurizi:
A couple of small changes here:
 - A code cleanup to remove a monkey patch that is no longer needed
 - Some changes to test setup to fix regressions introduced by the switch to Webpack
 - A fix to a UI test that seems to have started failing due to timing issues

Connects to OpenTreeMap/otm-cloud#258

### DIFF
--- a/opentreemap/treemap/tests/base.py
+++ b/opentreemap/treemap/tests/base.py
@@ -5,7 +5,7 @@ test_settings = {
     # Use a faster password hasher for unit tests to improve performance
     'PASSWORD_HASHERS': ('django.contrib.auth.hashers.MD5PasswordHasher',),
 
-    'STATIC_URL': 'http://localhost:/static/',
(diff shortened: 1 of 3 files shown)


RobinIsTheBird:
Waiting for all the ui tests to run before I give it the +1. So far, looks headed for +1.

RobinIsTheBird:
I'm going to give this a :+1: even though that one test still fails, because it's so much better than before.

Would it make sense to open a separate issue for the straggler?

maurizi:
> I'm going to give this a :+1: even though that one test still fails, because it's so much better than before.

> Would it make sense to open a separate issue for the straggler?

I want to get that last test working as well. I suspect adding some waiting is all it will need.

I'll re-assign this to you when I have the last test fixed.

RobinIsTheBird:
:+1:

PR #2680, opened 2016-07-21
Add green infrastructure unit default settings

RobinIsTheBird:
Adding green infrastructure fields to the units management page

---

connects to OpenTreeMap/otm-addons#1211


RobinIsTheBird:
connects to issue OpenTreeMap/otm-addons#1211

maurizi:
:+1:

Issue #2679, opened 2016-07-20
Searching from tree detail page does not return to map page


dboyer:
Tested on 2.10.0 on staging in Chrome and Firefox.

1. Viewed and edited a tree detail page
2. Entered a species in the search bar while still on the tree detail page
3. Clicked Search - the "advanced" button highlighted but the map page did not load
4. Opened Advanced search and saw that "Missing" had a blue dot indicating active searches for show missing species, show missing tree diameter, show missing photos

I had not configured any search filters and this map was newly created for testing. Searching by location has the same results.

dboyer:
Replicated on phillytreemap on staging.

dboyer:
Likely related - the Reset search button does not work when on a tree detail page.

PR #2682, opened 2016-07-21
Fix searchBar.js entry module.

maurizi:
This was a regression caused from the switch-over to using Webpack.
It was calling `searchBar.init()` when we should have been calling
`searchBar.initDefault()`, as `.init()` is intended for use on the map page.

Connects to #2679
Connects to OpenTreeMap/otm-clients#250


RobinIsTheBird:
:+1:

PR #2691, opened 2016-08-01
Minor changes in support of FUF

RobinIsTheBird:
Changes:
stormwater/models.py - default should_show_eco to False
treemap/models.py - save kwarg to determine whether setting a config attr
should save the instance

--

Connects to OpenTreeMap/otm-addons#1209
Depends on OpenTreeMap/otm-cloud#263


maurizi:
:+1:

Issue #2686, opened 2016-07-26
It is possible to save a stewardship entry with an invalid date, resulting in a crash when trying to view the map feature - we should require 4 digits for years


jwalgran:
- Create an instance
- Add a plot with a tree
- Edit the map feature
- Add a tree stewardship item and set the date to `7/26/16`
- Save the map feature

Error behavior: After clicking the `+` button to add the stewardship entry you can see the date formatted as `0016-07-26`. This value passes validation, presumably because it technically a valid date, but all future attempts to load the map feature fail while trying to convert the date to a string.

```
>>> Tree.objects.get(pk=2077090)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 255, in iterator
    obj = model_cls.from_db(db, init_list, row[model_fields_start:model_fields_end])
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 489, in from_db
    new = cls(*values)
  File "/opt/app/core/opentreemap/treemap/units.py", line 23, in __init__
    super(Convertible, self).__init__(*args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/udf.py", line 1078, in __init__
    super(UDFModel, self).__init__(*args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/audit.py", line 704, in __init__
    super(Authorizable, self).__init__(*args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/audit.py", line 1045, in __init__
    super(_PendingAuditable, self).__init__(*args, **kwargs)
  File "/opt/app/core/opentreemap/treemap/audit.py", line 488, in __init__
    self.populate_previous_state()
  File "/opt/app/core/opentreemap/treemap/audit.py", line 582, in populate_previous_state
[...]

PR #2695, opened 2016-08-04
Prevent errors from invalid date UDFs

kdeloach:
Prevents invalid datetime UDFs from being saved, by adding extra server-side validation. The client-side datepicker has also been updated to disallow entering dates before `1/1/1900`.

**Note:** This PR *does not* clean up existing invalid UDF data. The `treemap_userdefinedcollectionvalue` and `treemap_audit` tables need to be corrected by hand. The plot detail page will continue to raise a validation error until the data has been fixed.

Connects #2686

### DIFF
--- a/opentreemap/treemap/templates/treemap/partials/udf_edit_field.html
+++ b/opentreemap/treemap/templates/treemap/partials/udf_edit_field.html
@@ -9,6 +9,8 @@
 </select>
 {% elif field.type == 'date' %}
 <input class="form-control" type="text" data-field-name="{{ field.name }}"
+       data-date-start-date="{{ datepicker_start_date|date:request.instance.short_date_format|datepicker_format }}"
+       data-date-assume-nearby-year="true"
        data-date-display-format="{{ request.instance.short_date_format|moment_format }}"
        data-date-format="{{ request.instance.short_date_format|datepicker_format }}"
        data-date-serialize-format="YYYY-MM-DD" value="{% trans "Today" %}" />
--- a/assets/css/vendor/datepicker.css
+++ b/assets/css/vendor/datepicker.css
@@ -1,22 +1,13 @@
 /*!
- * Datepicker for Bootstrap
+ * Datepicker for Bootstrap v1.6.4 (https://github.com/eternicode/bootstrap-datepicker)
  *
  * Copyright 2012 Stefan Petre
  * Improvements by Andrew Rowls
- * Licensed under the Apache License v2.0
- * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * Licensed under the Apache License v2.0 (http://www.apache.org/licenses/LICENSE-2.0)
  */
 .datepicker {
-  padding: 4px;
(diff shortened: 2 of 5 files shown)


opentreemap/treemap/templates/treemap/partials/udf_edit_field.html
```diff
@@ -9,6 +9,7 @@
 </select>
 {% elif field.type == 'date' %}
 <input class="form-control" type="text" data-field-name="{{ field.name }}"
+       data-date-start-date="{{ datepicker_start_date|date:request.instance.short_date_format|datepicker_format }}"
```
jwalgran: This is a nice, small, working fix to a temporary problem (living trees were planted before 1900, so we need to remove this date restriction from the application in general) however the most common workflow has a problem.

1. On the edit view of a tree detail page, scroll to the planting site stewardship section
2. Focus the date field, enter "8/5/16" then immediately click the "+" button without unfocusing the date field.

This will add a stewardship row with the date 01/01/1900, which is not what the user entered, and there is no delete button to correct it.

I would suggest attempting to add some JS that checks the value and shows some error "toast" rather than adding a row if the date is equal to the minimum value. In this situation, we may want to consider setting the minimum on the date picker to 1899-12-31 so that the legitimate minimum value is still selectable.

kdeloach: Good point. This is really a bandaid over a larger issue. We'll likely need to use a third party library to handle parsing and formatting dates prior to 1/1/1900.

The latest version of bootstrap datepicker has an option that would be very useful here called [assumeNearbyYear](https://bootstrap-datepicker.readthedocs.io/en/latest/options.html#assumenearbyyear). I could try to implement something similar, where, if the user entered a year that was interpreted as `00xx` we convert it to `20xx`?

jwalgran: `assumeNearbyYear` makes a better guess, but we still run into the problem of the date picker guessing what the user means, and then not giving them the opportunity to correct it. I'll ask the team for some more feedback.

jwalgran: As a short term, reasonable fix I would like to try updating the datepicker and enabling `assumeNearbyYear`.

kdeloach: Will do.

kdeloach: I upgraded the bootstrap datepicker and enabled `assumeNearbyYear`. I also verified that the start date option is inclusive, so entering "1/1/1900" will work. I did not add a toast error message.

jwalgran:
Thanks for researching the updated library version. Works well. 

👍

kdeloach:
Let me know if you would also like me to follow-up with a data migration to fix existing records.

jwalgran:
Thanks. I manually fixed the only pre-1900 value in the production database, so a data migration won't be required.

Issue #2688, opened 2016-07-28
Add drainage area square footage as an optional field on rain gardens and bioswales


jwalgran:
After further discussions with our advisors, we have determined that eco benefit calculations for these resources will need to take into account the square footage of area that drains to the rain garden/bioswale in addition to the area of the rain garden/bioswale itself.

Use rain barrels as a reference, since they already collect and display a capacity field.

dboyer:
Remember to show and collect in the map's units, but convert and store in a specific database units.

RickMohr:
Users could be confused whether the drainage area should include or omit the area of the rain garden, and we don't currently have a mechanism for including help text.

dboyer:
Could we repurpose how we use help text by the diameter?

RobinIsTheBird:
Both tree species and diameter have help text that are hyperlinks off site, and the latter is a video so it has a video icon.

A hyperlink strikes me as overkill for this clarification, which could be simply stated in one sentence  (@dboyer, can you provide one?)

I could see either including the help text right below the input, like species and diameter, or adding an info icon (i in a filled circle) to the label, with a tooltip on hover. There may be many other solutions, but those are what leap to my mind.

What would you like to see?

dboyer:
Just saw this! A line of text by the field is fine with me. Text is:

Please enter an estimate of the area that drains to the feature. Do not include the area of the feature itself.

If possible, I'd like "feature" to use whatever term the map owner has entered for "resource."

RickMohr:
Since the field will be shown on the detail page for a rain garden or bioswale we can use the actual resource name, e.g.

Please enter an estimate of the area that drains to the bioswale. Do not include the area of the bioswale itself.

PR #2697, opened 2016-08-08
Feature/drainage area

RobinIsTheBird:
Code to store, add, display and edit the new drainage_area field on Bioswale and RainGarden, including a migration to add the drainage_area field in the database and another to invalidate the cache.

--

Connects to #2688

### DIFF
--- a/opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
+++ b/opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.db.models import F
+
+
+def add_permissions_all(apps, schema_editor):
+    Instance = apps.get_model('treemap', 'Instance')
+    Role = apps.get_model('treemap', 'Role')
+    FieldPermission = apps.get_model('treemap', 'FieldPermission')
+
+    # Adapted from treemap.audit.add_default_permissions.
+    # We can't use that function directly, partly because of its
+    # dependencies on model classes, which aren't the right version
+    # for a migration, and partly because a migration is a snapshot
+    # in time, and we need to snapshot the logic as well.
+    def update_field_permissions(perm_specs, role, instance):
+        existing = FieldPermission.objects.filter(
+            role=role, instance=instance)
+        if existing.exists():
+            for perm in perm_specs:
+                perm['defaults'] = {
+                    'permission_level': role.default_permission}
+                FieldPermission.objects.get_or_create(**perm)
+            return False
+        else:
+            perm_specs = [FieldPermission(**perm) for perm in perm_specs]
+            for perm in perm_specs:
+                perm.permission_level = role.default_permission
+            FieldPermission.objects.bulk_create(perm_specs)
+            return True
+
+    for instance in Instance.objects.filter(config__contains='"Plot",'):
+        types = instance.config['map_feature_types']
+        if 'Bioswale' in types or 'RainGarden' in types:
+            for role in Role.objects.filter(instance=instance):
+                perm_specs = [
+                    {'model_name': 'Bioswale',
+                     'field_name': 'drainage_area',
+                     'role': role,
+                     'instance': instance},
+                    {'model_name': 'RainGarden',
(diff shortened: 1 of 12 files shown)


RobinIsTheBird:
@azavea-bot rebuild

opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
```diff
@@ -0,0 +1,28 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('stormwater', '0006_stormwater_drainage_area'),
+        ('treemap', '0011_instance_universal_rev'),
+        ('treemap', '0026_add_canopy_fields'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_permissions_all)
```
maurizi: You should add a `noop` backwards migration, so that we can rollback this migration if necessary (see https://docs.djangoproject.com/ja/1.9/ref/migration-operations/#django.db.migrations.operations.RunPython.noop)

RobinIsTheBird:
@azavea-bot rebuild

opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
```diff
@@ -0,0 +1,67 @@
...
+                     'field_name': 'drainage_area',
+                     'role': role,
+                     'instance': instance}]
+                need_to_clear_cache = \
+                    update_field_permissions(perm_specs, role, instance) or \
+                    need_to_clear_cache
+
+    if need_to_clear_cache:
+        from object_caches import clear_caches
+        clear_caches()
```
maurizi: `clear_caches()` is not the method we want to call here - it will clear the in-memory cache, but will not update the cache key.  The method we need to call to do that is [`increment_adjuncts_timestamp`](https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/lib/object_caches.py#L63-L72) (though we can't actually call it directly, because we need to use migration models...), and by incrementing it, the next attempt to access the cache will clear the cache for us.

Instead, we should update the cache key on a per instance basis:
```python
for instance in Instance.objects.filter(config__contains='"Plot",'):
    need_to_clear_cache = False
    types = instance.config['map_feature_types']
        if 'Bioswale' in types or 'RainGarden' in types:
            for role in Role.objects.filter(instance=instance):
                ...
                ...
                if update_field_permissions(perm_specs, role, instance):
                    need_to_clear_cache = True

    if need_to_clear_cache:
        # Copied from object_caches.increment_adjuncts_timestamp
        Instance.objects.filter(pk=instance.id) \
            .update(adjuncts_timestamp=F('adjuncts_timestamp') + 1)
```

jwalgran: I recommended `clear_caches` to get around import issues, thinking that we do not care about the cache key value if we are clearing the entire cache. I did not think about the cache in other processes. My mistake.

Do we need  to worry about looping over the `Role`s to set the `need_to_clear_cache` flag? To simplify things a little, can't we just increment the `adjuncts_timestamp` on any instance with bioswales or rain gardens in `map_feature_types`?

RickMohr: I had the same thought -- there's no harm in incrementing the adjuncts timestamp for more instances if it simplifies the code.

maurizi: Well we're looping over the `Role`s anyways to update field permissions.... but yes, it would definitely be fine to just increment `adjuncts_timestamp` on any `Instance` with Bioswales or Rain Gardens - the only problem with incrementing the timestamp too much would be performance concerns, and there aren't many instances with those map features enabled.

RobinIsTheBird:
Bitten by #2698 again. I can't imagine anything I could have done to affect this specific test, and notice that it passed on the other checks.

maurizi:
The units are not displaying properly on in the add process or on the detail page.

Instead of seeing `m²` or `meters²` the unit appears as `sq_m`.

Additionally, changing the unit type on the units management page does not change the unit or convert the displayed value.
sq_m

maurizi:
Once the `sq_m` bug is fixed, this should be good to go :+1:

PR #2705, opened 2016-08-10
Add Tampa Tree Map migration rules 

jwalgran:
I am committing these customer-specific rules to this repository because
similar rule files already exist in
`opentreemap/otm1_migrator/migration_rules`.

I based these Tampa rules on:

* The previous Philadelphia and San Diego rules.
* The fields available on the OTM1 tree detail pages.
* The choices JSON returned from the OTM1 `/choices/` endpoint.

This PR also includes a bug fix to the migrator. If a user of an OTM1 account had created an OTM2 account using the same email address, but had never logged in with that account, an exception would be thrown when attempting to migrate the users. This commit avoids the exception by checking for a `None` value.

Connects to https://github.com/OpenTreeMap/otm-clients/issues/256


RickMohr:
+1

Issue #2703, opened 2016-08-10
instance config does not play well with test infrastructure


RobinIsTheBird:
An otm-addons test best demonstrates the problem, but I believe the underlying problem is in otm-core.

The demonstration test case is otm-addons.cloud_management.tests.test_management.UnitsTest.

The following code excerpt succeeds on my localhost but fails on Jenkins, even if I run all 11 1/2 minutes worth of tests locally.

```
        instance = make_instance(url_name='basic')
        owner = make_admin_user(instance)
        ei = EnhancedInstance(instance=instance, owner=owner)
        ei.feature_enabled_green_infrastructure = False
        ei.save_with_user(owner)
        specs = units(None, instance)

        self.assertEqual(len(specs['value_specs']), 3)
```

On Jenkins, it fails, saying 4 != 3. Instrumentation reveals that on Jenkins, the config includes Bioswale.

If you add the following line before the call to `units()`, it then passes on Jenkins.

`        instance.remove_map_feature_types(keep=['Plot'])
`

A new instance from make_instance ought not to have any stormwater map feature types in its config.

RickMohr:
A likely suspect is mistakenly updating the default value of an instance config property. A quick thing to check would be changing this line of code:

https://github.com/OpenTreeMap/otm-core/blob/5c139aea0a34e3420239e4ba428924fb448000e3/opentreemap/treemap/instance.py#L243

to something like:

```
return self.config.get(prop, deepcopy(default))
```

maurizi:
Another possible solution is to bring in this Python module and run `freeze` on all of our default config dictionaries  - which would fix the issue for things used as `_make_config_property` defaults and prevent any problems when/if they are used outside of `_make_config_property`

RobinIsTheBird:
I introduced the bug, and the bug is here: https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/stormwater/tests.py#L22

That said, I think the list ought to be kept as a tuple to prevent noob errors like that, which are so costly to track down or work around. I'll make both changes tomorrow.

RickMohr:
My `deepcopy` proposal would prevent code like this from affecting the default value.

RobinIsTheBird:
Yep, on closer inspection, you are correct Rick. I'm using deepcopy.

I've also renamed map_feature_types to _map_feature_types and added a read-only map_feature_types property, as well as docstrings explaining how to use it.

RobinIsTheBird:
[...]

PR #2709, opened 2016-08-12
Fix broken test and make the map_feature_types property safer

RobinIsTheBird:
UdfGenericCreateTest.test_non_treemap_model was simply appending 'Bioswale' to
map_feature_types with the += operator, bypassing add_map_feature_types.
Fix it to use add_map_feature_types.

In instance, return deepcopy(default) from the getter of the property returned
by _make_config_property, to prevent any code from changing the default for all
instances by mistake.

Additionally, replace map_feature_types with \_map_feature_types, to signal that
the setter is "private", and provide a get-only map_feature_types property,
with a docstring pointing at add_ & remove_map_feature_types functions.

--

Connects to #2703
Depends on OpenTreeMap/otm-addons#1227

### DIFF
--- a/opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
+++ b/opentreemap/stormwater/migrations/0007_drainage_area_permissions.py
@@ -32,7 +32,7 @@ def add_permissions_all(apps, schema_editor):
             return True
 
     for instance in Instance.objects.filter(config__contains='"Plot",'):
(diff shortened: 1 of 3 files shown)


RobinIsTheBird:
@azavea-bot rebuild

RickMohr:
+1 Great idea to rename to `_map_feature_types`

Issue #2707, opened 2016-08-11
Newly-added bioswale doesn't contribute to total ecobenefits


RickMohr:
If you add a bioswale and select it you will see its ecobenefits but they do not contribute to the total ecobenefits. If you modify one of its fields then it will contribute to the total.

RobinIsTheBird:
I've also seen removed resources continuing to be reflected in the total ecobenefits. Clearly a caching issue.

RobinIsTheBird:
Puzzling. The code looks like it does the right thing with the cache. Must be something else.

RobinIsTheBird:
I know I've seen this sort of misbehavior in the past, but now I can't reproduce it.

RickMohr:
This happened during RRP on staging. Closing because @RobinIsTheBird can't reproduce it and says the code looks right.

Issue #2689, opened 2016-07-28
Incorporate drainage area into rain garden / bioswale eco benefit calculation


jwalgran:
The drainage area is a new field added in https://github.com/OpenTreeMap/otm-core/issues/2688

The new proposed equation is:

Annual rainfall x area of polygon x square footage of area that drains to the rain garden x constant number = Annual stormwater diverted

Drainage area is not a required field so when calculating aggregate benefits we need to skip over features that do not have a value, and reflect that in the "basis."

dboyer:
Perhaps instead:

Annual rainfall x (area of polygon + (square footage of area that drains to the rain garden x constant number ))

dboyer:
The client confirmed that is the correct equation.

dboyer:
The constnat number should be 0.85

RickMohr:
Just to clarify, you mean the default value for the "Percent Stormwater Diverted" constant should be 0.85, but users should be able to change it, right?

dboyer:
Correct. I'm guessing most people will not change the default, but editing it should be an option for map owners. 

I'm working on confirmation of what we should call that field. Civil engineers sometimes refer to it as the "runoff coefficient."

RobinIsTheBird:
Request for clarification:

@dboyer I presume that the client confirmed mmaurizi's equation, right?

On the Green Infrastructure management page, we are presenting the number as 0..100, although internally it is converted to a fraction. Is this the desired display representation?

Last but not least, are we ready to commit to calling it "runoff coefficient" yet?

dboyer:
Just received confirmation that we can call it "runoff coefficient." The equation we proposed (below) is correct.

Annual rainfall x (area of polygon + (square footage of area that drains to the rain garden x constant number ))

Displaying it as 0.85 on the management page would be preferable as that's how it is displayed as part of something called the "Rational Equation Method" - http://www.lmnoeng.com/Hydrology/rational.php - which is often used for this type of GSI calculation.

PR #2710, opened 2016-08-15
Update stormwater benefit calculation to include drainage area

RobinIsTheBird:
Note: you might still see incorrect aggregate results in your browser due to
OpenTreeMap/otm-core#2707. Going in with the shell and kicking universal_revs a
couple times seems to resolve it.

Added the new calculation to stormwater/benefits.py. Since drainage_area is not
in PolygonalMapFeature, had to change the interface to the internal
_benefits_for_feature_qs and pass in a qs of the original map feature class.

The new calculation also calls for a default constant of 0.85.

Updated tests accordingly.

--

Connected to #2689

### DIFF
--- a/opentreemap/stormwater/tests.py
+++ b/opentreemap/stormwater/tests.py
@@ -32,22 +32,12 @@ class UdfGenericCreateTest(UdfCRUTestCase):
 class PolygonalMapFeatureTest(OTMTestCase):
     def setUp(self):
         (x, y) = -76, 39
-        self.point = Point(x, y, srid=4326)
-        self.point.transform(3857)
-        d = 0.1
-        self.polygon = MultiPolygon(
-            Polygon(
-                ((x, y),
-                 (x, y + d),
-                 (x + d, y + d),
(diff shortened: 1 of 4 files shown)


opentreemap/stormwater/tests.py
```diff
@@ -85,36 +95,58 @@ def _get_runoff_reduced(self, feature, expect_empty=False):
...
         runoff_reduced = self._get_runoff_reduced(feature, expect_empty=True)
         self.assertEqual(runoff_reduced, 0)
 
     def test_bulk(self):
-        self._make_map_feature(Bioswale)
-        self._make_map_feature(Bioswale)
+        drainage_area_sq_meters = 100000000.0
+        # NOTE
+        # Today, no check is made for overlapping stormwater resources.
+        # In the future, either overlap should be invalid,
```
jwalgran: I would like to open an issue to implement a validation that prevents creating overlapping polygonal map features.

jwalgran:
👍

Issue #2706, opened 2016-08-11
Default units for Rain Garden / Bioswale area and drainage  area should be in imperial units, not metric


maurizi:
Also, the area field should use `meter²` instead of `m²`

maurizi:
Also ensure that the calculated area is going through the unit conversion process

PR #2717, opened 2016-08-17
Default area units should be imperial

RobinIsTheBird:
Rain Garden and Bioswale area and drainage area default units should
be square feet, not square meters.

Changes:
* Add Rain Garden, Bioswale, and Green Infrastructure to the
  STORAGE_UNITS default setting
* Add a conversion from square feet to square meters or feet
  to units
* In lib/map_feature, change context_dict_for_resource to
  return treemap units conversion for the resource area

--

Connects to #2706

### DIFF
--- a/opentreemap/opentreemap/settings/default_settings.py
+++ b/opentreemap/opentreemap/settings/default_settings.py
@@ -357,7 +357,17 @@ REGISTRATION_EMAIL_HTML = False
 # STORAGE_UNITS is the unit the value will be stored/computed as,
 # if different from DISPLAY_DEFAULTS
 #
-STORAGE_UNITS = {}
+STORAGE_UNITS = {
+    'bioswale': {
+        'drainage_area': 'sq_m'
+    },
+    'rainGarden': {
+        'drainage_area': 'sq_m'
+    },
+    'greenInfrastructure': {
+        'area': 'sq_m'
+    }
+}
 DISPLAY_DEFAULTS = {
     'plot': {
         'width':  {'units': 'in', 'digits': 1},
@@ -376,17 +386,17 @@ DISPLAY_DEFAULTS = {
         'airquality': {'units': 'lbs/year', 'digits': 1}
     },
     'bioswale': {
-        'drainage_area': {'units': 'sq_m', 'digits': 1}
+        'drainage_area': {'units': 'sq_ft', 'digits': 1}
     },
     'rainBarrel': {
         'capacity': {'units': 'gal', 'digits': 1}
     },
     'rainGarden': {
-        'drainage_area': {'units': 'sq_m', 'digits': 1}
+        'drainage_area': {'units': 'sq_ft', 'digits': 1}
     },
     'greenInfrastructure': {
         'rainfall': {'units': 'in', 'digits': 1},
-        'area':     {'units': 'sq_m', 'digits': 1}
+        'area':     {'units': 'sq_ft', 'digits': 1}
     }
 }
 
--- a/opentreemap/opentreemap/settings/default_settings.py
+++ b/opentreemap/opentreemap/settings/default_settings.py
@@ -358,12 +358,6 @@ REGISTRATION_EMAIL_HTML = False
 # if different from DISPLAY_DEFAULTS
 #
 STORAGE_UNITS = {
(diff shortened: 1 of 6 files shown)


opentreemap/treemap/lib/map_feature.py
```diff
@@ -312,8 +313,12 @@ def context_dict_for_resource(request, resource, **kwargs):
     if isinstance(resource, PolygonalMapFeature):
         context['contained_plots'] = resource.contained_plots()
         # TODO: Convert to map owner prefered units
```
maurizi: Can we get rid of this `TODO` comment now?

opentreemap/opentreemap/settings/default_settings.py
```diff
@@ -357,7 +357,17 @@
...
+    'bioswale': {
+        'drainage_area': 'sq_m'
+    },
+    'rainGarden': {
+        'drainage_area': 'sq_m'
+    },
+    'greenInfrastructure': {
+        'area': 'sq_m'
+    }
+}
```
maurizi: All of our other storage units are in imperial.

Since we haven't deployed this feature to production yet, we don't have any data migration issues to worry about, so I think we should use imperial units for storage for these fields as well (and continue to leave `STORAGE_UNITS` empty)

opentreemap/treemap/units.py
```diff
@@ -98,6 +98,9 @@ def convert_to_database_units(self):
 }
 _unit_conversions["ft"] = {u: v * 12 for (u, v)
                            in _unit_conversions["in"].iteritems()}
+_unit_conversions["sq_ft"] = {
+    u: v / _unit_conversions["sq_m"]["sq_ft"]
+    for u, v in _unit_conversions["sq_m"].iteritems()}
```
maurizi: If we switch to storing `drainage_area`, I think we can remove `sq_m` from the `_unit_conversions` dict and replace it with a line for converting `sq_ft` like so:

```diff
diff --git a/opentreemap/treemap/units.py b/opentreemap/treemap/units.py
index 2e7152b..57d1686 100644
--- a/opentreemap/treemap/units.py
+++ b/opentreemap/treemap/units.py
@@ -94,13 +94,10 @@ _unit_conversions = {
     "gal": {"gal": 1, "L": 3.785},
     "gal/year": {"gal/year": 1, "L/year": 3.785},
     "kwh/year": {"kwh/year": 1},
-    "sq_m": {"sq_m": 1, "sq_ft": 10.7639}
+    "sq_ft": {"sq_ft": 1, "sq_m": 0.092903}
 }
 _unit_conversions["ft"] = {u: v * 12 for (u, v)
                            in _unit_conversions["in"].iteritems()}
-_unit_conversions["sq_ft"] = {
-    u: v / _unit_conversions["sq_m"]["sq_ft"]
-    for u, v in _unit_conversions["sq_m"].iteritems()}
 
 
 def get_unit_name(abbrev):
```

opentreemap/opentreemap/settings/default_settings.py
RobinIsTheBird: Polygon coordinates are in roughly meters, using map projection 3857 for use with standard tiling systems. The PostGIS that calculates area calculates sq meters because the points are in meters, which makes perfect sense. So I'd definitely prefer to keep greenInfrastructure.area natively in sq_m, since that's closest to what's really in the db.

I can see an argument for changing drainage_area, though.

Given that this is a 1-point task, I'm not sure it's worth the work. How strongly do you feel about it?

opentreemap/treemap/units.py
RobinIsTheBird: Only if we switch both drainage_area and greenInfrastructure.area, and I'm very reluctant to switch the latter.

opentreemap/treemap/lib/map_feature.py
RobinIsTheBird: Yes. Meant to do that.

opentreemap/opentreemap/settings/default_settings.py
maurizi: Makes sense. As we discussed, just changing `drainage_area` should be fine.

maurizi:
:+1:

Issue #2667, opened 2016-07-14
After editing a rain garden / bioswale polygon on the detail page, the area field does not update


jwalgran:
1. Add a rain garden and go to its detail page
2. Edit the rain garden's polygon and save
3. The area field does not update
4. Refresh the page and the area is updated

RickMohr:
On a successful save, re-fetch and replace inner content of page.
Also replace special-purpose code that was introduced to solve similar problems without a refresh. (Possibly review past issues to locate such code.)

jwalgran:
I was testing https://github.com/OpenTreeMap/otm-core/pull/2710/files and saw that the eco benefits were not updating after editing the drainage area. Will the solution you are working on handle eco benefits?

RickMohr:
Yes!

Issue #2664, opened 2016-07-13
Plot detail: Enabled state of "Delete" button not refreshed after Edit / Save


RickMohr:
You will need a user with role "Editor" and another user with role "Administrator".
* As the admin user, make plot length "Read Only" for Editors
* As the editor user, create an empty planting site, go to plot detail, "Edit", "Add a Tree", and "Save".

Expected: The "Delete" button becomes enabled, since the editor is allowed to delete the tree.
Actual: The "Delete" button is disabled. If you refresh the page, the "Delete" button is correctly enabled.

Issue #1719, opened 2014-10-09
Refresh all tree detail subsections when photos or inline edit form changes are saved


jwalgran:
1. Create an empty plot and go to the detail page for that plot
2. Add a photo

Expected (and what appears after a page refresh):

![screen shot 2014-10-08 at 11 32 46 pm](https://cloud.githubusercontent.com/assets/17363/4570747/0445f38c-4f65-11e4-85e3-646eaeb7e48d.png)

Actual:

![screen shot 2014-10-08 at 11 32 31 pm](https://cloud.githubusercontent.com/assets/17363/4570755/1233dba8-4f65-11e4-8d3d-4b713b83c57f.png)

RickMohr:
Problem still exists as described.

Issue #1585, opened 2014-09-05
After saving on tree detail page, the recent edits in the sidebar are not updated


jwalgran:
After a successful inline edit form save, there should be an ajax request for recent edits sidebar content.

RickMohr:
This is fixed.

PR #2720, opened 2016-08-19
Refresh details when hitting "Save" on a plot/resource detail page

RickMohr:
* Refactor templates and JS to allow reloading map feature details as a partial
* The partials (which extend `partials/map_feature_detail_base.html`) include the title area, action buttons, and detail area.
* The partials do not include the mini map and streetview, since that would cause an obvious flash
* Detail page tells inline edit form not to update on successful save (we'll wait for refresh)
* Show spinner on both "Save" and "Delete"
* Simplify code that was directly updating details

NOTE - looking at individual commits will make it easier to review the code, since there were a couple of re-indentation steps

Depends on OpenTreeMap/otm-addons#1230

Connects #2667
Connects #2664
Connects #1719
Connects #975

Testing
* Verify that the above open bugs are fixed
* Verify that these closed bugs are still fixed
  * #1585
  * #861
  * #627
  * #596
  * #396
* Verify that you can still add and delete a tree many times in succession
* Verify that after saving you can still edit and save everything

### DIFF
--- a/opentreemap/stormwater/templates/stormwater/Bioswale_detail.html
+++ b/opentreemap/stormwater/templates/stormwater/Bioswale_detail.html
@@ -1,4 +1,4 @@
-{% extends "treemap/resource_detail.html" %}
+{% extends "treemap/partials/resource_detail.html" %}
 {% load i18n %}
 {% load l10n %}
 {% load form_extras %}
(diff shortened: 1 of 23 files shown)


maurizi:
If I update a multichoice field like perennial plants, those changes are not reflected when I hit save, but there are when I refresh the page.

maurizi:
Everything is looking pretty good though, nice refactor! Let me know when to give it another look.

RickMohr:
@azavea-bot rebuild

RickMohr:
@maurizi, thanks for the great find. Ready for another look.

Issue #2652, opened 2016-07-11
A popup is not displayed when clicking a stormwater feature


jwalgran:
1. Create an instance with the green infrastructure module
1. Activate "Rain Gardens"
1. Add a tree 
1. Add a rain garden
1. On the map page, click the tree, then click the rain garden

Expected: A popup is shown over every clicked feature
Actual: A popup is shown over a clicked tree, but not a clicked rain garden

In the network tab of the web inspector you can see a successful request for the rain garden popup content, the content is just not being displayed.

RickMohr:
I wonder if it's collateral damage from our recent work updating the popups with canopy info.

PR #2723, opened 2016-08-22
Display polygonal map feature popups

RobinIsTheBird:
The map returned by MapManager.js createTreeMap needs to have
utfEvents that merge the plotUtfEventStream with a zip
that combines fields from the emptyUtfEventStream with
the fields returned byt the polygonDataStream.

Using lodash's merge rather than assign to make a deep copy,
because I don't know the potential side effects of different stream
elements sharing descendant objects.

Let me know if you think _.assign would be sufficient.

--

Connects to #2652

### DIFF
--- a/opentreemap/treemap/js/src/lib/MapManager.js
+++ b/opentreemap/treemap/js/src/lib/MapManager.js
@@ -89,7 +89,9 @@ MapManager.prototype = {
 
                 map.utfEvents = Bacon.mergeAll(
                     plotUtfEventStream,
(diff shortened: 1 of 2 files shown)


RickMohr:
+1 Nice digesting of some complex Bacon code.
I think it would be fine to use `_.assign` (or `_.extend` as we've more typically used) because all uses of the UTF and polygon stream data are read-only. On the other hand I have no problem with `_.merge` as the cost of the deepcopy seems minimal.

Issue #2692, opened 2016-08-01
Extra space above header on plot detail page (styling regression?)


RickMohr:
![image](https://cloud.githubusercontent.com/assets/2162993/17303418/790e3a88-57d5-11e6-8f52-699e6c293a95.png)

RickMohr:
This doesn't appear to be a regression. But to me it looks better this way:

![image](https://cloud.githubusercontent.com/assets/2162993/17933941/03f71dac-69e4-11e6-944a-88b06190a770.png)

Issue #1553, opened 2014-08-18
Styling issue - top level comments should not be indented


maurizi:
![snapshot1](https://cloud.githubusercontent.com/assets/4432106/3956178/4f44fdc2-2708-11e4-8230-c56b69324286.png)

PR #2727, opened 2016-08-24
Reduce top spacing in detail page sidebar bubbles

RickMohr:
A 25px top margin was introduced [here](https://github.com/OpenTreeMap/otm-core/commit/6c90967#diff-bf0bf7f1b303c2fd6c2d997a3726893eR9) as part of fixing spacing in nested comments for #1553.

Removing that margin makes the sidebar bubbles look right again, and nested comments still look fine.

Connects #2692


lederer:
👍

Issue #2711, opened 2016-08-15
Spelling out units (instead of using abbreviation) causes layout problems


RickMohr:
Because we're now showing "inches" instead of "in" there's not enough room to enter tree diameter in a non-full-width browser:

![image](https://cloud.githubusercontent.com/assets/2162993/17675415/b30962ee-62f7-11e6-820a-8ce83be3289e.png)

Why did we suddenly decide to change that presentation design?

maurizi:
The change was made in https://github.com/OpenTreeMap/otm-core/pull/2697 because with the existing system, we were using internal `dict` keys to display items, which caused the unit for drainage area to display as `sq_m`.

We were also being inconsistent with unit display, showing "feet" in the sidebar, but "ft" on the field row.

The layout problems are a concern though.  Perhaps we should explicitly use abbreviations everywhere? Or try and change the design for the diameter calculator to fit the longer unit names?

jwalgran:
I put this in a new 2.11.0 milestone because I think we need to fix the layout before we ship.

Adding a dictionary of unit keys to abbreviations (or incorporating that mapping into an existing dictionary) seems like the most straightforward solution.

dboyer:
Agreed we need to fix this before the next release. Adding to priority backlog.

PR #2728, opened 2016-08-24
Revert to displaying unit abbreviations, and fix display of area units

RickMohr:
Connects #2711

### DIFF
--- a/opentreemap/treemap/lib/map_feature.py
+++ b/opentreemap/treemap/lib/map_feature.py
@@ -21,7 +21,7 @@ from treemap.models import Tree, MapFeature, User, Favorite
 
 from treemap.lib import format_benefits
 from treemap.lib.photo import context_dict_for_photo
-from treemap.units import get_display_value, get_units, get_unit_name
+from treemap.units import get_display_value, get_units, get_unit_abbreviation
 from treemap.util import leaf_models_of_class, to_object_name
 
 from stormwater.models import PolygonalMapFeature
@@ -315,8 +315,8 @@ def context_dict_for_resource(request, resource, **kwargs):
         __, display_area = get_display_value(instance,
                                              'greenInfrastructure', 'area',
                                              resource.calculate_area())
-        display_units = get_unit_name(get_units(instance,
-                                      'greenInfrastructure', 'area'))
+        display_units = get_unit_abbreviation(
+            get_units(instance, 'greenInfrastructure', 'area'))
         context['area'] = '{} {}'.format(display_area, display_units)
 
     return context
--- a/opentreemap/treemap/templatetags/form_extras.py
+++ b/opentreemap/treemap/templatetags/form_extras.py
@@ -21,7 +21,7 @@ from treemap.json_field import (is_json_field_reference,
                                 get_attr_from_json_field)
 from treemap.units import (get_digits_if_formattable, get_units_if_convertible,
                            is_convertible_or_formattable, format_value,
-                           get_unit_name)
+                           get_unit_abbreviation)
 
 register = template.Library()
 
@@ -386,7 +386,7 @@ class AbstractNode(template.Node):
             units = get_units_if_convertible(
                 model.instance, object_name, field_name)
             if units != '':
-                units = get_unit_name(units)
+                units = get_unit_abbreviation(units)
 
         if field_value is None:
             display_val = None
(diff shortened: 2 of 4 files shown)


opentreemap/treemap/lib/map_feature.py
```diff
@@ -315,8 +315,8 @@ def context_dict_for_resource(request, resource, **kwargs):
         __, display_area = get_display_value(instance,
                                              'greenInfrastructure', 'area',
                                              resource.calculate_area())
-        display_units = get_unit_name(get_units(instance,
-                                      'greenInfrastructure', 'area'))
+        display_units = get_unit_abbreviation(
```
RobinIsTheBird: I am unable to induce a display problem for area using develop. I'm using the chrome window resizer extension and resizing to a width of 768px, "A High-end Windows Phone". I'd say to keep the full-length unit name for area until/unless it becomes a problem.

opentreemap/treemap/templatetags/form_extras.py
```diff
@@ -386,7 +386,7 @@ def _field_value(model, field_name, data_type):
             units = get_units_if_convertible(
                 model.instance, object_name, field_name)
             if units != '':
-                units = get_unit_name(units)
+                units = get_unit_abbreviation(units)
```
RobinIsTheBird: The width only affects layout of the diameter/circumference, which is the only place where we have side-by-side form layout. Can we add an option for abbreviation or full, and default to full?

RickMohr: Using full unit names got introduced by accident to fix the problem of `sq_m` being displayed. If we want to consider showing full unit names instead of abbreviations we should open a larger issue, but we use abbreviations everywhere and that's been fine. This is really just a less-drastic fix for the problem of `sq_m` being displayed.

opentreemap/treemap/lib/map_feature.py
RickMohr: There's not a display problem, this is just changing to displaying unit abbreviations (which we do everywhere else).

maurizi: Well we _do_ use full unit names in consistently and purposefully [in the audit system](https://github.com/OpenTreeMap/otm-core/blob/d586ae00dcba28c1f11226c483be29bc5f0fced8/opentreemap/treemap/audit.py#L1318-L1342) for displaying items in the plot detail sidebar and on the edits pages.

So we haven't been using abbreviations _everywhere_...

That said, going back to abbreviations fixes the display issue with diameter/circumference, and it matches what we were doing before.  It would be nice to be consistent though, and use either full names or abbreviations everywhere on the public-facing portion of the site.

opentreemap/treemap/units.py
```diff
@@ -87,6 +87,11 @@ def convert_to_database_units(self):
     "sq_m": _("meters²")
 }
 
+_unit_abbreviations = {
+    "sq_ft": _("ft²"),
+    "sq_m": _("m²")
+}
```
RobinIsTheBird: I now understand that the intent is to deviate as little as possible from the production appearance, while maintaining correctness.

I feel that either all the abbreviations should be subject to translation, or none of them.

Googling for `translate "10 sq ft" into arabic` results in something like: 10 قدم مربع, except that the 10 appears after the arabic (but github won't let me do that). "10 ft²" same thing. "10 in" translates to في  , similarly with the 10 after the arabic.

The problem with text-width based layout in a localized page is that translations may be a different width, notably often wider. Of course we're optimizing for English, but what other languages are we supporting?

RickMohr:
@RobinIsTheBird, thanks for your comments. Fixed up.

opentreemap/treemap/units.py
```diff
@@ -107,6 +140,10 @@ def get_unit_name(abbrev):
     return _unit_names[abbrev]
 
 
+def get_unit_abbreviation(abbrev):
+    return _unit_abbreviations.get(abbrev, abbrev)
+
```
RobinIsTheBird: get(abbrev, abbrev) as opposed to [abbrev] can delay noticing that we missed a unit. Seems to me that if we raise an exception, then if someone adds a new unit, they'll find out really quick that they missed adding its abbrev.

Are we really worried about a new unit escaping into the wild without an abbrev?

RickMohr: Good catch, meant to fix that. Now fixed.

RobinIsTheBird:
:+1:

Issue #2721, opened 2016-08-22
3 forms fail to focus on first input (webpack fallout?)


RickMohr:
Likely caused by switch to webpack.
Exceptions occur on these lines:
https://github.com/OpenTreeMap/otm-core/blob/8de1867f0af9be3e6570c6a48d19cda4f6cfde45/opentreemap/treemap/templates/registration/password_reset_confirm.html#L30
https://github.com/OpenTreeMap/otm-core/blob/8de1867f0af9be3e6570c6a48d19cda4f6cfde45/opentreemap/treemap/templates/registration/password_reset_form.html#L24
https://github.com/OpenTreeMap/otm-core/blob/8de1867f0af9be3e6570c6a48d19cda4f6cfde45/opentreemap/treemap/templates/registration/registration_form.html#L57

PR #2730, opened 2016-08-25
Unify JS on pages containing a simple form (fixing JS errors)

RickMohr:
We missed updating JS on a few pages when we switched to webpack. So:
* Create `simpleForm.js` which just focuses on the first non-hidden `input` element.
* Use it on pages for login, registration, and password reset.

Note that `password_reset_form.html` had been adding the class `form-control` to its input element.
Since that doesn't change its appearance significantly I've taken the simpler route of using the shared `simpleForm.js`.

Connects #2721


maurizi:
Looks good to me

PR #2738, opened 2016-08-26
Allow 0 as a value for resource drainage area

RickMohr:
Changes to `validate_positive_nullable_float_field`:
* Add `zero_ok` argument
* DRY it out
* Add tests

Connects #2732


RobinIsTheBird:
:+1:

PR #2742, opened 2016-08-29
Add meta species lookup table for Tampa migration

jwalgran:
I added a static lookup table in the Tampa migration code to handle species that do not have an i-Tree code set in the species fixture. I opted to handle these 7 edge case species with a literal lookup rather than make the non-trivial changes to the migrator to properly handle the one-to-many relationship between species rows and resource rows. The migrator is reaching the end of its life and the cost of the changes does not justify the relatively small gain in general
correctness.

I created the table by running this query against the otm-legacy database:

```sql
SELECT lower(scientific_name), meta_species
FROM treemap_species s
JOIN treemap_species_resource sr ON s.id = sr.species_id
JOIN treemap_resource r ON sr.resource_id = r.id
WHERE itree_code = '';
```
---

##### Testing

After running a migration using this branch, trees that have a non-zero diameter and any one of the 7 affected species will have properly calculated eco benefits. 

yunnanensis
dactylifera
chrysotricha
rupicola
impetiginosa
unknown
sylvestris

---

Connects to https://github.com/OpenTreeMap/otm-clients/issues/276

### DIFF
--- a/opentreemap/otm1_migrator/migration_rules/tampa.py
+++ b/opentreemap/otm1_migrator/migration_rules/tampa.py
@@ -59,9 +59,9 @@ meta_species = {
 
 
 def create_override(species_obj, species_dict):
-    sci_name = species_dict['fields']['scientific_name'].lower()
     itree_code = species_dict['fields'].get('itree_code', None)
     if not itree_code:
+        sci_name = species_dict['fields'].get('scientific_name', '').lower()
         print('No itree_code for "%d: %s"' % (species_dict['pk'], sci_name))
         itree_code = meta_species.get(sci_name, '')
         print('Looked up meta species "%s"' % itree_code)
--- a/opentreemap/otm1_migrator/migration_rules/tampa.py
+++ b/opentreemap/otm1_migrator/migration_rules/tampa.py
@@ -28,12 +28,47 @@ MIGRATION_RULES['boundary']['presave_actions'] = (MIGRATION_RULES['boundary']
                                                   .get('presave_actions', [])
                                                   + [set_boundary_fields])
 
+# At the time of migration, 7 of the Tampa species did not have an i-Tree code
+# defined in the species fixture. otm-legacy has a one-to-many relationship
+# between the `treemap_species` table and the `treemap_resources` table via
+# `treemap_species_resource` and the i-Tree code used by the eco benefit
+# calculations is read from the `meta_species` column of the
+# `treemap_resources` table.
+#
+# Rather than spend time updating the migrator to
+# handle this edge case, we have opted to add a literal dictionary lookup for
+# these species so we can properly calculate eco benefits for these species.
+#
+# The lookup table was created by running this query against the otm-legacy
+# database:
+#
+# SELECT lower(scientific_name), meta_species
+# FROM treemap_species s
+# JOIN treemap_species_resource sr ON s.id = sr.species_id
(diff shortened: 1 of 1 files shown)


opentreemap/otm1_migrator/migration_rules/tampa.py
```diff
@@ -28,12 +28,47 @@ def set_boundary_fields(boundary_obj, boundary_dict):
...
+    'tabebuia chrysotricha': 'BDS OTHER',
+    'phoenix rupicola': 'PEM OTHER',
+    'tabebuia impetiginosa': 'BDM OTHER',
+    'unknown unknown': 'BDM OTHER',
+    'phoenix sylvestris': 'PEM OTHER'
+}
+
 
 def create_override(species_obj, species_dict):
+    sci_name = species_dict['fields']['scientific_name'].lower()
```
RickMohr: This would raise an exception if `scientific_name` were missing. Perhaps safer (and clearer) to move it inside the `if` since it's only used there.

RickMohr:
Otherwise looks good. Let me know if you think I should run a Tampa migration to test it.

opentreemap/otm1_migrator/migration_rules/tampa.py
jwalgran: `scientific_name` is not nullable, so it is safe to assume that it will be there, but it is not a bad idea to make things more robust by using `.get`. Moving the declaration of the variable closer to its usage is also a good idea.

```
otmtampa=# \d treemap_species
                                       Table "public.treemap_species"
       Column       |          Type          |                          Modifiers
--------------------+------------------------+--------------------------------------------------------------
...
 scientific_name    | character varying(255) | not null
...

**Comment by jwalgran** [comment]
I don't think you need to run your own migration. I used this branch to import into `tampa8` on my VM. Here are my findings:

There are 580 species in the fixture, and they were all loaded.

```
$ node
> var s = require('./species.json')
undefined
> s.length
580
```

```sql
SELECT COUNT(*) FROM treemap_species
WHERE instance_id = (
  SELECT id FROM treemap_instance
  WHERE url_name = 'tampa8');

 count
-------
   580
(1 row)
```

All 580 have an override.

```sql
SELECT COUNT(*)
FROM treemap_species s
JOIN treemap_itreecodeoverride o
ON s.id = o.instance_species_id
WHERE instance_id = (
  SELECT id
  FROM treemap_instance
  WHERE url_name = 'tampa8');

 count
-------
   580
(1 row)
```

All of the species override rows have an `itree_code`.

```sql
SELECT genus, species, itree_code
FROM treemap_species s
JOIN treemap_itreecodeoverride o ON s.id = o.instance_species_id
WHERE instance_id = (
  SELECT id
  FROM treemap_instance
  WHERE url_name = 'tampa8')
AND (itree_code IS null OR itree_code = '');

 genus | species | itree_code
-------+---------+------------
(0 rows)
```

RickMohr:
Looks great!

opentreemap/otm1_migrator/migration_rules/tampa.py
jwalgran: I implemented these changes in 4268053 and ran another migration into a new, empty instance to confirm that the species are still imported correctly.

RickMohr:
+1 Thanks for the careful work.

Issue #2747, opened 2016-08-30
The develop and release/2.11.0 branches have model changes not reflected in a migration


jwalgran:
Running `./scripts/manage.sh migrate`

```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

hectcastro:
Tried to reproduce this on staging, but wasn't able to:

```bash
+ envdir /etc/otm.d/env /opt/app/core/opentreemap/manage.py migrate
Operations to perform:
  Synchronize unmigrated apps: django_hstore, gis, webpack_loader, staticfiles, messages, geocode, django_js_reverse, humanize, url_tools
  Apply all migrations: map_features, threadedcomments, appevents, django_comments, sessions, admin, treemap, djcelery, importer, auth, sites, contenttypes, api, otm1_migrator, stormwater, registration, modeling, super_admin, cloud_management, exporter, otm_comments
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying stormwater.0008_benefits-calc-cache-flush... OK
  Applying stormwater.0009_drainage_area_imperial_units... OK
```

Not sure where the inconsistency is at.

jwalgran:
I originally saw the migration message when looking through the console output of a failed jenkins build.

The commit that added these columns is missing `blank=True` in the migration
https://github.com/OpenTreeMap/otm-core/commit/c805dc8385dfe595be21379b5a8858c24bca8eb3#diff-5a028dba261197e5ca02f3c9d82216fbR17
but has `blank=True` in the model
https://github.com/OpenTreeMap/otm-core/commit/c805dc8385dfe595be21379b5a8858c24bca8eb3#diff-7d1df61bdeceb7fd562de2706895c629R61

@RobinIsTheBird may have some more insight.

jwalgran:
The console output of build 40 of the `otm-release` Jenkins job contains the migration message.

RobinIsTheBird:
The PR was merged. Is there more? Can this issue be moved to In Review or OTM Release?

jwalgran:
Build 41 of the `otm-release` Jenkins job successfully applied this migration. I am going to move this to "tested on staging" so that we include it as part of our review next week.

PR #2748, opened 2016-08-30
Add missing stormwater migration

jwalgran:
Note: This is targeted at the `release/2.11.0` branch.

Running `./scripts/manage.sh migrate` in development was printing a "Your models have changes that are not yet reflected in a migration..." message.

I ran `./scripts/manage.sh makemigrations` to generate the missing migration. The two fields were added in stormwater migration 0006. The migration is missing the `blank=True` keyword argument, but the changes to `models.py` in c805dc8 include the `blank=True` keyword argument. I am not sure how this mismatch occurred.

---

##### Testing

Migrations should be applied without any warning messages.

---

Connects to #2747

### DIFF
--- a/opentreemap/stormwater/migrations/0010_stormwater_blank_true.py
+++ b/opentreemap/stormwater/migrations/0010_stormwater_blank_true.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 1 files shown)


RobinIsTheBird:
I must have added blank=True to the model after I created the migration.

RobinIsTheBird:
:+1:

Issue #2754, opened 2016-08-31
Diameter warning limits space for diameter field


dboyer:
Tested in staging on Chrome. See attached screenshot. It's hard to type in a diameter since the warning message limits the space for the diameter field to less than one character.
![otm_detail_page_error](https://cloud.githubusercontent.com/assets/516764/18135694/bdfa9e68-6f70-11e6-9ba2-9bf2f5f42860.png)

RobinIsTheBird:
@lederer How do you suggest we display the warning? And should we change the position of all the warnings?

RobinIsTheBird:
@lederer , I know we talked about putting the message below the label, but it turns out to be easier to make it a separate row with a colspan="2" cell. How does this look to you?

![treedetailerrormessage](https://cloud.githubusercontent.com/assets/1473263/18281148/c9de90a0-7429-11e6-9a69-84dc13130138.png)

lederer:
+1 I'll take it.

PR #2760, opened 2016-09-02
Message makes tree diameter too narrow to use

RobinIsTheBird:
@lederer recommended we display the message beneath the field label because there is plenty of room there.

Also, the message should be in red, like the message in the add-a-tree sidebar wizard.

Also, the label should not be vertical-align middle. Its baseline should be even with the baseline of the text in the second table cell, whether it's in edit mode or not.

This needed to be done for all rows in the detail pages. Otherwise, the presence of a message on the right on any row would make the tree diameter too narrow to use.

Changes:

opentreemap/treemap/js/src/lib/
------------------------------------------
* editableForm.js
* inlineEditForm.js
* simpleEditForm.js

Manipulate a [data-mode] element to be either edit or display.

opentreemap/treemap/templates/treemap/partials/
------------------------------------------------------------------
* map_feature_detail_base.html

Set #map-feature-form to have a data-mode attribute

* plot_detail.html
* resource_detail.html

Add a can-inline-edit class to the tables to distinguish them from ordinary tables for styling purposes. Specifically, it distinguishes them from the diameter-circumference control, which uses table layout.

opentreemap/treemap/templates/treemap/field/
--------------------------------------------------------------
* attrs.html
* tr.html
* diameter_tr.html
* species_tr.html

Add a css_class variable to attrs.html to use in the class attribute, as distinct from the class variable already in use for the data-class attribute.

The *tr.html templates set css_class to `alert alert-danger` on the error elements to make them styled with red.

assets/css/sass/partials/pages/_treedetails.scss
------------------------------------------------

Remove the diameter-row style.
Style the first td in a row differently depending on whether
* the table is in edit mode
* the table contents can be edited inline
* the row the td is in is itself editable

Attention to whether the table contents can be edited allows the style to apply just to the desired table's cells, not to cells in child tables such as the diameter-circumference editor.

--

Connects to #2754

### DIFF
--- a/assets/css/sass/partials/pages/_treedetails.scss
+++ b/assets/css/sass/partials/pages/_treedetails.scss
@@ -43,17 +43,24 @@
 #map-feature-form {
     margin-top: 0;
 
-    table {
+    table.can-inline-edit {
         > tbody {
             > tr {
                 > td:first-child {
                     width: 45%;
-                    vertical-align: middle;
+                    > .alert {
+                        margin-top: 8px;
+                    }
                 }
-
-                &.diameter-row {
-                    td:first-child {
-                        vertical-align: top;
+            }
+        }
+    }
+    &[data-mode="edit"] {
+        table.can-inline-edit {
+            > tbody {
+                > tr.editable {
+                    > td:first-child {
+                        padding-top: 16px;
                     }
                 }
             }
--- a/opentreemap/treemap/js/src/lib/editableForm.js
+++ b/opentreemap/treemap/js/src/lib/editableForm.js
@@ -142,7 +142,9 @@ exports.init = function(options) {
             typeaheadToDisplayValues();
         },
 
-        hideAndShowElements = function (fields, actions, action) {
+        hideAndShowElements = function (mode, fields, actions, action) {
+            var $modeEl = $(fields).parents('[data-mode]');
+            $modeEl.attr('data-mode', mode);
             if (_.contains(actions, action)) {
                 $(fields).show();
             } else {
--- a/opentreemap/treemap/js/src/lib/inlineEditForm.js
+++ b/opentreemap/treemap/js/src/lib/inlineEditForm.js
@@ -277,14 +277,14 @@ exports.init = function(options) {
     actionStream.plug(saveOkStream.map('save:ok'));
     actionStream.plug(responseErrorStream.map('save:error'));
     actionStream.plug(modeChangeStream);
-    actionStream.onValue(hideAndShowElements, editFields, eventsLandingInEditMode);
-    actionStream.onValue(hideAndShowElements, displayFields, eventsLandingInDisplayMode);
-    actionStream.onValue(editForm.hideAndShowElements, validationFields, ['save:error']);
+    actionStream.onValue(hideAndShowElements, 'edit', editFields, eventsLandingInEditMode);
+    actionStream.onValue(hideAndShowElements, 'display', displayFields, eventsLandingInDisplayMode);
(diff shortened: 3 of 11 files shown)


opentreemap/exporter/tests.py
```diff
@@ -287,39 +287,39 @@ def test_export_users_json_keep_info_private(self):
...
-                             'last_edit_model_id': str(self.plot.pk),
-                             'last_edit_instance_id': str(self.instance.pk),
-                             'last_edit_user_id': str(self.user1.pk),
-                             'email': 'genly@example.com',
-                             'email_hash': self.user2.email_hash,
-                             'last_edit_model': 'Tree',
-                             'last_edit_model_id': str(self.tree.pk),
-                             'last_edit_instance_id': str(self.instance.pk),
-                             'last_edit_user_id': str(self.user2.pk)})
+#    def test_export_users_json_make_info_public(self):
```
jwalgran: The assert in this test does work, but has what looks like some copy/paste or merge cruft.

Python will let you redefine keys when declaring a literal map. The last value wins.

```python
In [1]: d = {'a':1 , 'b': 2, 'a': 3}

In [2]: d
Out[2]: {'a': 3, 'b': 2}
```

The literal hash can be fixed to remove redefined keys, and the test still passes.

```diff
diff --git a/opentreemap/exporter/tests.py b/opentreemap/exporter/tests.py
index c08c3c8..ab9235b 100644
--- a/opentreemap/exporter/tests.py
+++ b/opentreemap/exporter/tests.py
@@ -310,16 +310,12 @@ class UserExportsTest(UserExportsTestCase):
                              'created': str(self.user1.created)})
 
         self.assertUserJSON(user2data,
-                            {'last_edit_model': 'Plot',
-                             'last_edit_model_id': str(self.plot.pk),
-                             'last_edit_instance_id': str(self.instance.pk),
-                             'last_edit_user_id': str(self.user1.pk),
-                             'email': 'genly@example.com',
-                             'email_hash': self.user2.email_hash,
-                             'last_edit_model': 'Tree',
+                            {'last_edit_model': 'Tree',
                              'last_edit_model_id': str(self.tree.pk),
                              'last_edit_instance_id': str(self.instance.pk),
-                             'last_edit_user_id': str(self.user2.pk)})
+                             'last_edit_user_id': str(self.user2.pk),
+                             'email': 'genly@example.com',
+                             'email_hash': self.user2.email_hash})
 
     def test_min_edit_date(self):
         last_week = now() - datetime.timedelta(days=7)
```

I tried both commenting out and fixing this test and it did not resolve the hard to pin down memory pressure issues we are currently running into.

I would prefer that we not include this commit in the PR.

jwalgran:
The Jenkins build failure is caused by the fact that we can no longer mix and match the `develop` branches of repos with the `release/2.11.0` branches. The reviewer can run the tests with all of the `release/2.11.0` branches checked out to verify that there are no regressions.

opentreemap/exporter/tests.py
RobinIsTheBird: Not saying the test had been failing, but that it was passing in spite of errors in the test.

If the repetition was caused by merge kruft, it was caused two years ago. Simply an incorrect test, whether it fails or passes.

Meanwhile, I can't get through travis without this change.

RobinIsTheBird:
Submitting this PR for code review in spite of ci failures, because the failures are unrelated to my changes.

Commenting a test out of exporter.tests gets past travis, but that doesn't belong in this PR, so I took it out.

Jenkins fails trying to provision because of a statsd incompatibility between develop and release. I don't understand why release is pulling in that change, but at this point it's all beyond me.

Apart from transient memory issues, also apparently unrelated to my changes, tests pass on my local vm.

Issue #2761, opened 2016-09-03
exporter/tests.py fails some versions of flake8


RobinIsTheBird:
On travis:
```
4.69s$ flake8 --exclude migrations,opentreemap/settings/local_settings.py,*.pyc opentreemap
opentreemap/exporter/tests.py:313:1: F999 dictionary key u'last_edit_model' repeated with different values
opentreemap/exporter/tests.py:314:1: F999 dictionary key u'last_edit_model_id' repeated with different values
opentreemap/exporter/tests.py:315:1: F999 dictionary key u'last_edit_instance_id' repeated with different values
opentreemap/exporter/tests.py:316:1: F999 dictionary key u'last_edit_user_id' repeated with different values
opentreemap/exporter/tests.py:319:1: F999 dictionary key u'last_edit_model' repeated with different values
opentreemap/exporter/tests.py:320:1: F999 dictionary key u'last_edit_model_id' repeated with different values
opentreemap/exporter/tests.py:321:1: F999 dictionary key u'last_edit_instance_id' repeated with different values
opentreemap/exporter/tests.py:322:1: F999 dictionary key u'last_edit_user_id' repeated with different values
```

The complaints are correct. Here's the code at fault:
```
        self.assertUserJSON(user2data,
                            {'last_edit_model': 'Plot',
                             'last_edit_model_id': str(self.plot.pk),
                             'last_edit_instance_id': str(self.instance.pk),
                             'last_edit_user_id': str(self.user1.pk),
                             'email': 'genly@example.com',
                             'email_hash': self.user2.email_hash,
                             'last_edit_model': 'Tree',
                             'last_edit_model_id': str(self.tree.pk),
                             'last_edit_instance_id': str(self.instance.pk),
                             'last_edit_user_id': str(self.user2.pk)})
```

It was last edited in 2014 by Steve Lamb. I don't know how anybody ever thought this test actually worked.

Commenting out the test until this bug gets scheduled.

RobinIsTheBird:
To clarify, having a test _pass_ is not the same thing as having the test _work_.  A test _works_ if it actually tests what it sets out to test, whether it passes or fails. Given that the code at fault adds keys to a dict and immediately replaces them, we can't even really say what the intent of the test was. In that sense, it is not a working test even though it passes.

jwalgran:
I pulled these lines from the build log of a Travis job that failed:

```
[...]

PR #2762, opened 2016-09-03
Remove duplicate keys from object literal in test

jwalgran:
The object literal confusingly defines keys twice. Based on the content of the `setUp` function and the assert statement

```python
    self.assertUserJSON(user2data, ...
```

it is clear that the intention was to assert `user2` values.

Connects to #2761


RobinIsTheBird:
I know you didn't assign me, @jwalgran , but :+1:  anyway.

PR #2765, opened 2016-09-06
Tree detail error message vs. diameter field

RobinIsTheBird:
When there was a tree detail error, it was displayed in a new third table column. This caused the
diameter/circumference edit widget to get so narrow you couldn't see the values.

Changes: 1. Show the message in a new <tr>. 2. Show/hide the <tr>. 3. Show the message in a
colspan="2" <td> in the <tr>,
  so it fills the width of the row. 4. Show the message in a div inside the <td colspan="2">
  to get the styling right. 5. Style it.

The styling problem is that bootstrap.css orders
.table > tbody > tr > td later than .alert-danger, resulting in a td-looking top border, instead of 
4 alert-danger borders, which looks awful.

The alternative would have been to specify the border-top in the td[colspan="2"] style, but then we
have to copy-paste from bootstrap, which has maintainability problems.

--

Connects to #2754

### DIFF
--- a/opentreemap/treemap/templates/treemap/field/diameter_tr.html
+++ b/opentreemap/treemap/templates/treemap/field/diameter_tr.html
@@ -14,8 +14,14 @@
         </div>
         {% include "treemap/partials/diameter_calculator.html" %}
       </td>
-    <td style="display:none;"
-        {% include "treemap/field/attrs.html" with class="error" %}></td>
   {% endif %}
   </tr>
+  {% if field.is_editable %}
+  <tr class="error" style="display:none">
+    <td colspan="2">
+      <div class="alert alert-danger text-danger"
+           {% include "treemap/field/attrs.html" with class="error" %}></div>
+    </td>
+  </tr>
+  {% endif %}
 {% endif %}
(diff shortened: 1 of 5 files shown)


RobinIsTheBird:
For testing purposes:
The affected pages are tree/plot details, resource details, and LA solution details.

None of the LA solution details are editable, and Rain Barrel is not editable, so the change has no effect.

Rain Garden and Bioswale drainage area is editable, but doesn't produce an error message, so the change has unknown effect there.

opentreemap/treemap/templates/treemap/field/diameter_tr.html
```diff
@@ -14,8 +14,14 @@
...
         {% include "treemap/partials/diameter_calculator.html" %}
       </td>
-    <td style="display:none;"
-        {% include "treemap/field/attrs.html" with class="error" %}></td>
   {% endif %}
   </tr>
+  {% if field.is_editable %}
+  <tr class="error" style="display:none">
+    <td colspan="2">
+      <div class="alert alert-danger text-danger"
```
RickMohr: In my Chrome dev tools, `text-danger` adds no attributes.

RickMohr:
Looks good. Works with Bioswale's Adjacent Drainage Area too (you can cause an error by entering a negative number or a non-number).

+1 after removing `text-danger`

Issue #2734, opened 2016-08-25
Search by Owner Original ID


RickMohr:
A new advanced search option on the "General" tab.

Also, display planting site ID instead of tree ID on plot detail page.

RobinIsTheBird:
Since we had the meeting, we're blocked waiting on Scott's wireframe, and then we need to re-point this issue.

RobinIsTheBird:
Based on the outcome of the trees-team issue 78, the queued issue is now about searching by Owner Original ID.

That issue is spawning a number of new issues, which must be deferred to future sprints.

For this issue, here is what we are going to do, pending approval by @lederer , @RickMohr , and @jwalgran .
1. Make search by Owner Original ID configurable on the management / field configuration / search fields page.
2. Add the option to Advanced Search.
3. Operate as a filter - making non-matching trees transparent.
4. If it resolves to a single tree, select it and zoom, similar to search by location.

If it resolves to multiple trees, it would be nice to zoom to a level that shows just those results, but that is out of scope for this issue.

If a planting site has an Owner Original ID, it presently appears on the detail page. If this feature gets popular, we might consider also showing it in the sidebar detail panel and/or the popup, but that is out of the scope of this issue.

RobinIsTheBird:
The user story, as I understand it from @hadleystein1 , goes something like this:

As a _city manager (or urban forester?)_, I want to _locate a specific tree_, so that I _can tell the crew exactly which tree to service_.

According to Hadley, the manager has a list of trees based on the owner original ID, not from zooming on the open treemap map and selecting.

jwalgran:
This sounds like a workable solution.

I have a question about zooming behavior. If I filter by Owner Original ID and get 1 result, the map zooms and selects. If I filter by a custom field and get 1 result, what happens? Do we also zoom and select?

hadleystein1:
Right now, if you filter by a custom field and get one result that result is highlighted and all other results are grayed out. The single result is not selected and the area is not zoomed in on further. See example below.

![screen shot 2016-08-30 at 9 24 01 pm](https://cloud.githubusercontent.com/assets/15219245/18112767/2668a8ba-6ef8-11e6-9a7d-5f828f6a574e.png)

Is that what you are getting at @jwalgran?

lederer:
I think we should automatically _zoom_ and _select_ any singleton result. This would apply to any search, over any field(s), that results in a single match.

[...]

PR #2766, opened 2016-09-06
Add support for searching on the 'owner_orig_id' field.

maurizi:
Also renamed the `owner_orig_id` field to be "Custom ID", and updates its name in the importer template.

With this change in place, "Custom ID" should be a search field option in the search fields management page under the planting site section, and "Missing Planting Site Custom ID" should be an option under the "Missing Data" section.

Connects to #2734

### DIFF
--- a/opentreemap/importer/fields.py
+++ b/opentreemap/importer/fields.py
@@ -95,7 +95,7 @@ class trees(object):
     # READ_ONLY = 'read only'
 
     OPENTREEMAP_PLOT_ID = 'planting site id'
-    EXTERNAL_ID_NUMBER = 'owner original id'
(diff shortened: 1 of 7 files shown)


RickMohr:
The plot detail page still says "Owner Original ID" instead of "Custom ID"

RickMohr:
I think searching by Custom ID should be enabled by default. 

It will require a non-trivial migration, but will be more easily discoverable. 

@dboyer, thoughts?

RickMohr:
Otherwise nice and simple and works great.

dboyer:
I'm fine with enabling it by default.

RickMohr:
Entered enabling by default as #2769

RickMohr:
+1

Issue #2736, opened 2016-08-26
Embed query string


RobinIsTheBird:
1. Handle an embed query string parameter in the urls for the map page, the detail panel, and the popup.
2. On the map page, forward the embed query string to the detail and popup urls.
3. Strip header and footer from the map page for embed.
4. Modify the following links to redirect to opentreemap.org for the same treemap, in a separate tab or window:
   - popup View All Details
   - popup Edit Details
   - detail Full Details
   - detail Quick Edit
   - man map Export link
5. Add the html snippet for the iframe in a template, parameterized by size in the context.
6. Demonstrate the appearance of the iframe on a webpage for a demo website.

dboyer:
Just to confirm - the "detail panel" means the right sidebar, correct? Not the detail page? I see from 4 that the view all details and full details should open in a new page (which is correct) so wanted to confirm what the detail panel meant.

RobinIsTheBird:
The links in the sidebar detail panel need to be modified to pop up a new tab or window. The url in the new tab or window would refer to the detail page, in view or edit mode. If authentication is required, the new tab|window will display a prompt to login.

Sorry for the confusion.

dboyer:
Thanks for clarifying. Sounds great!

RobinIsTheBird:
@dboyer , @lederer , for now, I've got embed mode taking away the Quick Edit button in the sidebar details panel when a tree/plot is selected.

This is easier than changing the name to just "Edit" and opening a new tab or window.

Is this acceptable for now?

PR #2768, opened 2016-09-07
Honor embed query string parameter

RobinIsTheBird:
On map page:
* Add embed context variable.
* Prune header and footer HTML if embed.
* Add an embed class to wrapper.
* Adjust data for Add a Tree or Resource buttons.
* Fix up button anchors
* For embed, don't include the quick edit button.

Style:
* Style adjustments for .wrapper.embed.
* Eliminated ambiguous "map" css class.

JavaScript:
* Save the embed query string parameter in the mapPage.
* If embed is true, fix up the add button's href,
  and let the browser do its thing.
* Change Add to open a new tab or window,
  and preserve zoom.
* Change details links to open a new tab or window.

Add a test to make sure we still get the page when
?embed= is in the query string.

--

Connects to #2736

### DIFF
--- a/opentreemap/treemap/templates/instance_base.html
+++ b/opentreemap/treemap/templates/instance_base.html
@@ -137,19 +137,22 @@ ga('set', 'page', pathname);
     <div class="half">
       <div style="display: inline;" id="tree-and-planting-site-counts">
       </div>
-      {% block subhead_exports %}
-        {% if request.instance|export_enabled_for:request.user %}
-        <a href="javascript:;" class="btn btn-primary btn-xs exportBtn"
-           data-export-start-url="{% url 'begin_export' instance_url_name=request.instance.url_name model='tree' %}">
-          <i class="icon-export"></i> {% trans "Export Search Results" %}
-        </a>
-        {% endif %}
-      {% endblock subhead_exports %}
+      {% if not embed %}
+        {% block subhead_exports %}
+          {% if request.instance|export_enabled_for:request.user %}
+          <a href="javascript:;" class="btn btn-primary btn-xs exportBtn"
+             data-export-start-url="{% url 'begin_export' instance_url_name=request.instance.url_name model='tree' %}">
+            <i class="icon-export"></i> {% trans "Export Search Results" %}
+          </a>
+          {% endif %}
+        {% endblock subhead_exports %}
+      {% endif %}
     </div>
     <div class="half">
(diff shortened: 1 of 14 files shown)


RobinIsTheBird:
Note, travis is still failing on flake8 of exporter/tests, nothing to do with this change. Everything else passes.

opentreemap/treemap/templates/instance_base.html
```diff
@@ -158,9 +161,14 @@
            disabled="disabled"><i class="icon-plus"></i> {% trans "Add a" %} {{ term.Resource.singular }}</a>
         {% endif %}
       <a class="btn btn-primary addBtn"
-         data-action='addtree'
+         {% if embed %}
+         target="_blank"
+         {% else %}
          data-feature="add_plot"
```
jwalgran: Is it important that we remove these data attributes here if we are in embed mode? The "Add Resource" button only has an if embed/endif around the `target="_blank"`. We may want to keep them the same for consistency.

jwalgran:
This is working well. Nice job.

I did find a pair of related edge cases. If the treemap is suspended, the suspension page needs to hide the header and add a `target="_blank"` attribute to the "login" link if `embed=true`

screen shot 2016-09-07 at 11 20 05 am

If the instance is not publicly accessible, the iframe gets redirected to the login page. We should probably redirect to `/not-available` instead if `embed=true`

screen shot 2016-09-07 at 11 17 09 am

opentreemap/treemap/templates/instance_base.html
RobinIsTheBird: `data-action` is still present, below, line 171.

`data-feature` is there for controlling add tree mode, which embed disables, and hiding the add tree control if save returns a result with `enabled` set to `False`. It seems unnecessary at best, and removing it helps assure that none of these actions will be triggered.

`data-event-category` and `data-event-action` trigger google analytics, which will be triggered when the new tab or window opens in `add` mode. We probably do not want to trigger them twice.

For reasons unknown to me, "Add a Resource" doesn't have these attributes to begin with.

Exactly what change are you proposing?

jwalgran: No change needed. I was just verifying that these attributes were being hidden by design.

RobinIsTheBird:
See also [addons #1238](https://github.com/OpenTreeMap/otm-addons/pull/1238), for treatment of suspended treemaps. This PR doesn't depend on the addons one, but the addons one depends on this.

jwalgran:
👍

Issue #2745, opened 2016-08-30
Searches with single result should select that result


RobinIsTheBird:
Any search with a single result — over any fields — selects that result.

jwalgran:
Right now, the results of a filter are returned as rendered HTML (from `/{instance}/benefit/search/`). Deciding how to add the additional information we need to show a selected feature be a design challenge. One possible implementation would be to conditionally add a data attribute to an element in the returned markup, something like `data-matched-feature-id`, and then have the handler that patches the markup into the DOM also trigger the selection process.

PR #2776, opened 2016-09-12
Select single search result, and show empty planting site count

RickMohr:
* Show count of empty planting sites rather than total planting sites
* Add helper instance properties `map_feature_classes` and `resource_classes`
* If search returns a single result, select that result

Connects #2772 
Connects #2745

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -511,6 +511,18 @@ class Instance(models.Model):
 
         add_default_permissions(self, models=classes)
 
+    @property
+    def map_feature_classes(self):
+        from treemap.models import MapFeature
+        classes = {MapFeature.get_subclass(m)
+                   for m in self.map_feature_types}
+        return classes
+
+    @property
+    def resource_classes(self):
+        from treemap.models import Plot
+        return self.map_feature_classes - {Plot}
+
     def update_geo_rev(self):
         self.update_revs('geo_rev')
 
--- a/opentreemap/treemap/views/tree.py
+++ b/opentreemap/treemap/views/tree.py
@@ -7,7 +7,7 @@ import hashlib
 
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext as _, ungettext
 from django.shortcuts import get_object_or_404
 from django.db import transaction
 from django.http import HttpResponseRedirect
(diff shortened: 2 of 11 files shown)


opentreemap/treemap/views/tree.py
```diff
@@ -90,16 +90,59 @@ def search_tree_benefits(request, instance):
...
+    n_trees = basis['plot']['n_total']
+    n_plots = basis['plot']['n_plots']
+    n_empty_plots = n_plots - n_trees
+    n_resources = 0
+
+    tree_count_label = _('tree') if n_trees == 1 else _('trees')
+    tree_count_label += ','
+    empty_plot_count_label = (
+        _('empty planting site') if n_empty_plots == 1 else
+        _('empty planting sites'))
```
maurizi: Note that the Django docs recommend not comparing against 1 for pluralization, but using the support built-in to `ungettext` for pluralization (to support languages with different suffixes for 2, 3, etc.) https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#pluralization

I think that would make this code looks like this:
```python
tree_count_label = ungettext('tree', 'trees', n_trees)
empty_plot_count_label = _('empty planting site', 'empty planting sites', n_empty_plots)
```

RickMohr: Thanks for the tip @maurizi. Since `_` is aliased to `ugettext` not `ungettext` we have to say instead e.g.
```python
tree_count_label = ungettext('tree', 'trees', n_trees)
```

RickMohr:
Assigning to @RobinIsTheBird at her request

opentreemap/treemap/views/tree.py
```diff
@@ -90,16 +90,57 @@ def search_tree_benefits(request, instance):
...
+    else:
+        if n_plots == 1:
+            qs = filter.get_objects(Plot)
+        else:  # n_resources == 1
+            for Resource in instance.resource_classes:
+                qs = filter.get_objects(Resource)
+                if qs.count() == 1:
+                    break
+        if qs.count() != 1:
+            raise Exception('Failed to retrieve single resource')
```
RobinIsTheBird: Is this unlikely enough that omitting this test and just letting it blow up on the next line where it indexes qs should be sufficient?

opentreemap/treemap/instance.py
```diff
@@ -511,6 +511,18 @@ def add_map_feature_types(self, types):
...
+        from treemap.models import MapFeature
+        classes = {MapFeature.get_subclass(m)
+                   for m in self.map_feature_types}
+        return classes
+
+    @property
+    def resource_classes(self):
+        from treemap.models import Plot
+        return self.map_feature_classes - {Plot}
+
```
RobinIsTheBird: :+1:  on these properties. They sure clean up a lot of kruft.

RobinIsTheBird:
Not sure whether my approval generated a notification to that effect, so to be on the safe side, :+1:

opentreemap/treemap/views/tree.py
RickMohr: Good point, fixed.

PR #2778, opened 2016-09-13
Change Owner Orig Id to Custom Id in role management

RobinIsTheBird:
In treemap.audit, change FieldPermission.display_field_name to
use the field verbose name if it has one.

--

Connects to OpenTreeMap/otm-addons#1239


RickMohr:
+1

PR #2779, opened 2016-09-15
Isolate UDF choice renaming

jwalgran:
UDF choices were being renamed across instances.

---

##### Testing

If you revert the change to `opentreemap/treemap/udf.py` then run `./scripts/manage.sh test treemap.tests.test_udfs.ScalarUDFInstanceIsolationTest` you should see an assertion fail toward the bottom of the test case.

You can also interactively reproduce the bug after reverting the change to `opentreemap/treemap/udf.py` 

* Create two instances
* Add a single choice field with the same name and same choices to both instances
* Add a tree to both instances and set the same choice value on both trees.
* Rename the choice on one instance, then load the tree detail page on the unchanged instance.

### DIFF
--- a/opentreemap/treemap/tests/test_udfs.py
+++ b/opentreemap/treemap/tests/test_udfs.py
@@ -728,6 +728,85 @@ class UDFDefTest(OTMTestCase):
                 name='__contains')
 
 
+class ScalarUDFInstanceIsolationTest(OTMTestCase):
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


RickMohr:
+1 tested. Great attention to detail in discovering the problem!

Releasing next sprint may suffice instead of a hotfix. While the potential harm is serious, this bug has been there since Adam wrote the code and takes a rare combination of circumstances to exercise.

Issue #2737, opened 2016-08-26
Responsive design glitch


RobinIsTheBird:
See the gray stripe beneath the Add a Tree, Add a Mitigator bar. Happens between 768 and 990 pixel widths.

![responsiveglitch](https://cloud.githubusercontent.com/assets/1473263/18012789/07516916-6b89-11e6-955c-866a902f325d.png)

RobinIsTheBird:
Also happens at 600px with the `embed` query param.

RobinIsTheBird:
Related - not sure if it belongs in a separate issue, is a problem at 800px depending on # of trees and terminology customization. Now that the planting site count is called "empty planting sites", the problem is even easier to provoke, even with the `embed` query param, which hides the `Export Search Results` button.

![otm 800w statusline layout issue](https://cloud.githubusercontent.com/assets/1473263/18601578/a24ed99a-7c31-11e6-902b-22bdb3fdf251.png)

PR #2787, opened 2016-09-20
Refactor stats bar to be more responsive on small-medium screen sizes

maurizi:
-  Fixes responsive styling issue with gray bar on screens smaller than 990px
 -  Combines add tree and add resource buttons into a single button group

screenshot

With these changes, it should not be possible for items to overlap or get pushed down in the header, no matter the screen size or the length of the "resources" terminology.

Connects to #2737

### DIFF
--- a/assets/css/sass/partials/_subheader.scss
+++ b/assets/css/sass/partials/_subheader.scss
@@ -187,32 +187,32 @@
         span {
             font-weight: bold;
         }
-
-        .half{
-            display: inline;
-            width: inherit;
+        .stats-list {
+            display: block;
+            width: 100%;
             padding: 0px;
 
-            label{
-                float: right;
+            .exportBtn {
+                margin-left: 10px;
+                margin-top: -12px;
+                width: 135px;
+                display: inline-block;
+            }
+            #tree-and-planting-site-counts {
+                height: 22px;
+                // 305px is approximately the size of the export button + the add tree button group + margins
+                // as determined from trial-and-error.
+                max-width: calc(100% - 305px);
+                display: inline-block;
+                overflow: hidden;
+                white-space: nowrap;
+                text-overflow: ellipsis;
             }
         }
-
-        .exportBtn {
-            margin-left: 10px;
-            margin-top: -2px;
-        }
-
         .addBtn {
+            display: inline;
             float: right;
-            margin-top: -6px;
-            margin-left: 8px;
-
-            i {
-                position: relative;
-                left: -4px;
-                top: 0;
-            }
+            margin-top: -33px;
         }
     }
 }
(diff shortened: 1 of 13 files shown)


opentreemap/treemap/js/src/lib/addResourceMode.js
```diff
@@ -120,11 +120,15 @@ function init(options) {
         U.$find('select', $form).on('change', enableFinalStep);
     }
```
RobinIsTheBird: Not sure if this is significant, but **Add a Bioswale** causes lines 118 and 120 to generate warnings in `utility.js` that their respective selectors are not found.

opentreemap/treemap/templates/treemap/partials/add_plot_btn.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
```
maurizi: I think the name right now is accurate enough - adding a tree necessarily entails adding a plot as well, and in fact the add process won't create a tree if you do not fill in tree fields.

Also, I don't think changing the name is enough of an improvement to refactor the code that is relying on the current value of the `data-feature` attribute.

opentreemap/treemap/templates/instance_base.html
```diff
@@ -134,8 +120,8 @@
   {% endif %}
 
   <div class="stats-bar hidden-xs">
-    <div class="half">
-      <div style="display: inline;" id="tree-and-planting-site-counts">
+    <div class="stats-list">
+      <div id="tree-and-planting-site-counts">
```
maurizi: That is more accurate, but there is JS code that uses this `id`, and I don't think it's a great use of our time to go changing it everywhere it is used.

opentreemap/treemap/js/src/lib/urlState.js
```diff
@@ -185,6 +195,13 @@ module.exports = {
         });
```
maurizi: Looks like there is already an open issue for this https://github.com/OpenTreeMap/otm-core/issues/2784

assets/css/sass/partials/_subheader.scss
```diff
@@ -187,31 +187,32 @@
...
-                float: right;
+            .exportBtn {
+                margin-left: 10px;
+                margin-top: -12px;
+                width: 135px;
+                display: inline-block;
+            }
+            #tree-and-planting-site-counts {
+                max-width: calc(100% - 305px);
+                display: inline-block;
```
maurizi: Honestly I arrived at most of these pixel amounts via trial-and-error, not from adding up the widths and margins of various other things. I added a comment to this effect in https://github.com/OpenTreeMap/otm-core/pull/2787/commits/e97505c17612282f7555c11a6cf045703589080e

opentreemap/treemap/js/src/lib/urlState.js
```diff
@@ -80,6 +80,12 @@ var serializers = {
         if (state.modeName) {
             query.m = state.modeName;
         }
+    },
+
+    modeType: function(state, query) {
+        if (state.modeType) {
+            query.t = state.modeType;
+        }
```
maurizi: Fixed in https://github.com/OpenTreeMap/otm-core/pull/2776

RobinIsTheBird:
:+1:

Issue #2769, opened 2016-09-07
Searching by Custom ID (and by missing Custom ID) should be enabled by default


RickMohr:
It will require a non-trivial data migration, but will be more easily discoverable.

This may be a useful starting point: https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/migrations/0009_restructure_replaceable_terms.py

PR #2789, opened 2016-09-20
Enable search by Custom ID & missing Custom ID

RobinIsTheBird:
Search by Custom ID and by missing Custom ID should be enabled
by default, for all treemap instances.

Changes
=======

## 0031_add_custom_id_to_default_search_fields.py
   Data migration to add the Custom ID internal identifier (`plot.owner_orig_id`) to every instance that already has a `search_config` or `mobile_search_fields` config property, with its reverse.

## treemap.search_fields.py
   Add `plot.owner_orig_id` to default search fields, and make default search fields `DotDict`s so that `instance.search_config` and `instance.mobile_search_fields` return the same type whether explicitly configured or not.

## treemap.DotDict.py
   Fix `DotDict().setdefault` to work for dotted keys.

--

Connects to #2769

### DIFF
--- a/opentreemap/treemap/migrations/0031_add_custom_id_to_default_search_fields.py
+++ b/opentreemap/treemap/migrations/0031_add_custom_id_to_default_search_fields.py
@@ -11,25 +11,20 @@ from treemap.DotDict import DotDict
 identifier = u'plot.owner_orig_id'
 
 
-def update_config_property(apps, update_fn, prop_name, *categories):
+def update_config_property(apps, update_fn, *categories):
     Instance = apps.get_model("treemap", "Instance")
-    filter_pattern = '\"{}\":'.format(prop_name)
-    for instance in Instance.objects.filter(config__contains=filter_pattern):
-        instance.config = update_fn(instance.config, prop_name, *categories)
+    for instance in Instance.objects.filter(
+            config__contains='\"search_config\":'):
+        instance.config = update_fn(instance.config, *categories)
         instance.save()
 
 
-def is_in_values(specs, pattern):
-    values = [spec.values() for spec in specs]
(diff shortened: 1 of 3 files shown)


opentreemap/treemap/migrations/0031_add_custom_id_to_default_search_fields.py
```diff
@@ -0,0 +1,76 @@
...
+            # mutates config[lookup]
+            specs.append({u'identifier': identifier})
+
+    return config
+
+
+def add_custom_id_forward(apps, schema_editor):
+    update_config_property(apps, add_to_config, 'search_config',
+                           'Plot', 'missing')
+    update_config_property(apps, add_to_config, 'mobile_search_fields',
```
jwalgran: There are a few outstanding problems with searching by Custom ID on mobile:

- The mobile apps do not know how to handle this special field, so they will do a substring search rather than an exact search.
- The mobile apps do not yet know how to select a single search result.

For these reasons, I think we should not add Custom ID as a default mobile search field.

opentreemap/treemap/search_fields.py
```diff
@@ -13,33 +13,39 @@
...
-        {'identifier': 'tree.height'}
+        {'identifier': 'tree.height'},
+        {'identifier': 'plot.owner_orig_id'}
     ],
     'missing': [
         {'identifier': 'species.id'},
         {'identifier': 'tree.diameter'},
-        {'identifier': 'mapFeaturePhoto.id'}
+        {'identifier': 'mapFeaturePhoto.id'},
+        {'identifier': 'plot.owner_orig_id'}
```
jwalgran: Mobile needs some more features before searching by Custom ID works well, so we should hold of on making it a default search field.

PR #2790, opened 2016-09-21
Add comment about pinning Firefox and selenium versions

RickMohr:


test-requirements.txt
```diff
@@ -1,3 +1,12 @@
...
+
+# We use Firefox and selenium for running UI tests. But using the latest 
+# versions is fragile because every few months a Firefox release is incompatible
+# with the current selenium. Since our UI tests don't depend on the latest features
+# of either we now use specific versions known to work together.
+#
+# Only update the selenium version number below if you know it works with
+# the currently-specified Firefox version.
+
+selenium==2.53.6             # https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES
```
maurizi: You can prevent requires.io from bugging us about updates to this by [adding a special comment](https://requires.io/features/)

It would look like this:
```txt
# https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES
selenium==2.53.6 # rq.filter: ==2.53.6

Comment (RickMohr) [inline test-requirements.txt]: Thanks, fixed.

File: test-requirements.txt [hunk 2]
```diff
@@ -1,3 +1,13 @@
...
 PyVirtualDisplay==0.2        # https://github.com/ponty/PyVirtualDisplay/releases
-selenium>=2.53.6,<3.0.0      # https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES
+
+# We use Firefox and selenium for running UI tests. But using the latest 
+# versions is fragile because every few months a Firefox release is incompatible
+# with the current selenium. Since our UI tests don't depend on the latest features
+# of either we now use specific versions known to work together.
+#
+# Only update the selenium version number below if you know it works with
+# the currently-specified Firefox version.
```
jwalgran: The specific version of Firefox is determined by the provisioner and is not specified in this repository. To clarify this, we may want to change this statement to something like "... if you know it works with the currently-provisioned Firefox version."

RickMohr: Thanks, updated.

PR #2792, opened 2016-09-21
Update requirements.txt to fix require.io version comment

maurizi:
It looks like requires.io won't pick up its `rq.filter` comment if there is another "comment" after it on that line.


RickMohr:
+1 (not tested)

Issue #1961, opened 2015-02-25
Need an "Add to Favorites" message when hovering on star


dboyer:
We need some type of message as to what happens when you click the star. If the star is empty, it should be "Add to Favorites." If it's filled in, it should be "Remove from Favorites."

PR #2801, opened 2016-09-22
When hovering on "Favorites" star, show message saying what will happen

RickMohr:
* Message is either "Add to Favorites" or "Remove from Favorites"
* Messages stored in data attributes so they can be translated
* Refactor favorite management code so message is shown in all cases

Testing:

Make sure favorite star and hover message appear correctly in all cases, including after an Edit/Save cycle.

Connects #1961

### DIFF
--- a/opentreemap/treemap/js/src/mapFeatureDetail.js
+++ b/opentreemap/treemap/js/src/mapFeatureDetail.js
@@ -103,6 +103,7 @@ function init() {
         } else {
             resourceDetail.init(form);
         }
+        updateFavoritedState(isFavoriteNow());
     }
 
     initDetail(form);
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/js/src/mapFeatureDetail.js
```diff
@@ -198,39 +199,62 @@ function init() {
...
+
+function setFavoritedState() {
     if (config.loggedIn) {
-        $('body').on('click', dom.favoriteLink, function(e) {
-            var $favoriteLink = $(dom.favoriteLink),
-                $favoriteIcon = $(dom.favoriteIcon),
-                wasFavorited = $favoriteLink.attr('data-is-favorited') === 'True',
-                url = $favoriteLink.attr(wasFavorited ? 'data-unfavorite-url' : 'data-favorite-url');
+        updateFavoritedState(isFavoriteNow());
+    }
```
maurizi: I think the `if (config.loggedIn) {` check needs to be removed from this function.

With it in place, I do not see a tooltip when not logged in.

RickMohr: Good catch! Fixed.

maurizi:
:+1:

PR #2804, opened 2016-09-23
Embed management page iframe sizes

RobinIsTheBird:
The iframe needs to use content-box because we can't count on the
customer's website styling the border or padding the same as ours.

--

Connects to OpenTreeMap/otm-addons#1270
Depends on OpenTreeMap/otm-addons#1273


maurizi:
:+1: (didn't test)

Issue #2185, opened 2015-07-30
The login link on the comments section should redirect back to the feature detail page


maurizi:
When not logged in, the plot detail page shows:

![image](https://cloud.githubusercontent.com/assets/2162993/18725967/636ad57a-8010-11e6-8350-00ed81aaf031.png)

jwalgran:
Related to https://github.com/OpenTreeMap/otm-core/issues/2239

RickMohr:
Blocked until #2239 is fixed

maurizi:
@RickMohr I believe this is unrelated to https://github.com/OpenTreeMap/otm-core/issues/2239.

RobinIsTheBird:
@maurizi, oth, some utility work for #2239 may come in handy in fixing this one.

PR #2805, opened 2016-09-23
Header login link preserves current page

RobinIsTheBird:
The header login link href now specifies the login page,
with the urlencoded current page in the `?next=` query string param.

Also changed the `InstanceUser` `__unicode__` method to do something
sensible when there is no logged in user.
Needed that for the django debug toolbar templates tab,
which was crashing in `InstanceUser.__unicode__`.

--

Connects to #2239
Connects to #2185

### DIFF
--- a/opentreemap/treemap/templatetags/auth_extras.py
+++ b/opentreemap/treemap/templatetags/auth_extras.py
@@ -4,6 +4,15 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django import template
+from django.conf import settings
+from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.core.exceptions import ValidationError
+from django.shortcuts import resolve_url
+from django.utils.encoding import force_str
+from django.utils.http import urlencode
+from django.utils.translation import ugettext_lazy as _
+
+from treemap.util import get_login_redirect_path
 
 register = template.Library()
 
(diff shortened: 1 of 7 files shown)


opentreemap/treemap/templatetags/auth_extras.py
```diff
@@ -215,3 +224,28 @@ def render(self, context):
...
+
+    Return an empty string if the current page is not an instance page.
+    """
+    request = template.Variable('request').resolve(context)
+
+    if getattr(request, 'user', None) and request.user.is_authenticated():
+        raise ValidationError(
+            _('Can\'t forward login if already logged in'))
+    if not getattr(request, 'instance', None):
+        return ''  # No point in forwarding back to a static page
```
jwalgran: Could you elaborate on why there is no point in forwarding back to a non-instance page? If I am on `/users/{username}` and click "Login" it would be nice and consistent to return to the profile page after logging in. Does including non-instance pages cause more trouble than it is worth?

opentreemap/treemap/templatetags/auth_extras.py
```diff
@@ -215,3 +224,28 @@ def render(self, context):
...
+    # urlparse chokes on lazy objects in Python 3, force to str
+    resolved_login_url = force_str(resolve_url(settings.LOGIN_URL))
+    return urlencode([(REDIRECT_FIELD_NAME,
```
jwalgran: Nice use of the constant to avoid having a literal "next" in the code.

opentreemap/treemap/templatetags/auth_extras.py
RobinIsTheBird: I forgot about `/users/{username}`. All the other non-instance pages I thought of were things like `not-available` or the static home page, in which case the current default of navigating to one's own account page seemed prudent.

I can also check for /users before returning the empty string. Are there more urls to consider? Is there a general criterion to check, or should I just special case each url as we think of it?

RobinIsTheBird:
Forgot to branch for the fix to #2185.

But I was waiting to merge this anyway until I can talk to @jwalgran about which pages should forward.

RobinIsTheBird:
@jwalgran, I'll enhance the criteria for adding the forward.

I think I also want put the `?` into login_forward and parameterize it to be `?` or `&` because the `?` in the html doesn't make sense if there's no query string.

PR #2806, opened 2016-09-23
Threaded comments migration

RobinIsTheBird:
New version of django module changed threaded comment metadata.
Update the schema for the subclass of the django model.
Necessary for release 2.12.0.

--

Connects to azavea/trees-team#85


maurizi:
:+1:

Issue #2698, opened 2016-08-08
api.tests.InstancesClosestToPoint.test_nearby_list_distance fails randomly


RobinIsTheBird:
Jenkins failed in api.tests.InstancesClosestToPoint.test_nearby_list_distance, line 1303:

```
self.assertEqual(self.i1.pk, instance_infos['nearby'][0]['id'])
AssertionError: 864 != 868
```

After an `@ azavea-bot rebuild`, it succeeded.

I suspect that the ordering of the 'nearby' instance infos is not guaranteed. We should investigate, and if it isn't guaranteed, then assertIn may be more appropriate.

PR #2817, opened 2016-10-05
Fix intermittently-failing test

RickMohr:
We were calling `instances_closest_to_point` with four instances, adding another instance, and calling `instances_closest_to_point` again. The second call was intermittently failing because the distance is 0 to two of the instances, making for an indeterminate ordering when sorting by distance. (The distance is computed to the instance bounds, and it's 0 for two instances because
the point is inside their bounds.)

I removed the second call because it does not appear to test any functionality not covered
by the first call. The results of the first call don't have the indeterminate order problem.

Connects #2698

### DIFF
--- a/opentreemap/api/tests.py
+++ b/opentreemap/api/tests.py
@@ -1286,7 +1286,6 @@ class InstancesClosestToPoint(OTMTestCase):
 
         self.assertEqual(0, len(instance_infos['personal']))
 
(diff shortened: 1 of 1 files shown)


RobinIsTheBird:
:+1:  assuming jenkins approves.

I wonder what the purpose of the added instance was.

Issue #2700, opened 2016-08-10
treemap.tests.test_udfs is unreliable


RobinIsTheBird:
After pushing a PR, jenkins failed with the complaints below. Rerunning it with @ azavea-bot reload resulted in success. Nothing in the PR has anything to do with UDFs AFAIK.

```
======================================================================
FAIL: test_can_delete_choice_option (treemap.tests.test_udfs.CollectionUDFTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/app/core/opentreemap/treemap/tests/test_udfs.py", line 1040, in
test_can_delete_choice_option
    self.assertEqual(plot.udfs['Stewardship'][0]['action'], '')
AssertionError: 'prune' != u''

======================================================================
FAIL: test_can_update_choice_option (treemap.tests.test_udfs.CollectionUDFTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/app/core/opentreemap/treemap/tests/test_udfs.py", line 1015, in
test_can_update_choice_option
    self.assertEqual(plot.udfs['Stewardship'][0]['action'], 'h2o')
AssertionError: 'prune' != u'h2o'
```

RobinIsTheBird:
Note that 3 tests in this test module are now being skipped.

PR #2818, opened 2016-10-05
Fix tests which rely on order of collection field values

RickMohr:
The values in a collection field (like Stewardship) are unordered. That was a surprise to me, but we get away with it because the collection fields we've used (Stewardship, Alerts) always have a date member and we can sort by that.

When we reference the values of a collection field, like here:
https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/tests/test_udfs.py#L1087
their order is indeterminate. Here's where we build the list of values:
https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/udf.py#L935

On the other hand, note that the order of `plot.audits()` is deterministic, since we can sort by `created`:
https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/udf.py#L1117

To fix the tests:
* Use sets when comparing values of a collection field.
* Use a filter to remove a particular collection field value.

Connects #2700

### DIFF
--- a/opentreemap/treemap/tests/test_udfs.py
+++ b/opentreemap/treemap/tests/test_udfs.py
@@ -7,7 +7,6 @@ import json
 from random import shuffle
 from datetime import datetime
 import psycopg2
-from unittest.case import skip
 
 from django.db import connection
 from django.db.models import Q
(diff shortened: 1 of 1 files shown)


RobinIsTheBird:
:+1: after you remove the unused import of `unittest.case.skip`

Issue #1127, opened 2014-02-20
On iPad in vertical mode, the species search box disappears


dboyer:
May be linked to the logo being so large that it interferes with the search boxes.

lederer:
Verified this is still the case. 

The correct fix is to make the site responsive to a suitable range of viewport widths.

In the meantime, users can rotate to landscape, which is arguably the more likely orientation for using OTM on a tablet anyway.

Setting to low priority.

PR #2824, opened 2016-10-10
Implement mobile responsive search for small screen sizes

maurizi:
With these changes, all advanced search functionality is now available on small screen sizes.

Map page on page load:
screen shot 2016-10-10 at 14 15 46

When tapping on the location search box:
screen shot 2016-10-10 at 14 18 43

After tapping on the "Advanced search" toggle in the header:
screen shot 2016-10-10 at 14 19 59

Search buttons are sticky when scrolling:
screen shot 2016-10-10 at 14 20 05

For medium-sized screens we now always show both the species and location search boxes (by hiding the map logo):
screen shot 2016-10-10 at 14 20 24

Connects to https://github.com/OpenTreeMap/otm-core/issues/1127

### DIFF
--- a/assets/css/sass/_base.scss
+++ b/assets/css/sass/_base.scss
@@ -44,8 +44,8 @@ $table-gray                     : #DFE2E8;
 $screen-xs    : "(max-width: 767px)";
 $screen-gtxs  : "(min-width: 768px)";
 $screen-sm    : "(min-width: 768px) and (max-width: 991px)";
+$screen-gtsm  : "(min-width: 992px)";
 $screen-ltmd  : "(max-width: 991px)";
 $screen-md    : "(min-width: 992px) and (max-width: 1199px)";
 $screen-ltlg  : "(max-width: 1199px)";
 $screen-lg    : "(min-width: 1200px)";
-
--- a/assets/css/sass/_base.scss
+++ b/assets/css/sass/_base.scss
@@ -41,10 +41,11 @@ $table-gray                     : #DFE2E8;
 
 
 // Media Queries
-$screen-xs			 : "(max-width: 767px)";
-$screen-gtxs         : "(min-width: 768px)";
-$screen-sm			 : "(min-width: 768px) and (max-width: 991px)";
-$screen-md			 : "(min-width: 992px) and (max-width: 1199px)";
-$screen-ltlg         : "(max-width: 1199px)";
-$screen-lg			 : "(min-width: 1200px)";
+$screen-xs    : "(max-width: 767px)";
+$screen-gtxs  : "(min-width: 768px)";
+$screen-sm    : "(min-width: 768px) and (max-width: 991px)";
+$screen-ltmd  : "(max-width: 991px)";
+$screen-md    : "(min-width: 992px) and (max-width: 1199px)";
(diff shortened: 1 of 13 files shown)


maurizi:
@lederer I implemented the changes you suggested in 2277e49

lederer:
@maurizi Thanks for that. I'd forgotten to update the PR with the mock yesterday.

lederer:
👍

RickMohr:
The little user image photo is gone. Before:
image
After:
image

RickMohr:
If I do a location search on mobile and want to clear the results I have to think to click "Advanced Search", which might be a stretch for many. (Then after I click "Reset" I have to click the "X" to get back out.)

Likewise, if I do an advanced search and want to clear the results I have to click "Advanced Search" + Reset + X.

Can we find a way to get that "Reset" button onto the header for mobile?

RickMohr:
The black header bar is slightly taller in "mobile" mode than the other modes, and the first label is a little further to the right:
image  image
image

(Watch it dynamically by changing the window width.)

RickMohr:
Can we put "Tree Care" at the bottom in mobile mode? Seems least important, but it's shown first.

RickMohr:
In mobile mode, the "Search by Species" and "Search by Location" headers should be the same color as the other headers:
image

RickMohr:
There's a temporary layout glitch when loading in non-mobile mode. In the "Search by Species" and "Search by Location" boxes the placeholder text overlays the hamburger icon for a noticeable amount of time, looking ugly. Then it's resolved.

Not present on staging.

RickMohr:
That's all I can find for now.

maurizi:
> There's a temporary layout glitch when loading in non-mobile mode. In the "Search by Species" and "Search by Location" boxes the placeholder text overlays the hamburger icon for a noticeable amount of time, looking ugly. Then it's resolved.

I saw this too, and I have no idea how to fix it.  @lederer do you have any ideas? I could use an assist.

maurizi:
@RickMohr Most of the comments you raised have been addressed, this is ready for another look.

lederer:
@maurizi Re that layout glitch, I haven't been able to track down the precise cause, but I've got some evidence that might help you get there:

1. Using Chrome Inspector, I recorded the timeline of the page reload.
1. The frame that gets rendered at 1.64s has the offending styles, which come from `input[type="text"]`.
1. That frame remains onscreen until 2.41s, or ~7.64ms, which is a lifetime.
1. The next frame is when the correct styles for those elements get rendered. They come from `.search-block-wrapper .search-block .search-field-group .twitter-typeahead input[type="text"]`.

So at the risk of a tautology, I reckon there's something going on with that huge rendering delay / low FPS, and with Twitter Typeahead. Holler and I can show you the timeline I captured.

maurizi:
@lederer That was a huge help :smile: . `twitter-typeahead` is a class that gets added by JS, so depending on it for styling stuff wasn't a good idea.

@RickMohr I put back the user logo and fixed a styling bug with it, fixed the species/location label coloring, moved tree care to the bottom of  the search area, and reworked how the advanced / reset buttons work.

thanks for all the thoughtful comments!

RickMohr:
@maurizi, all looks good!

:+1:

PR #2827, opened 2016-10-14
Show plot detail page fields based on configurable order

maurizi:
Required by https://github.com/OpenTreeMap/otm-addons/pull/1297

Connects to OpenTreeMap/otm-addons#1278

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -21,11 +21,11 @@ from urllib import urlencode
 
 from opentreemap.util import extent_intersection, extent_as_json
 
-from treemap.search_fields import (DEFAULT_MOBILE_SEARCH_FIELDS,
-                                   DEFAULT_MOBILE_API_FIELDS,
-                                   DEFAULT_SEARCH_FIELDS, API_FIELD_ERRORS,
-                                   advanced_search_fields,
-                                   get_udfc_search_fields)
+from treemap.search_fields import (
+    DEFAULT_MOBILE_SEARCH_FIELDS, DEFAULT_MOBILE_API_FIELDS,
+    DEFAULT_WEB_DETAIL_FIELDS, DEFAULT_SEARCH_FIELDS, INSTANCE_FIELD_ERRORS,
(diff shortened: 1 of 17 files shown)


opentreemap/treemap/instance.py
```diff
@@ -274,6 +274,9 @@ def set_config(self, value):
     mobile_api_fields = _make_config_property('mobile_api_fields',
                                               DEFAULT_MOBILE_API_FIELDS)
 
+    web_detail_fields = _make_config_property('web_detail_fields',
```
jwalgran: While testing I created a new UDF, set its editability permissions, and was briefly puzzled by the fact that I did not see it appear on the tree detail page. We now have to visit 3 management pages to add a field, one to create, one to adjust role permissions, and one to add it to the detail page. There is never a reason to add a UDF and not display it to anyone so I think we need to do something to help the manager understand what they need to do.

Some ideas:

- Automatically add the new field to the fields list.
- Add another notification "bubble."

screen shot 2016-10-17 at 3 13 06 pm

jwalgran:
I had originally included a comment in this review that was meant for a different PR. I have now moved it.

opentreemap/treemap/views/map_feature.py
```diff
@@ -101,13 +102,48 @@ def _map_feature_detail_context(request, instance, feature_id, edit=False):
...
+        v: k for k, v in context['plot'].scalar_udf_names_and_fields})
+
+    def info(group):
+        group['fields'] = [
+            (field, labels.get(field),
+             templates.get(field, "treemap/field/tr.html"))
+            for field in group.get('field_keys', [])
+        ]
+        group['collection_udfs'] = [
+            next(udf for udf in udf_defs(instance)
```
maurizi: I considered writing defensive code here to protect from that, but decided it wasn't worth it.  If we start making `Instance`s that lack Stewardship, a _lot_ of things will blow up all over the code base

maurizi:
@jwalgran This is ready for another look

PR #2828, opened 2016-10-17
Separate authorization from tracking

RobinIsTheBird:
Rename `joint_writable` to start with `_`, and comment more fully. Always initialize it to include
`id` in `Authorizable`. Update it with MapFeature and Tree fields, both the fields that already were
in it and fields that we no longer want to expose to role administration.

Also minor refactoring.

--

Depends on #2821 
Depends on OpenTreeMap/otm-cloud#305
Depends on OpenTreeMap/otm-addons#1290

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -781,6 +781,11 @@ class Authorizable(UserTrackable):
 
         self._has_been_masked = False
 
+        # When any other field is writeable, any field in `_joint_writable`
+        # is writable.
+        # `_joint_writable` fields are always readable.
+        self._joint_writable = set(['id'])
+
     def _get_writable_perms_set(self, user, direct_only=False):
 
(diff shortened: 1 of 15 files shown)


opentreemap/treemap/audit.py
```diff
@@ -781,14 +781,19 @@ def __init__(self, *args, **kwargs):
 
         self._has_been_masked = False
 
+        # When any other field is writeable, any field in `_joint_writable`
+        # is writable.
+        # `_joint_writable` fields are always readable.
+        self._joint_writable = set(['id'])
```
maurizi: Any reason this is `set(['id'])` instead of `{'id'}`?

RobinIsTheBird:
Forgot to tell the PR to compare against RickMohr/instance-permissions. Closing and will re-open properly.

Issue #2825, opened 2016-10-12
Separate audit trackability from permissibility


RobinIsTheBird:
This issue splits out of #2176 : Create non-field permissions...

Once instance-level permissions are in place, permission is meaningless on certain fields, notably `Tree.plot`, and `id`, `readonly` and `geom` fields on all relevant models. However, we still want to track them in the audit trail in case of errors or hacks.

`treemap.audit.Authorizable`, which is chiefly though not solely responsible for permissioning, is already separate from `treemap.audit.Auditable`, both of which are based on `treemap.audit.UserTrackable`.

`UserTrackable` has a `do_not_track` list.

We could add a `skip_authorization` list to `Authorizable`. A number of other modules would be affected.

RickMohr:
Looks good. I'd remove `readonly` from the list, since if we implement it we will want a field permission.

RobinIsTheBird:
@RickMohr it's easy enough to put back. While it's unavailable, I think it should be hidden from the user, and would prefer not to put 2 criteria for hiding from the user into the Roles page.

RickMohr:
Fair enough. Have a good sleep, `readonly`.

PR #2830, opened 2016-10-18
Separate authorization from tracking

RobinIsTheBird:
Leverage `joint_writable` and comment more fully.
It now encompasses all required fields that make no sense
to authorize separately from the new add and delete permission.

Initialize `joint_writable` in `Authorizable` to include `id`.
Update it with MapFeature and Tree fields.

Making required fields not authorizable conflicts with
insertion and deletion pending approval, so disable
relevant pending tests and drop support for
insert and delete pending approval.

Also minor refactoring.

--

Connects to #2825

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -29,7 +29,7 @@ from treemap.units import (is_convertible, is_convertible_or_formattable,
 from treemap.util import (all_models_of_class, leaf_models_of_class,
                           to_object_name, safe_get_model_class)
 
-from treemap.lib.object_caches import (permissions as field_permissions,
+from treemap.lib.object_caches import (field_permissions,
                                        invalidate_adjuncts, udf_defs)
 from treemap.lib.dates import datesafe_eq
 
@@ -790,10 +790,8 @@ class Authorizable(UserTrackable):
 
         self._has_been_masked = False
 
-        # When any other field is writable, any field in `_joint_writable`
-        # is writable.
-        # `_joint_writable` fields are always readable.
-        self._joint_writable = {'id'}
+        # `_always_writable` fields are also always readable.
+        self._always_writable = {'id'}
 
     def get_instance(self):
         instance = getattr(self, 'instance', None)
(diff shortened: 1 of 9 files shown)


opentreemap/treemap/audit.py
```diff
@@ -812,17 +817,20 @@ def _get_writeable_perms_set(self, user, direct_only=False):
...
+        field_perms = field_permissions(user, self.get_instance())
         can_write_anything = bool({perm.field_name for perm
-                                   in permissions(user, self.get_instance())
+                                   in field_perms
                                    if perm.allows_writes})
         if can_write_anything:
-            return getattr(type(self), 'joint_writable', set())
+            return self._joint_writable
         else:
             return set()
```
RickMohr: I think we should remove the `can_write_anything` check (and update the comment). It complicates the logic and doesn't add significant safety. It raises the possibility that I could have create permission for e.g. plots but no permission to modify fields, and thus be prevented from creating a plot because `id` was not in the set of writable fields. `id`, `updated_at`, `geom`, and `tree.plot` should truly be exempt from field permissions.

opentreemap/treemap/audit.py
```diff
@@ -790,6 +790,11 @@ def __init__(self, *args, **kwargs):
 
         self._has_been_masked = False
 
+        # When any other field is writable, any field in `_joint_writable`
+        # is writable.
+        # `_joint_writable` fields are always readable.
+        self._joint_writable = {'id'}
```
RickMohr: I think we should call this field `_always_writable` instead of `_joint_writable`, and remove the first two comment lines.

opentreemap/treemap/audit.py
```diff
@@ -29,7 +29,7 @@
 from treemap.util import (all_models_of_class, leaf_models_of_class,
                           to_object_name, safe_get_model_class)
 
-from treemap.lib.object_caches import (permissions,
+from treemap.lib.object_caches import (permissions as field_permissions,
```
RickMohr: In `object_caches` we should rename `permissions` to `field_permissions` so this clarification would be used everywhere. Not necessarily as part of this PR.

opentreemap/treemap/models.py
```diff
@@ -593,6 +590,15 @@ def __init__(self, *args, **kwargs):
         self._do_not_track.add('feature_type')
         self._do_not_track.add('mapfeature_ptr')
         self._do_not_track.add('hide_at_zoom')
+
+        # Tells the permission system that if any other field in the role
+        # is writable, _joint_writable fields are also writable.
```
RickMohr: If you agree with my proposals above, we can remove the first two comment lines.

RobinIsTheBird:
@RickMohr , Partly took your suggestion of renaming `permissions` to `field_permissions` in `treemap.lib.object_caches`. Left an alias named `permissions` until plug-in repositories catch up.

RickMohr:
Looks good but a couple tests failed.

RobinIsTheBird:
@azavea-bot rebuild

RobinIsTheBird:
@RickMohr The tests succeeded. The red X is from coveralls because I skipped a few tests, just until the cutover to model-level permissions. Please go ahead and re-review.

RickMohr:
:+1:

PR #2831, opened 2016-10-18
Add ipython and ipdb to dev-requirements.txt

RobinIsTheBird:
For easy access to ipython,
* enable `django-extensions` in `local_settings.py`
* `manage.py shell_plus`

`ipython-genutils` came with the `ipdb` install. We can ignore it, but it seemed better to make it (and its version) explicit in dev-requirements.txt.

`notebook` is an `ipython` extension for serving documentation. If someone takes the time to add our markdown files to a notebook, we can serve them as html on our localhosts. Could be convenient, especially for previewing changes.

We can also use a notebook to script access to a `shell_plus` kernel, for live, interactive documentation, if we can figure out how.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,3 +1,5 @@
 django-debug-toolbar==1.5     # https://django-debug-toolbar.readthedocs.io/en/stable/changes.html
 django-extensions==1.7.4
+ipython==5.1.0
+ipdb==0.10.1
 six==1.10.0


dev-requirements.txt
```diff
@@ -1,3 +1,7 @@
 django-debug-toolbar==1.5     # https://django-debug-toolbar.readthedocs.io/en/stable/changes.html
 django-extensions==1.7.4
+ipython==5.1.0
+ipython-genutils==0.1.0
+ipdb==0.10.1
+notebook==4.2.3
```
maurizi: I recognize `ipython` and `ipdb`, but `notebook` and `ipython-genutils` are new to me.  What do they get us?

RobinIsTheBird: See the updated PR description above, and the (identical) amended commit message.

RickMohr: Great to have `ipython` and `ipdb`.

Seems to me that if `ipdb` is bringing in `ipython-genutils` we should let it manage the versioning. Wouldn't omitting `ipython-genutils` prevent possible version number conflicts?

I'd leave out `notebook` until we're actually using it. If someone wants to create doc notebooks as you suggest, starting with a `pip install notebook` won't be a high barrier.

jwalgran:
Looks good to me

Issue #2833, opened 2016-10-26
When loading from a zoomed-out URL, initial zoom is too far in


RickMohr:
1) Load a treemap
2) Zoom out
3) Reload the page

Expected: map is zoomed out one level from its initial zoom
Actual: map is at its initial zoom

PR #2834, opened 2016-10-26
Fix initial map zoom when loading from a URL

RickMohr:
`MapManager.setCenterAndZoomLL` prevents zooming out (so e.g. if user is choosing a tree location via geolocation but is zoomed in further than ZOOM_PLOT, we won't zoom them out). That's not appropriate when initializing a map from a saved zoom, so just use `map.setView`.

Connects #2833


maurizi:
:+1:

PR #2837, opened 2016-10-27
Prevent urlState crash

RickMohr:
`urlState.js` keeps an object in `window.history.state` and, on history changes, uses it to detect which part of a URL has changed. Things work fine when history changes are initiated by the `urlState` module itself. But clicking a link with `href="#"` without calling `preventDefault()` also causes a history change, and we weren't handling that correctly.

So:
1. Replace occurrences of `href="*"` with `href="javascript:;"`. It's just safer. Adding `#` to the URL is not always innocuous, and we don't want to rely on `preventDefault` being called.
2. Don't assume our history state object exists. On history changes, construct our state object from the URL.

Either of these changes would suffice by itself. Do both -- belt and suspenders.

And it turns out that we don't actually need to store an object in `window.history.state`, so `urlstate.js` can be simplified accordingly.

Also added module overview documentation for `urlState.js`.

Connects #2784

To make sure nothing is broken, verify that app changes are reflected in the URL, and properly re-established when refreshing the page:
- Change zoom/lat/long
- Change mode of map page (add tree / add resource / quick edit)
- Change search

### DIFF
--- a/opentreemap/importer/templates/importer/partials/clientside_templates.html
+++ b/opentreemap/importer/templates/importer/partials/clientside_templates.html
@@ -81,16 +81,16 @@
 <script type="text/template" id="pager-template">
   <div class="pages pagination pagination-small">
     <ul>
-      <li><a href="#">&lt;</a></li>
-      <li><a href="#">&lt;&lt;</a></li>
+      <li><a href="javascript:;">&lt;</a></li>
+      <li><a href="javascript:;">&lt;&lt;</a></li>
       <% _.each(_.range(start_page, end_page), function(i) { %>
         <li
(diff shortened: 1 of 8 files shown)


RobinIsTheBird:
quick edit does not change the url.
zoom, pan, search, and search reset all do.

RickMohr:
Quick Edit doesn't change the URL on production either.

RobinIsTheBird:
@RickMohr the only problem I found is quick edit not changing the url. Also tested RainBarrel and Bioswale. The code all looks fine. If you can track down quick edit, I can approve the PR.

RobinIsTheBird:
:+1:

PR #2840, opened 2016-11-03
Add migration rules for Treezilla

jwalgran:
- The otm-legacy version of Treezilla stores tree diameter and height in the database in meters. OTM2 stores diameters in inches and heights in feet and converts to metric units on the fly.

- The `city` column of the `treemap_neighborhood` table in the otm-legacy database groups regions into different categories, but based on our research [1] [2] different categories do not overlap. I added a presave action to give the regions full category names and gave them the same sort order value, since they do not overlap.

- I set the UDF choice field values based on the the response from the otm-legacy endpoint http://treezilla.org/choices/.

This PR also includes a fix to comment migration handling. The edited line in `model_processors.py` previously raised an exception if the comment was skipped and the comment contained one or more non-ASCII characters.

[1] https://github.com/openaustralia/phplib/blob/802a25b1831d5ba9e0c8f8c859ddf8180d9d70c1/votingarea.php#L176
[2] http://www.cpag.org.uk/sites/default/files/ONS%20Map%20UK%20local%20authorities%202009.pdf

---

Connects to https://github.com/OpenTreeMap/otm-clients/issues/260

### DIFF
--- a/opentreemap/otm1_migrator/migration_rules/treezilla.py
+++ b/opentreemap/otm1_migrator/migration_rules/treezilla.py
@@ -1,0 +1,140 @@
+from otm1_migrator.data_util import MigrationException
+from otm1_migrator.migration_rules.standard_otm1 import MIGRATION_RULES
+
+from treemap.models import ITreeCodeOverride, ITreeRegion, User
+
+TREEZILLA_ITREE_REGION_CODE = 'NoEastXXX'
+METERS_TO_INCHES = 39.3701
+METERS_TO_FEET = 3.28084
+
+BOUNDARY_TYPES = {
+    'DIS': {'name': 'District', 'sort_order': 10},
+    'LBO': {'name': 'London Borough', 'sort_order': 10},
+    'MTD': {'name': 'Metropolitan District', 'sort_order': 10},
+    'UTA': {'name': 'Unitary Authority', 'sort_order': 10},
+}
+
+
+def set_boundary_fields(boundary_obj, boundary_dict):
+    # The 'city' column in the Treezilla boundary table is a boundary type code
+    boundary_type = BOUNDARY_TYPES.get(
+        boundary_dict['fields'].get('city', None), None)
+    if boundary_type is None:
+        raise MigrationException("boundary_dict missing valid city value: " +
+                                 str(boundary_dict))
+    boundary_obj.category = boundary_type['name']
+    boundary_obj.sort_order = boundary_type['sort_order']
+    return boundary_obj
+
+MIGRATION_RULES['boundary']['presave_actions'] = (MIGRATION_RULES['boundary']
+                                                  .get('presave_actions', [])
+                                                  + [set_boundary_fields])
+
+
+def convert_tree_measurements(tree_obj, tree_dict):
+    """
+    Treezilla's OTM1 database stores both the tree diamater and height
+    in meters. OTM2 stores diameter in inches and height in feet and
+    converts on the fly to metric. Treezilla does not have any values
+    in their canopy_height column.
+    """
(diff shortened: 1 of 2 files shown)


opentreemap/otm1_migrator/migration_rules/treezilla.py
```diff
@@ -0,0 +1,140 @@
...
+    'tree': {
+        'sponsor': {'udf.name': 'Sponsor'},
+        'pests': {
+            'udf.name': 'Pests',
+            'udf.choices': [
+                'Ash dieback (Chalara fraxiniae)',
+                'Asian longhorn beetle (Anoplophora glabripennis)',
+                'Bronze birch borer (Agrilus anxius)',
+                'Emerald ash borer (Agrilus planipennis)',
+                'Horse chestnut bleeding canker (Pseudomonas syringae pv aesculi)',
```
maurizi: `flake8` is complaining about line length here.

maurizi:
@jwalgran Code looks good to me. I haven't tested an OTM1 migration before, perhaps @RickMohr would be better for this?

RickMohr:
When I tested the Tampa migration I did not run the migration myself, just compared things between the live OTM1 site and staged OTM2 site (which is currently http://pixie3.open.ac.uk/treezilla)

RickMohr:
...and I'm happy to do it if that would help.

RickMohr:
Should we remove redundancy in the boundary names? Currently:
image
Could consider:
* Aylesbury Vale
* Babergh
* Barking and Dagenham
(with the category names shown below each entry as now)

jwalgran:
I do think would look better if the category name was not part of the boundary name. I was going to avoid doing any string manipulation on the boundary names until we had Open University review the migration.

RickMohr:
Ecobenefit totals are a bit different between the OTM1 and OTM2 instances, e.g.:
* Energy:  932,038 OTM1 vs. 961,923 OTM2
* Stormwater: 6,561,463 OTM1 vs. 6,711,502 OTM2

Do we care?

RickMohr:
Did we lose postal code information?

OTM1:
image

OTM2:
image

RickMohr:
[This OTM1 tree](http://www.treezilla.org/plots/4002) has "Original ID" = 280

But I don't see "Owner Original ID" on [the corresponding OTM2 tree/plot](http://pixie3.open.ac.uk/treezilla/features/1511) or any other plot.

RickMohr:
Ecobenefits are quite different for specific trees. For the tree links in the previous comment:

OTM1:
image

OTM2:
image

jwalgran:
> Ecobenefit totals are a bit different

I don't think this discrepancy is important, but I would like to hear from @dboyer 

> Did we lose postal code information?

OTM1 has a `geocoded_address` field, which is what is being displayed on the detail page. The migrator does not attempt to parse this field and only migrates the address1, address2, etc. fields.

> ... I don't see "Owner Original ID"  ...

I had incorrectly set Owner Original ID as invisible. It is now fixed.

> Ecobenefits are quite different for specific trees

Thank you for finding this example. We do need to determine why the discrepancy is so large for specific trees.

RickMohr:
Was setting "Original Owner ID" to be not invisible a manual step? If so we'll want to capture that kind of  configuration -- ideally in a script, but at least in a doc.

jwalgran:
> Was setting "Original Owner ID" to be not invisible a manual step? If so we'll want to capture that

A guide to setting field permissions is in a draft README in a format similar to ones used for previous migrations.

RickMohr:
Nice work! I'd call it a :+1: with a follow-on issue to investigate the ecobenefits.

PR #2858, opened 2016-11-16
Allow using non-ASCII characters when hashing for an etag

RickMohr:
Connects OpenTreeMap/otm-addons#1160


RickMohr:
Close/reopen to link correct issue in waffle.

RickMohr:
@azavea-bot rebuild

RickMohr:
The PR builder shows a failing test. But the test passes on my machine and on circleci/travis, and inspecting the code I don't see how the failure could happen.

Issue #2870, opened 2016-11-23
Cannot move tree as part of quick edit


dboyer:
Tested in Chrome 

1. Added a tree
2. Moved tree's location on the tree detail page. Location updated correctly.
3. Returned to map page and selected the tree dot
4. Clicked Quick Edit
5. The "Move Tree" button was grayed out and hover text said "Editing a plot's location is not available to all users"

I'm the map creator and an admin and have full write access to all fields according to the Roles page.

dboyer:
I can now edit the tree's location from the quick edit screen as an admin but not user with the public role. I can edit the tree's location as a public user from the detail page.

I changed the public role setting to have full write access to the address, city, and postal code fields. After that, I was able to save location changes using quick edit.

jwalgran:
Deb was able to reproduce the issue on her test map, but I had trouble recreating it on a new test map of my own.

I created a new map and added a tree, then switched to a user in the Public role for that map. I was unable to click Quick Edit on the map or Edit on the detail page because the public role does not have any write access by default.

When I made address fields editable, I was able to Quick Edit move the planting site on the map and Edit move the planting site on the detail page.

RobinIsTheBird:
@jwalgran, may I take this issue back?

RobinIsTheBird:
@dboyer and @jwalgran,

I haven't been able to reproduce the problem, but do have a few thoughts on it:

1. It is necessary to have Write Access to the Planting Site Address field in order to move a planting site at all. Currently, the Move buttons are enabled even when Address is not permitted. They should be disabled. Is that a release blocker, or should I just create a new issue?
2. If you have Write Access to Address but not to City or Zip, then you can move the tree within the same city or zip and save the move, but if you move it across city or zip boundaries, you cannot save. If the tree was close to the boundary, and you moved it within the boundary on the detail page but across the boundary on the map page, that could lead to the appearance of inconsistent behavior. Could this be what happened?

I'll try to reproduce #2 on my local host.

If #2 turns out to be what you were experiencing, I really don't know what to do about it. We would need to consult Scott. At the minimum, our error reporting could be vastly improved.

RobinIsTheBird:
Created some issues:

- #2884 Move buttons are enabled when the user cannot actually move a map feature
- #2885 Authorization Exception causes a stack trace and no user message
- #2886 Logout then Save produces no message

Are any of these release blockers?

RobinIsTheBird:
[...]

PR #2872, opened 2016-11-23
Fix quick edit move tree

RobinIsTheBird:
Check whether the plot is writable, not specific field.

--

Connects to #2870


maurizi:
Code looks good to me (didn't test)

jwalgran:
Is it possible to add to the commit message and explain why this is a good fix? Looking at just this diff without the broader scope of the permission system it may appear counterintuitive, since moving the plot only affects geometry field.

RobinIsTheBird:
@jwalgran, I amended the commit message.

Git decided that my local branch was behind my fork branch. That's why there's a merge commit.

maurizi:
:+1:

Issue #2864, opened 2016-11-17
Add endpoint for searching for users on a map by username


RobinIsTheBird:
Intended to support an autocomplete for "Last Updated By"

Required for #388

The response format should be similar to the species and boundary autocomplete endpoints, so that it can be used by typeahead.js (but it _may_ not need the `tokens` list as filtering will be implemented by the backend)

Issue #388, opened 2013-09-19
Allow filtering map features / ecobenefits by last updated by (user)


dboyer:
This is just for the UI - the backend is a separate feature

The username input field should be an autocomplete that allows for searching by username.

RobinIsTheBird:
Blocked by https://github.com/OpenTreeMap/otm-core/issues/2863

RobinIsTheBird:
Counting users per instance via
```
users_per_instance = Instance.objects\
    .annotate(num_users=Count('instanceuser'))\
    .order_by('-num_users','url_name')
for upi in users_per_instance[:20]:
    print('{} {}'.format(upi.num_users, upi.url_name))
```
we see
```
1161 phillytreemap
742 latreemap
357 treefolksaustin
```
then four between 100 and 200, and the rest less than 100.

Imo the search view should prefetch N, and fetch the rest dynamically. Then the questions become what is N, and which N to fetch if there are more than that - probably the first N ordered by `username`.

I'll profile prefetching `phillytreemap`'s thousand plus users and see whether it's worth the extra complexity of round-trip searching.

RickMohr:
Seems like it would be simpler to always prefetch or never prefetch rather than needing to handle the case of a partial prefetch. If `phillytreemap` works fine with prefetch, great. And if not, performance without prefetch might be fine for this feature.

maurizi:
> Seems like it would be simpler to always prefetch or never prefetch rather than needing to handle the case of a partial prefetch.

Agreed, I don't think its worth from a code-complexity standpoint to do a partial prefetch.

Since we expect the user list to be invalidated fairly frequently (certainly more so than species or boundaries), we don't derive as much value from prefetching it, since it won't be cached all that often.

RobinIsTheBird:
How long does a user typically stay on the map page in the web browser? I doubt if the users are added so frequently that it would invalidate the prefetch on page enter very often. And if the seeker is looking for changes from a new user, they probably have an idea that the new user just joined, and would think of doing a page refresh.

The code from #2873 is designed for the round trip per search paradigm.

What say I write an alternate endpoint for prefetch and try it with phillytreemap?

This feature will not be ready in time for tomorrow's RRP in any case, regrettably.

RickMohr:
I'd be inclined to build it without prefetch first and see how it performs. If it's fine we save ourselves the time of implementing the prefetch.

maurizi:
From our Slack discussion:
[...]

PR #2873, opened 2016-11-23
Add endpoint for searching for users on a map

maurizi:
Intended for implementing OpenTreeMap/otm-core#388

Connects to #2864

### DIFF
--- a/opentreemap/treemap/tests/test_views.py
+++ b/opentreemap/treemap/tests/test_views.py
@@ -1963,37 +1963,34 @@ class UserAutocompleteTest(OTMTestCase):
         self.i2 = make_instance()
 
         self.mike = make_user(instance=self.i1, username='mike')
+        self.also_mike = make_user(instance=self.i1, username='i-am-mike')
         self.maria = make_user(instance=self.i1, username='Maria')
         self.matt = make_user(instance=self.i2, username='MATT')
 
-    def assert_users_in_list(self, users_list, *users):
-        self.assertEqual(len(users_list), len(users))
+    def assert_users_in_list(self, instance, params, *expected_users):
+        users_list = users(make_request(params), instance)
+        self.assertEqual(len(users_list), len(expected_users))
 
-        for i in range(0, len(users)):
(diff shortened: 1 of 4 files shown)


opentreemap/treemap/views/user.py
```diff
@@ -208,3 +208,34 @@ def user(request, username):
...
+                           .values('user_id', 'user__username',
+                                   'user__first_name', 'user__last_name',
+                                   'user__make_info_public')
+
+    if query:
+        users_qs = users_qs.filter(user__username__icontains=query)
+
+    if max_items:
+        users_qs = users_qs[:int(max_items)]
+
```
RobinIsTheBird: Whether or not to accept this depends on how much we expect it to get used on maps with tens of thousands of users. Maybe we need input from @dboyer.

If there are many results, but the search string is late in the alphabet, the matches that start with the search string (and therefore are best matches) are likely to be clipped from the result by `max_items`.

Having used many search boxes in other web apps that do exactly that, I can attest that it is annoying. But if we regard this feature as fringe, maybe we don't care.

@dboyer, do we care?

opentreemap/treemap/tests/test_views.py
```diff
@@ -1955,3 +1955,45 @@ def test_instance_list_only_public(self):
...
+            self.assertIn('username', user_dict)
+            self.assertEqual(user.pk, user_dict['id'])
+            self.assertEqual(user.username, user_dict['username'])
+
+    def test_full_results(self):
+        users_list = users(make_request(), self.i2)
+        self.assert_users_in_list(users_list, self.matt)
+
+    def test_filtering(self):
+        users_list = users(make_request(), self.i1)
```
RobinIsTheBird: Seems to me that using this request in `test_full_results` would spare having to repeat it here and in `test_max`.

RobinIsTheBird:
:+1:

maurizi:
@RobinIsTheBird I've improved the sorting of results, can you take another look?

RobinIsTheBird:
:+1:

Issue #2871, opened 2016-11-23
Custom fields appear do not appear on the tree detail page by default


dboyer:
When adding a new custom field, the user currently needs to:

1. Add the field on the custom fields page (may include setting permissions)
2. Set access permissions via the Roles page if not completed on the Custom Fields page
3. Go to the Data Fields page and set the fields to display on the detail page 

I would like to have the fields display by default. The user can go to the Data Fields page to set them to not display if desired.

dboyer:
Now I can't reproduce this and they are appearing by default.

RobinIsTheBird:
@dboyer, I'll try to reproduce it.

RobinIsTheBird:
I could not reproduce this.

My local database is brand new, so there wouldn't be any caching on the server.

@dboyer and @jwalgran, what do you suggest we do with this issue? Should we close it?

jwalgran:
I was able to reproduce this on staging by following these steps

- Create a new instance centered on Los Angeles
- Add a decimal field to the Tree model
- Update the permissions on the field to make it writable by everyone

The field was not checked on the `/{instance}/management/field-configuration/` page and did not appear on the tree detail page.

RobinIsTheBird:
@dboyer, here is what I found, and what the PR I've made does.

I found code to enable udfs by default, but it was incorrect. It only enabled a newly created udf by default if one other udf already in existence had already been manually enabled.

I found no evidence of code to set permissions to make the new udf visible to _any_ role. I did find code to restrict it to be invisible to all roles, with a comment emphasizing that intent.

My PR corrects the udf enabling code to work the first time a udf is created, and assigns udf permissions per role to be the same as the role's default permission level. So udfs will be immediately visible for all our default roles - administrator, editor, and public.

There is presently no way (afaik) for an administrator to change the default permission level of a role. I think we now have enough code dependent on that default permission level (visible udfs, add/delete, external link, modeling...) that we should consider creating an issue to make it settable.

maurizi:
@dboyer @RobinIsTheBird 

The intended behavior for the Roles page was that newly added fields are by default Invisible.

When we added the data fields page, we decided to make things be enabled there by default, so that map owners would not have to go to _2_ places to enable a newly added field.

If we want to change the behavior of what happens on the Roles page, I think that should be done as a separate issue that does not block the release, as there is more to be done than just changing the code to use the default permissions (need to change help text, consider removing yellow notification bubbles, etc.)

[...]

PR #2874, opened 2016-11-25
Display new custom field

RobinIsTheBird:
- Enable new custom field by default
- Use role default field permissions instead of NONE for all the roles
- Fix a test to create stewardship udfs, to be consistent with instance config

Connects to #2871 
Depends on OpenTreeMap/otm-addons#1342


RobinIsTheBird:
@azavea-bot rebuild

maurizi:
:+1:

Issue #2876, opened 2016-11-29
Change wording on management page


dboyer:
Change "can delete tree" to "can delete any tree"
Change "can delete photo" to "can delete any photo"

Users always have the ability to delete their own photos and trees, but the wording "can delete tree" and "can delete photo" might confuse people and think they need to grant that privilege for people to have that ability. 

This could also be hover text if that seems like a better option.

PR #2881, opened 2016-11-29
Update delete permission labels

jwalgran:
Users may always delete objects they have created, regardless of their permissions. I have updated the names of the delete permissions to clarify that granting these permissions will give members of the role the ability to delete any object of that type, not only ones that they have created.

---

##### Testing

Create or use an instance that has all green infrastructure types enabled, and use a second instance that has LA infrastructure variants enabled.

Verify that the labels on all "delete" permissions include the word "any."

screen shot 2016-11-29 at 1 00 11 pm

---

Connects to #2876


RobinIsTheBird:
:+1: 

The test failures are due to braindead PR builder scripting. Tests pass on my localhost.

Issue #2863, opened 2016-11-17
Add backend to support last updated by (user) search


RobinIsTheBird:
This entails:
 - Adding a new `updated_by` column to the `MapFeature` table with a foreign key to `User` (same as treemap.Audit.user, which it denormalizes)
 - Change the code to update `updated_by` in all the places we currently update `updated_at`
 - Create a data migration to backfill `updated_by` from the Audit table (look at data migration for `updated_at` for inspiration)

PR #2882, opened 2016-11-29
Backend for filter by updated-by

RobinIsTheBird:
Changes:
* Add `updated_by` field, and make it not trackable
* Migration to add the column to the db
* Command in lieu of a migration to populate the column
* `treemap.model` updates `updated_by` along with `updated_at`
* Tests for the population command
* Tests for updating `updated_by`
* Minor refactor from references to `always_writable`
  to `bypasses_authorization`
* Minor test changes to facilitate `test --keepdb`

--

Connects to #2863

### DIFF
--- a/opentreemap/treemap/audit.py
+++ b/opentreemap/treemap/audit.py
@@ -558,10 +558,10 @@ class UserTrackable(Dictable):
         Subclasses of UserTrackable should return their own field names
         unioned with those of their immediate superclass.
         '''
-        # updated_at is "metadata" and it does not make sense to
-        # redundantly track when it changes, assign reputation for
-        # editing it, etc.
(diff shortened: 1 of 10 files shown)


RobinIsTheBird:
@maurizi, I don't think these queries can be simplified. If I remove `a.id` from the `GROUP BY`, postgresql gives me errors about needing `a.user_id` and `a.updated` to be in an aggregate or in the `GROUP BY`. And if I add them to the `GROUP BY`, it still tells me they need to be in an aggregate or in the `GROUP BY`. I guess I don't understand `DISTINCT ON` as well as I thought I did, because those errors in response to those changes make no sense to me, and I can't find an explanation that actually explains them with any search string I can think of.

maurizi:
👍

PR #2883, opened 2016-11-29
cleanup on the modeling work

designmatty:
### Overview:
- updated the label for _all trees_. Now reads _Annual mortality (%) for all trees_
- additionally updated the input-addon label to just a _%_ to mimic the table in the customize by species modal.
- added some spacing between the two Tree Mortality Rate options
- Added a _species_ table heading to the custom species modal.
- Moved the _delete button_ to the last column of the table and styled it.
- added a max-height and scroll to the _add a species dropdown_
- simplified the double row table header into a single row.
- cleaned up the scenarios menu bar a bit.

#### Testing
In order to test, pull down this branch and the branch from this PR https://github.com/OpenTreeMap/otm-addons/pull/1349

### DIFF
--- a/assets/css/sass/partials/pages/_modeling.scss
+++ b/assets/css/sass/partials/pages/_modeling.scss
@@ -103,7 +103,7 @@ $blue-to-red: #2791c3 10%,
     }
 
     #view-switcher {
-        padding-top: 5px;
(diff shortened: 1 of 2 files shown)


RickMohr:
:+1: Thanks!

Issue #2841, opened 2016-11-03
The default migration rules for threaded comments create comments with submit date set to the migration date, not the comment date


jwalgran:
An example from a customer migration

Fixture record

```
{
    "pk": 2,
    "model": "threadedcomments.threadedcomment",
    "fields": {
        "comment": "Tree died due to an attack of small cedar aphid and has been turned into a sculpture",
        "date_submitted": "2012-11-27 10:37:07",
        ... other fields ...
    }
}
```

Database row after migration

```sql
select id, comment, submit_date from django_comments;

1 | Tree died due to an attack of small cedar aphid and has been turned into a sculpture | 2016-11-01 20:03:58.164793+00
```

PR #2889, opened 2016-12-01
Fix comment handling in migrator

jwalgran:
Threaded comments were being migrated incorrectly. The `submit_date` was being set to the date of the migration, not the original comment submit date. The solution is to ensure that the renaming of `date_submitted` (which is the name used in the OTM1 fixtures) to `submit_date` is configured in the standard migration rules for both comments and threadedcomments, and that the mapping old name to new name is correctly defined.

I have opted to swallow a character encoding exception raised when
attempting to print a skipped migration warning message:

- The comment text does not provide extra information

- We do not anticipate using the OTM1 migrator in the future. Properly
  debugging the character encoding failure is not a good use of
  development time.

---

##### Testing

I tested this by:

- Creating abbreviated `config.json`

```json
{
    "user_fixture": "local/fixtures/users.json",
    "plot_fixture": "local/fixtures/plots.json",
    "threadedcomment_fixture": "local/fixtures/threadedcomments.json"
}
```

- Running through a migration, ensuring that the content type fixture is migrated before running the migrator with the abbreviated `config.json`

- Validating that the comments have the correct `submit_date` by running this query

```sql

SELECT object_pk AS feature_id, user_id, user_name, left(comment,20) AS comment, submit_date
FROM django_comments
WHERE id IN (
  SELECT otm2_model_id
  FROM otm1_migrator_otm1commentrelic
  WHERE migration_event_id = (SELECT MAX(id) FROM otm1_migrator_migrationevent)
  AND otm2_model_name = 'threadedcomment'
)
ORDER BY submit_date;

feature_id | user_id | user_name |       comment        |      submit_date
------------+---------+-----------+----------------------+------------------------
 2913077    |    5644 |           | Tree died due to an  | 2012-11-27 16:37:07+00
 2913075    |    5644 |           | Tree came down night | 2012-11-28 17:19:58+00
 2913769    |    5647 |           | I don't know the spe | 2012-11-28 17:22:51+00
 2913076    |    5644 |           | Tree came down night | 2012-11-28 17:26:22+00
 2913075    |    5644 |           | the above comment is | 2012-11-28 17:27:50+00
...
```

---

Connects to #2841

### DIFF
--- a/opentreemap/otm1_migrator/migration_rules/standard_otm1.py
+++ b/opentreemap/otm1_migrator/migration_rules/standard_otm1.py
@@ -185,7 +185,7 @@ MIGRATION_RULES = {
                          'contenttype': 'content_type'},
         'common_fields': {'comment', 'is_public',
                           'ip_address'},
-        'renamed_fields': {'submit_date': 'date_submitted'},
+        'renamed_fields': {'date_submitted': 'submit_date'},
         'removed_fields': {'is_removed', 'user_name', 'user_email',
                            'user_url', 'site', 'object_pk'},
     },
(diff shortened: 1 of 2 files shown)


RickMohr:
+1 Clear writeup, good strategy.

PR #2891, opened 2016-12-06
Add safety code in requires_permission decorator

RickMohr:
Prevents a crash when trying to load an instance you aren't authorized to access.


maurizi:
:+1:

PR #2893, opened 2016-12-08
Support for modeling module

RobinIsTheBird:
Replace FeatureNotEnabledException by PermissionDenied,
which django magically translates into a 403 Forbidden,
and returns the 403.html template.

--

Connects to OpenTreeMap/otm-addons#1359
Depends on OpenTreeMap/otm-addons#1371
Depends on OpenTreeMap/otm-website#146


RobinIsTheBird:
@azavea-bot rebuild

Issue #2852, opened 2016-11-15
Ungraceful failure creating tree when you have "add plot" but not "add tree" permission


RickMohr:
1. Configure a role with "Add planting site" permission but not "Add tree" permission
2. Using that role, try to create a tree. You're allowed to go all the way and click "Done", but then we fail with "Failed to add feature" toast.

What would be more graceful?
* Cadillac -- change UI to "Add an empty planting site", and offer no tree-specific fields
* Cheaper -- skip (or grey out) step 2 if user lacks "Add tree" permission

As we consider this we could also think about the requested feature of an explicit UI path to adding an empty planting site (vs. now where you just leave out the tree data).

RobinIsTheBird:
Taking the cheaper solution. The Cadillac solution is blocked on a spec and prioritization for the "Add an empty planting site" enhancement.

RobinIsTheBird:
Do we have a design for how this option should appear in the current UI design, or do we need a little of @lederer 's time?

dboyer:
I believe this was only whiteboarded and we don't have a wireframe. I'll get that.

lederer:
### Sidebar steps

When adding a planting site, I propose:
- Replace all "tree" strings with "planting site" throughout all 3 steps, obviously. Note: Not "empty planting site", but simply "planting site". Brevity counts.
- Retain the second step, but its only fields should be `planting site width` and `planting site length`.
- In step 3, the "After I add this planting site…" options should be:
  - "Add another planting site"
  - "Continue editing this planting site"
  - "I'm done"

### "Add" controls

Since the top nav add control will be deprecated in Spruce Up, I recommend we not get too fancy with changing it. I think the proposal below should be easily implementable.

![add menus](https://cloud.githubusercontent.com/assets/128699/22798014/66d7b37c-eece-11e6-873e-476df0eae4ed.png)


Holler if I'm missing or misunderstanding anything.

jwalgran:
@lederer Thanks for these mockups. The development team discussed them yesterday and did have one question.

Throughout the application, instead of hiding options, we have preferred to disable them, and show a tooltip explaining why they are disabled. Here is and example of an Add Tree button at the right of the map page

![screen shot 2017-02-10 at 9 26 27 am](https://cloud.githubusercontent.com/assets/17363/22834669/217f4394-ef73-11e6-9914-ed31ad67fbdc.png)

[...]

PR #2903, opened 2016-12-20
User can create plot but not tree

RobinIsTheBird:
When the user has permissions to create a plot but not to create a tree,
remove the second step of the add plot/tree wizard.

--

Connects to #2852

### DIFF
--- a/opentreemap/treemap/views/misc.py
+++ b/opentreemap/treemap/views/misc.py
@@ -18,7 +18,7 @@ from django.template import RequestContext
 
 from stormwater.models import PolygonalMapFeature
 
-from treemap.models import User, Species, StaticPage, Instance
+from treemap.models import User, Species, Tree, Plot, StaticPage, Instance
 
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/views/misc.py
```diff
@@ -71,6 +71,8 @@ def index(request, instance):
 
 
 def get_map_view_context(request, instance):
+    def make_dummy_tree():
```
maurizi: I would throw a comment above this method explaining _why_ the Map page needs a dummy tree variable

RobinIsTheBird:
Going to do the full-blown solution.

jwalgran:
When you say "the full blown solution" are you referring to implementing this option that Rick presented in #2852, or something different?

> Cadillac -- change UI to "Add an empty planting site", and offer no tree-specific fields

RobinIsTheBird:
@jwalgran yes, the #2852 "Cadillac" solution.

Issue #1818, opened 2014-12-01
Edits page loads sporadically


ahinz:
Sometimes loads, sometimes 502 (50%):
https://www.opentreemap.org/edmonton/edits/

Almost never loads:
https://www.opentreemap.org/edmonton/edits/?page=100

Really never loads:
https://www.opentreemap.org/edmonton/edits/?page=200

maurizi:
Helpful blog article about ways to speed up pagination: https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/

I think we would want to use what it describes as "Keyset pagination" - on each edits page, we would record the lowest `audit.id` and use it as a page parameter for the "Next" button, and then future queries would add a `WHERE audit.id < prev_page_lowest_id` clause, instead of the `OFFSET` that is currently added by slicing the `QuerySet`

PR #2904, opened 2016-12-20
Improve performance of instance and user edits pages

maurizi:
The edits page had exceptionally poor performance when paging beyond the first
few pages, to the point that the requests would time out when attempting to
reach pages past the first dozen or so.

This is primarily caused by the complex query we use to ensure that users only
see edits for fields they have permission to view, but was greatly exacerbated
by the use of `OFFSET` to get pages beyond the first, which requires the
database to retrieve and discard all rows for all previous pages.

By relying on the fact that we show edits in a deterministic order (newest to
oldest), we can switch from passing a page parameter to passing the starting
`Audit.id` and eliminate the `OFFSET`, making performance on each page roughly
equal.

One downside of this is that we need to keep track of the previous page
starting positions for every previous page, which means the URL can grow
to be rather long, but this seems like an OK trade-off to me.

While I was changing the code, I also cleaned up the API for `get_audits` and `get_audits_params`, as the number of parameters being returned from `get_audits_params` was getting excessive.

Connects to #1818

### DIFF
--- a/opentreemap/treemap/lib/user.py
+++ b/opentreemap/treemap/lib/user.py
@@ -3,8 +3,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import division
 
-import urllib
-
 from django.db.models import Q
 
 from treemap.audit import Audit, Authorizable, get_auditable_class
@@ -21,12 +19,14 @@ def _instance_ids_edited_by(user):
                         .distinct()
 
 
-def get_audits(logged_in_user, instance, query_vars, user, models,
-               model_id, page=0, page_size=20, exclude_pending=True,
-               should_count=False):
-    start_pos = page * page_size
-    end_pos = start_pos + page_size
+PAGE_DEFAULT = 20
+ALLOWED_MODELS = get_filterable_audit_models()
 
+
+def get_audits(logged_in_user, instance, query_vars, user=None,
+               models=ALLOWED_MODELS, model_id=None, start_id=None,
+               prev_start_ids=[], page_size=PAGE_DEFAULT, exclude_pending=True,
+               should_count=False):
     if instance:
         if instance.is_accessible_by(logged_in_user):
             instances = Instance.objects.filter(pk=instance.pk)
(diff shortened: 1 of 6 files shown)


opentreemap/treemap/lib/user.py
```diff
@@ -21,12 +19,14 @@ def _instance_ids_edited_by(user):
...
-def get_audits(logged_in_user, instance, query_vars, user, models,
-               model_id, page=0, page_size=20, exclude_pending=True,
-               should_count=False):
-    start_pos = page * page_size
-    end_pos = start_pos + page_size
+PAGE_DEFAULT = 20
+ALLOWED_MODELS = get_filterable_audit_models()
+
 
+def get_audits(logged_in_user, instance, query_vars, user=None,
```
maurizi: Nice catch! Fixed in https://github.com/OpenTreeMap/otm-core/pull/2904/commits/05859cc71145e7f874d79c2ea6d80b117b75e922 (with unit tests!)

opentreemap/treemap/lib/user.py
```diff
@@ -112,20 +112,38 @@ def get_audits(logged_in_user, instance, query_vars, user, models,
...
+    # By relying on the fact the our list is ordered by primary key from newest
+    # to oldest, we can rely on the index on the primary key, which is faster.
+    if start_id is not None:
+        audits = audits.filter(pk__lt=start_id)
+
     total_count = audits.count() if should_count else 0
-    audits = audits[start_pos:end_pos]
+    audits = audits[:page_size]
+
+    # Coerce the queryset into a list so we can get the last item on the page
```
maurizi: I added a comment about this in https://github.com/OpenTreeMap/otm-core/pull/2904/commits/05859cc71145e7f874d79c2ea6d80b117b75e922

Basically - QuerySets support indexing in general, but not negative indexing - so if we want to get the last item in the query set (which corresponds to the last row in the edits table on the page), we need to coerce the QuerySet into a list, which does support negative indexing.

maurizi:
@jwalgran This should be ready to test again.

Issue #2902, opened 2016-12-20
Advanced search headers are not sticky on mobile


maurizi:
On mobile, when scrolling down the expanded advanced search view, the "Search" and "Reset" are supposed to become fixed at the top.  Instead, they scroll out of view.

Issue #2901, opened 2016-12-20
links missing from tree popup


RobinIsTheBird:
Click a dot. The popup is missing the links.

RobinIsTheBird:
Blocks #2823

Issue #2900, opened 2016-12-20
x missing from add a resource


RobinIsTheBird:
If you start to add a resource, the sidebar is missing the "x" for dismissing it.

Issue #2869, opened 2016-11-21
Styling issue with register button on mobile


maurizi:
On the login screen, the Register button needs additional `margin-top`, and is off-center horizontally (compare to Login btn above it):
![screen shot 2016-11-21 at 2 18 24 pm](https://cloud.githubusercontent.com/assets/128699/20497885/e467984e-aff8-11e6-9a06-791e73967d4c.png)

PR #2906, opened 2016-12-20
Style fixes

maurizi:
Fixes a variety of small design bugs

Connects to #2859 
Connects to #2902
Connects to #2900
Connects to #2869
Connects to #2901


RobinIsTheBird:
In my testing, #2902 is not fixed. Everything else appears to be fixed.

Issue #2897, opened 2016-12-15
Quick Edit button switches to edit mode even when the user is not logged in


maurizi:
Steps to reproduce:
 - On the map page, click on a planting site when not logged in
 - Click "Quick Edit" in the sidebar
 - A window will popup asking if the user is sure they want to leave, and the form is switched to edit mode

What should happen:
 After clicking "Quick Edit" when not logged in, the user should be redirected to the login page.  No modal should appear asking the user if they wish to leave the page, and the form should not switch to edit mode.

RobinIsTheBird:
Blocked until I can get a second pair of developer eyes on the problem.

PR #2907, opened 2016-12-20
Fix quick edit for logged out user

RobinIsTheBird:
Go directly to the login page, without entering edit mode.

--

Connects to #2897

### DIFF
--- a/opentreemap/treemap/js/src/lib/inlineEditForm.js
+++ b/opentreemap/treemap/js/src/lib/inlineEditForm.js
@@ -270,17 +270,27 @@ exports.init = function(options) {
     $(window).on('submit', form, function(event) { event.preventDefault(); });
 
     // Merge the major streams on the page together so that it can centrally
-    // manage the cleanup of ui forms after the change in run mode
-    actionStream.plug(editStream);
-    actionStream.plug(saveStream);
-    actionStream.plug(cancelStream);
-    actionStream.plug(externalCancelStream);
-    actionStream.plug(saveOkStream.map('save:ok'));
-    actionStream.plug(responseErrorStream.map('save:error'));
-    actionStream.plug(modeChangeStream);
-    actionStream.onValue(hideAndShowElements, editFields, eventsLandingInEditMode);
+    // manage the cleanup of ui forms after the change in run mode.
+    //
+    // Note that if the user is not logged in, the buttonEnabler removes the
+    // data-class attribute. In that condition, editing is impossible,
+    // so avoid plugging edit related streams into the actionStream.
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/js/src/lib/inlineEditForm.js
```diff
@@ -19,6 +19,7 @@ exports.init = function(options) {
         form = options.form || section + ' form',
         $section = $(section),
         edit = options.edit || '.editBtn',
+        $edit = $(edit),
```
maurizi: I think finding the edit button when the module is loaded may be a problem for some forms like the plot detail where content/buttons can be reloaded from the server, like the plot detail page.

Not 100% sure though. @RickMohr would know more.

RickMohr: A single instantiation of an inline edit form can persist as the fields it manages are reloaded via AJAX -- in particular, the details sidebar of the main map page. For that reason the inline edit form code doesn't hang on to $ variables (other than $section, which is the container) but rather makes them when needed. 

In the details sidebar case the "edit" button is outside of the partial retrieved by AJAX so the code will probably work fine. But I would personally stick with the existing style (use `$(edit)` instead of `$edit`), and possibly add a comment for future developers.

RobinIsTheBird: @RickMohr and @jwalgran, the only place where the edit button gets dynamically loaded appears to be [in mapFeatureDetail.js](https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/mapFeatureDetail.js#L119), handling save success, which is irrelevant because if the user wasn't logged in, they would not have saved changes, never mind successfully.

Is it safe to assume that the base page is loaded before the javascript executes?

I assume that we do not care about any possible future case that could dynamically load details with the edit button during initial page load.

jwalgran: We always include our JS at the bottom of the `<body>`, so assuming the markup is loaded before scripts are parsed and run is safe.  https://github.com/OpenTreeMap/otm-core/blob/bec99c0e93495b6e65f0bc293249fd37dd045fa4/opentreemap/treemap/templates/base.html#L179-L217

RobinIsTheBird: Okay, then Mike's concern is unfounded.

@jwalgran, the review is in your hands.

RickMohr:
+1
Works well and code looks good!

RickMohr:
Don't forgot to squash your commits....

PR #2910, opened 2016-12-22
Fix search for IE 11

RickMohr:
Add polyfill for `String.startsWith()`, not supported by IE 11.

Note: I verified that search now works on IE 11.

Connects OpenTreeMap/otm-clients#328


RobinIsTheBird:
:+1:

Issue #2845, opened 2016-11-11
Use improved i-Tree region data


RickMohr:
Currently our i-Tree region boundaries fail to cover many coastal areas. Load improved boundaries from fileshare shapefile `projects/Azavea_OpenTreeMap/data/iTree_climate_zone_from_ForestService`

hadleystein1:
+1 Openlands (https://www.opentreemap.org/openlands/map/). All but 12 trees have species and diameter information.

![screen shot 2016-12-22 at 10 58 33 am](https://cloud.githubusercontent.com/assets/15219245/21431651/1dcb9888-c836-11e6-9209-0af8f33b12f5.png)

PR #2911, opened 2016-12-23
Replace i-Tree region boundaries with expanded boundaries

RickMohr:
Here's what I did:

1. Using QGIS, process the shapefile with expanded i-tree region boundaries:
   * Working directory `//fileshare/projects/Azavea_OpenTreeMap/data/iTree_climate_zone_from_ForestService`
   * Open `ClimateZone_edited_final.shp`
   * Combine polygons into multipolygons (Vector > Geometry Tools > Singleparts to Multipart)
   * Remove multipolygon with null region
   * Remove columns other than region name
   * Save in Web Mercator projection, as `itreeRegionsWebMercator.shp`

1. Load the data into a DB table:
   ```
   shp2pgsql -s 3857 itreeRegionsWebMercator.shp > itreeRegions.sql
   psql < itreeRegions.sql
   ```

1. Change the new data to use region abbreviations, and update the OTM i-Tree region table with the new boundaries:
   ```
   UPDATE itreeregionswebmercator
   SET czonesep07 = (
       CASE czonesep07
           WHEN 'Northern California Coast' THEN 'CaNCCoJBK'
           WHEN 'Central Florida'           THEN 'CenFlaXXX'
           WHEN 'Coastal Plain'             THEN 'GulfCoCHS'
           WHEN 'Inland Empire'             THEN 'InlEmpCLM'
           WHEN 'Inland Valleys'            THEN 'InlValMOD'
           WHEN 'Interior West'             THEN 'InterWABQ'
           WHEN 'Lower Midwest'             THEN 'LoMidWXXX'
           WHEN 'Midwest'                   THEN 'MidWstMSP'
           WHEN 'North'                     THEN 'NMtnPrFNL'
           WHEN 'Northeast'                 THEN 'NoEastXXX'
           WHEN 'Pacific Northwest'         THEN 'PacfNWLOG'
           WHEN 'South'                     THEN 'PiedmtCLT'
           WHEN 'Southern California Coast' THEN 'SoCalCSMA'
           WHEN 'Southwest Desert'          THEN 'SWDsrtGDL'
           WHEN 'Temperate Interior West'   THEN 'TpIntWBOI'
           WHEN 'Tropical'                  THEN 'TropicPacXXX'
       END);

   UPDATE treemap_itreeregion old
   SET    geometry=new.geom
   FROM   itreeregionswebmercator new
   WHERE  old.code = new.czonesep07;
   ```

1. Dump the OTM i-Tree region table to create a new fixture:
   ```
   opentreemap/manage.py dumpdata treemap.ItreeRegion > itree_regions_expanded.json
   ```
   and move to `assets/fixtures/itree_regions_expanded.json`

1. Add migration to load the new fixture (and reverse migration to load the old fixture)

I wanted to see if adding this migration slowed down our Django tests, so I timed running a single simple test with and without this migration. For 3 trials the times were 90-93 sec without and 92-94 sec with, so adding the migration may have had a small effect. Once we've lived with the change in production for a while we can consider removing the old boundaries (`treemap/fixtures/itree_regions.json`) so they aren't loaded when running Django tests.

TESTING:

1. Before running migrations, test with the old i-Tree region boundaries:
   * Load http://naples:6060/rhodytrees/features/659806, and note that it has no i-Tree region
   * Comment out [this line](https://github.com/OpenTreeMap/otm-core/pull/2911/files#diff-c01e5db3ba2e1e40aef84710540a3ed6R14) to disable the data migration when Django tests run
   * Run the [`ITreeRegionTest` tests](https://github.com/OpenTreeMap/otm-core/pull/2911/files#diff-831c1edd77b2d5e3bd239c6ff2195c12R482)
   * Note that all 12 tests fail -- they are all located on land, but not within any of our current i-Tree region boundaries.
   * Re-enable the migration

1. Run migrations, and test with the new i-Tree region boundaries:
   * Load http://naples:6060/rhodytrees/features/659806, and note that it now has an i-Tree region and reports ecobenefits
   * Run the [`ITreeRegionTest` tests](https://github.com/OpenTreeMap/otm-core/pull/2911/files#diff-831c1edd77b2d5e3bd239c6ff2195c12R482) - they should all pass

1. Migrate `treemap` to `0039` to restore the old region boundaries, and:
   * Load http://naples:6060/rhodytrees/features/659806, and note that it has no i-Tree region

Connects #2845


RickMohr:
waffle

RickMohr:
Another testing note -- in most cases, setting a default i-Tree region on a treemap only affects the ecobenefits displayed for individual trees. If the treemap intersects an i-Tree region we use that region to compute the ecobenefit totals displayed on the map page -- for the totals we only consider individual tree locations if the treemap intersects more than one region.

Issue #2914, opened 2017-01-04
Improve refactor of image upload module


RickMohr:
* Move to a separate module code that is specific to the map feature detail page
* Now that we can handle any file type, change various names to remove the word "image"
* When uploading a map feature photo, show errors in the modal rather than on the page

RickMohr:
This was completed

PR #2916, opened 2017-01-05
Improve generic file upload panel and its uses

RickMohr:
* Rename `upload_image.html` to `upload_file.html`
* Rename `imageUploadPanel.js` to `uploadPanel.js`
* Split out `imageLightbox.js` from `uploadPanel.js`

Improve upload panel template
* If modal element's ID is unspecified, use `upload-panel`
* Remove "image" from variable names

Improve uses of the upload panel
* Remove unnecessary `id` on button which opens the modal
* Make sure upload panel title is translated
* Use default panel ID rather than a custom ID

Also:
* On map feature detail pages show image upload errors using our standard pink well. (I chose not to change the display location to the modal because that would require a large back end refactor.)
* When changing bounds, show a better error message if a non-JSON file is uploaded

Connects #2914


RickMohr:
@azavea-bot rebuild

RickMohr:
@azavea-bot rebuild

RickMohr:
@jwalgran, I think we can ignore the failing tests from the PR builder, as the tests pass on Travis.

It looks like webpack is confused in the PR builder VM. I destroyed and recreated the VM but another PR got built before mine, which I believe re-introduced the problem.

opentreemap/treemap/templates/treemap/partials/photo_carousel.html
```diff
@@ -46,5 +46,5 @@
 <a class="carousel-control right" href="#photo-carousel" data-slide="next"><i class="icon-right-circled"></i></a>
 {% endif %}
 {% if error %}
-<div class="text-danger">{{ error }}</div>
+<div class="alert alert-danger js-image-upload-error">{{ error }}</div>
```
jwalgran: Should the class here be `js-image-upload-error` or `js-upload-error`? Other places in this diff I see us deleting markup with the `js-image-upload-error` class. When I grep the codebase for `js-image-upload-error` this is the only reference I find.

PR #2922, opened 2017-01-06
Change species thumbprint if a species is deleted

RickMohr:
While you can't currently delete a species via the UI, you'll be able to someday. And we sometimes need to delete them behind the scenes. Without this change, the previous species list is not removed from browser-local storage so the deleted species is still shown.


maurizi:
👍 Once the lint is fixed

Issue #1975, opened 2015-03-16
Eco-benefits not updated after species import


RickMohr:
1) Import a species with no i-Tree code.
2) Make a tree with that species. Note it has no eco-benefits.
3) Re-import the species, this time with an i-Tree code.

Expected: tree shows eco-benefits
Actual: tree still shows no eco-benefits

The problem is that the ecoservice is using stale cached data. We can fix it by hitting its `invalidate_cache` endpoint. But that's tricky in a multi-server environment.

jwalgran:
Using the pub/sub built into redis may help us. When we import species, we can publish a message that all servers will receive and react to by invalidating their local cache.

jwalgran:
http://redis.io/topics/pubsub

RickMohr:
Moving to priority backlog because the sales team keeps hitting this.

RickMohr:
Using pubsub looks straightforward, though it means running a persistent listener such as the thread in [this example](https://gist.github.com/jobliz/2596594)). We'd have to be confident the listeners would remain up, or else implement some kind of health check / restart logic.

Another option:
* add `instance.itree_override_rev`
* check it before any ecoservice call, and hit `invalidate_cache` if necessary

@jwalgran, thoughts?

jwalgran:
I'd rather keep things simple and have the ecoservice check a `*_rev` value. It's proven to be both simple and successful in other contexts.

I agree that we need a more specific cache key. Is `itree_override_rev` too specific a name? Looking at the invalidate cacheFunction, maybe not. https://github.com/OpenTreeMap/otm-ecoservice/blob/ef31e3b4c53a203abd4057b53eac6d4329598bd7/ecorest/cache/main.go#L32-L64

RickMohr:
Proposal:

* Note that the ecoservice caches i-Tree code overrides for all instances, and that invalidating the cache reloads overrides for all instances.
* We need a single `itree_code_override_rev` cache buster for all instances.
* There's currently no database table to store such a global value, so add a new model `treemap.Global` with an `itree_code_override_rev` field.
* When a Django app launches, store the current value of `Global.itree_code_override_rev` in a static `itree_code_override_rev` variable.
* When committing any `ITreeCodeOverride` change, increment `Global.itree_override_rev`.
[...]

PR #2923, opened 2017-01-09
On plot detail page always show i-Tree code to admins

RickMohr:
When debugging ecobenefit problems it's always useful to see the i-Tree code, and it doesn't hurt anything to show it as "None".


RickMohr:
On second thought I think the existing code is fine, so closing this. The real issue is
https://github.com/OpenTreeMap/otm-core/issues/1975, which I've moved to the priority backlog.

PR #2924, opened 2017-01-09
fixed buttons overlaying in modeling scenario editing

designmatty:
fixes https://github.com/OpenTreeMap/otm-addons/issues/1382

screen shot 2017-01-09 at 2 46 06 pm


RickMohr:
+1

PR #2925, opened 2017-01-09
fixed diameter inputs so in vs cm doesnt break layout

designmatty:
fixes https://github.com/OpenTreeMap/otm-addons/issues/239

I added a min-width to the input and made the padding small on the input-addon. The input will be big enough for both imperial and metric.

screen shot 2017-01-09 at 2 52 34 pm


RickMohr:
+1

PR #2927, opened 2017-01-09
fixes modeling toolbar button height and various sidebar issues

designmatty:
- fixes https://github.com/OpenTreeMap/otm-addons/issues/1378
- fixes https://github.com/OpenTreeMap/otm-addons/issues/1390

Button now has the border below it consistent with the toolbar:
screen shot 2017-01-09 at 3 51 20 pm

The sidebar will now span the full height available to it:
screen shot 2017-01-09 at 3 48 49 pm

Additionally, the scenario summary is no longer hidden below the planting areas legend. I don't believe this has been previously documented as an issue. It was always there, just being covered by the legend. This is also only visible when in a scenario.
screen shot 2017-01-09 at 3 49 19 pm


RickMohr:
+1
Triple yay! Thanks for finding that lost summary.

PR #2928, opened 2017-01-12
Feature/rtb/search by user

RobinIsTheBird:
Changes:

- Add foreign key handling to `form_extras` `SearchNode`
- Add HTML to express `SearchNode` foreign key fields
- `searchBar.js` translation of foreign key DOM into `otmTypeahead` options
- `otmTypeahead` now handles dynamic search without prefetch
- `search.js` adapts foreign key fields to tiler requests
- Reset resets both significant parts of the typeahead DOM
- Style fixes
- Remove a no-op bacon onValue

Complaints:

The General dropdown doesn't expand to include the typeahead
drop-down menu, which now appears to overlap the bottom of it.
Not pretty, but usable.

Not tested at all with responsive design.

--

Connects to #388

### DIFF
--- a/opentreemap/treemap/js/src/lib/otmTypeahead.js
+++ b/opentreemap/treemap/js/src/lib/otmTypeahead.js
@@ -84,6 +84,7 @@ var create = exports.create = function(options) {
         },
 
         prefetchEngine,
+        queryEngine,
         geocoderEngine,
         allDataStream,
 
@@ -102,6 +103,13 @@ var create = exports.create = function(options) {
                 }
             },
             display: 'value',
+            templates: {
+                suggestion: template
+            },
+        },
+        queryOptions = {
+            source: null,
+            display: options.display,
             templates: {
                 suggestion: template
             },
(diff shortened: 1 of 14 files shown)


opentreemap/treemap/templatetags/form_extras.py
```diff
@@ -464,6 +468,27 @@ def resolve_label_and_identifier(self, context):
...
 
+    def _fill_in_typeahead(self, field, field_name, model):
+        relation_lookup_infos = {
+            'treemap.models.User': {
+                'url': reverse(
+                    'users', kwargs={
+                        'instance_url_name': model.instance.url_name
+                    }
+                ) + '?q=%Q%',
+                'placeholder': _('Please type a user-name'),
```
maurizi: This should be `username` and not `user-name` to match what we have used elsewhere in the site

opentreemap/treemap/js/src/lib/otmTypeahead.js
```diff
@@ -84,11 +84,13 @@ var create = exports.create = function(options) {
...
 
         prefetchEngine,
+        queryEngine,
         geocoderEngine,
         allDataStream,
 
         typeaheadOptions = {
-            minLength: options.minLength || 0
+            minLength: options.minLength || 0,
+            initialize: false
```
maurizi: Why is this needed?

opentreemap/treemap/js/src/lib/searchBar.js
```diff
@@ -111,6 +112,25 @@ function initSearchUi(searchStream) {
         geocoderBbox: config.instance.extent
     });
 
+    var $query_typeaheads = $('.search-right .autocomplete-group');
+    $query_typeaheads.each(function () {
+        var $textInput = $(this).find('[type="text"]');
+        var $hiddenInput = $(this).find('[type="hidden"]');
+        // Can't use # notation because id contains a dot
+        var textId = '[id="' + $textInput.attr('id') + '"]';
+        var hiddenId = '[id="' + $hiddenInput.attr('id') + '"]';
```
maurizi: Given the way these are used in `otmTypeahead.js` (immediately wrapped in `$(...)`), you don't need to do this at all, you can just pass these in directly like so:
```javascript
otmTypahead.create({
    input: $textInput,
    hidden: $hiddenInput,
    ...
});
```

PR #2936, opened 2017-01-13
Adjust styles for "top areas"/"all areas" modeling slider

RickMohr:
Connects OpenTreeMap/otm-addons#1261


designmatty:
Only one change to make:

`assets/css/sass/partials/pages/_modeling.scss` on `line 383` add a `focus` state in addition to the `hover` state like the following:

```
&:hover, &:focus {
  .options-toggle { ...

}
```

This will prevent the popup from disappearing when you are actively moving a slider and accidentally move your mouse outside the confines of the popup.

RickMohr:
@designmatty, with your suggested change the popup still disappears for me when moving the mouse outside the popup while dragging.

PR #2938, opened 2017-01-13
Use universal_rev to compute ecobenefits etag instead of Audit table

RickMohr:
Connects #328


maurizi:
:+1: Code looks good, haven't tested. Make sure to clean up the lint before merging

PR #2942, opened 2017-01-18
Allow multipolygons in map boundary GeoJSON

RickMohr:
Testing:
1. Create an instance centered on Pleasanton, CA (or use `/alamedacounty` if you have it)
1. Change the instance bounds by uploading `users/rmohr/data/changeBounds/alameda.geojson` from the fileshare (which contains MultiPolygons).
1. Verify that you can only add a tree in places matching:
image


RobinIsTheBird:
:+1:, however,

But see issues #2943 and #2944.

PR #2950, opened 2017-01-25
Invalidate ecoservice cache when an i-Tree code override is created

RickMohr:
* Note that the ecoservice caches i-Tree code overrides for all instances, and that invalidating the cache reloads overrides for all instances.
* Store in Redis an `itree_code_override_rev` cache buster for all instances
* When a Django app launches, store the current value of the cache buster in a static variable
* Create signal handlers so that any `ITreeCodeOverride` change increments the cache buster in Redis
* When making any ecoservice call, if the cache buster is stale, invalidate the ecoservice cache

Here's an example of how this works with 2 app servers, A and B:
1. A calls `_init_ecoservice_cache_handling`, setting the Redis rev and `my_itree_code_override_rev` to 1.
1. B calls `_init_ecoservice_cache_handling`, reading the Redis rev and setting `my_itree_code_override_rev` to 1.
1. A saves an `ITreeCodeOverride`, causing `_increment_itree_code_override_rev` to set the Redis rev to 2.
1. B makes an ecoservice call, which calls `invalidate_ecoservice_cache_if_stale`. Since B's `my_itree_code_override_rev` is still 1, B calls `invalidate_cache` on its ecoservice and sets `my_itree_code_override_rev` to 2.
1. A makes an ecoservice call, which calls `invalidate_ecoservice_cache_if_stale`. Since A's `my_itree_code_override_rev` is still 1, A calls `invalidate_cache` on its ecoservice and sets `my_itree_code_override_rev` to 2.

Testing:
1. Create a new instance centered on Seattle
1. Add a tree in downtown Seattle with species Acer rubrum and diameter 11 inches, and note your total ecobenefits
1. Restart celery
1. Do a bulk import of users/rmohr/data/importer/acerRubrumOverride.csv
1. Reload the map page -- the total ecobenefits should have changed

Connects #1975

### DIFF
--- a/opentreemap/treemap/ecocache.py
+++ b/opentreemap/treemap/ecocache.py
@@ -6,8 +6,9 @@ import hashlib
 
 from django.conf import settings
 from django.core.cache import cache
+from django.db.models.signals import post_save, post_delete
 
-from treemap.models import Plot
+from treemap.models import Plot, ITreeCodeOverride
 
 # Cache the results of plot counts and tree ecobenefit summary requests.
 #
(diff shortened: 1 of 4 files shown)


opentreemap/treemap/ecocache.py
```diff
@@ -76,3 +77,56 @@ def _get_key(prefix, filter):
...
+
+    cache.incr(_ITREE_CODE_OVERRIDE_REV_KEY)
+
+
+def invalidate_ecoservice_cache_if_stale():
+    from treemap import ecobackend
+    global my_itree_code_override_rev
+    cached_rev = cache.get(_ITREE_CODE_OVERRIDE_REV_KEY)
+    if my_itree_code_override_rev != cached_rev:
+        __, err = ecobackend.json_benefits_call('invalidate_cache', {})
```
jwalgran: It does not look like this solves the issue of invalidating the cache in all the load balanced ecoservice instances. I don't think there is a way to implement this without altering the ecoservice to read the cache key.

RickMohr: I think it does, in our current scenario where the ecoservice runs on the app server. I updated the commit and PR comment with an example of how it works.

I agree that having the ecoservice read the cache key is a more robust solution, handling configurations where ecoservice servers are separate from app servers.

jwalgran: I see. I did not read closely enough into what json_benefits_call does. I made an incorrect  assumption that it was making a request to through the load balancer. Sorry for not looking at the value of ECO_SERVICE_URL more closely.

RickMohr: I didn't look at `ECO_SERVICE_URL` either until you pointed it out -- I just assumed local. Could well have been as you assumed.

PR #2953, opened 2017-01-26
Invalidate ecoservice cache when an i-Tree code override is created

RickMohr:
* Note that the ecoservice caches i-Tree code overrides for all instances, and that invalidating the cache reloads overrides for all instances.
* Store in Redis an `itree_code_override_rev` cache buster for all instances
* When a Django app launches, store the current value of the cache buster in a static variable
* Create signal handlers so that any `ITreeCodeOverride` change increments the cache buster in Redis
* When making any ecoservice call, if the cache buster is stale, invalidate the ecoservice cache

Here's an example of how this works with 2 app servers, A and B:
1. A calls `_init_ecoservice_cache_handling`, setting the Redis rev and `my_itree_code_override_rev` to 1.
1. B calls `_init_ecoservice_cache_handling`, reading the Redis rev and setting `my_itree_code_override_rev` to 1.
1. A saves an `ITreeCodeOverride`, causing `_increment_itree_code_override_rev` to set the Redis rev to 2.
1. B makes an ecoservice call, which calls `invalidate_ecoservice_cache_if_stale`. Since B's `my_itree_code_override_rev` is still 1, B calls `invalidate_cache` on its ecoservice and sets `my_itree_code_override_rev` to 2.
1. A makes an ecoservice call, which calls `invalidate_ecoservice_cache_if_stale`. Since A's `my_itree_code_override_rev` is still 1, A calls `invalidate_cache` on its ecoservice and sets `my_itree_code_override_rev` to 2.

Testing:
1. Create a new instance centered on Seattle
1. Add a tree in downtown Seattle with species Acer rubrum and diameter 11 inches, and note your total ecobenefits
1. Restart celery
1. Do a bulk import of users/rmohr/data/importer/acerRubrumOverride.csv
1. Reload the map page -- the total ecobenefits should have changed

Connects #1975

### DIFF
--- a/opentreemap/treemap/ecobackend.py
+++ b/opentreemap/treemap/ecobackend.py
@@ -15,6 +15,8 @@ from django.contrib.gis.geos import GEOSGeometry
 
 from opentreemap.util import add_rollbar_handler
 
+from treemap.ecocache import invalidate_ecoservice_cache_if_stale
+
 import logging
 
 # By default the level for the logger will be NOTSET, which falls back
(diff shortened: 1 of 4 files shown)


RickMohr:
Recreated to target `release/2.14.0` branch.

PR #2963, opened 2017-01-30
Ignore updated_at and updated_by field values when updating plots via the API.

RobinIsTheBird:
The android app sends `updated_by`. It doesn't know it's a foreign key,
neither does the api, so it gets set to the `id` rather than the model.
This causes a `ValueError` in auditing.

The api already has a list of attributes to ignore.
`updated_by` can be safely ignored because it is set in auditing.
Same for `updated_at`. Since they are supposed to be set in auditing,
they really should both be ignored.

So ignore them.

--

Connects to https://github.com/OpenTreeMap/otm-android/issues/303
Connects to https://github.com/OpenTreeMap/otm-android/issues/304

### DIFF
--- a/opentreemap/api/plots.py
+++ b/opentreemap/api/plots.py
@@ -83,10 +83,14 @@ def update_or_create_plot(request, instance, plot_id=None):
             for key, val in request_dict[model].iteritems():
                 data["%s.%s" % (model, key)] = val
 
-    # We explicitly disallow setting a plot's tree id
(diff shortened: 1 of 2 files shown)


RobinIsTheBird:
It passed jenkins when mis-targeted at `develop`. No reason to expect it to fail now, so assigning it.

jwalgran:
I would like to change the title of this PR to the more descriptive summary "Ignore updated_at and updated_by field values when updating plots via the API."

Issue #2538, opened 2016-03-22
Eliminate exporter warning


RickMohr:
Rollbar [OpenTreeMap] production - New Warning: Unrecognized export field name (/opt/app/core/opentreemap/exporter/tasks.py:_csv_field_header_map) https://rollbar.com/Azavea/OpenTreeMap/items/1019/

The field in question that was unrecognized was `tree__plot`

Nothing bad is happening; exports are fine. But we should fix the underlying problem and not try to export that field, and the warning will go away.

PR #2968, opened 2017-01-31
Fix exporting of geometry fields for tree exports.

maurizi:
Use `Model.visible_fields(user)` instead field permissions in
exporter in order to get fields that are always writable.

This also gives us a bunch of fields we don't want, so I added some
more fields to the list to be omitted.

Connects to #2538
Connects to OpenTreeMap/otm-clients#341


RickMohr:
Looks like there is some lint and a failing test.

maurizi:
@RickMohr I fixed the lint and deleted the failing test (since the situation it tested for now can't happen).

PR #2969, opened 2017-01-31
Correct display value in last updated by search

RobinIsTheBird:
Server and Client parts.

Could not get Bloodhound prepopulation to work with a literal,
but Last Updated By now shows the currect display value.


maurizi:
:+1:

Issue #2966, opened 2017-01-31
Reduce JS bundle size using esri-leaflet-bundler


RickMohr:
Including `esri-leaflet` increased our JS download by 56K (6%). Now that we've switched to Leaflet 1.0 we can reduce that greatly using https://github.com/esri/esri-leaflet-bundler to grab just the basemap layer code.

Issue #2812, opened 2016-09-30
Update to Leaflet 1.0


RickMohr:
We are on a patched version of Leaflet 0.6.4 -- very old.
1.0 has been extremely well-tested, and has features we want (like more control over stacking order of layers). Today I tried to use a feature but failed since it was introduced in 0.7.0
We've been waiting for 2 years. It's time.

Code comments relevant to Leaflet upgrade
* [x] https://github.com/OpenTreeMap/otm-addons/blob/develop/modeling/js/src/lib/scenarioUi.js#L349
* [x] https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/lib/boundarySelect.js#L22
* [x] https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/lib/layers.js#L98
* [x] https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/lib/MapManager.js#L283
* [x] https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/lib/MapManager.js#L358

### Plugins
npm
* [x] esri-leaflet
* [x] leaflet-pip

vendor
* [x] leaflet.draw.js

shim
* [x] leaflet.bing.js
* [x] leaflet-editable-polyline.js
* [x] Google maps tiles
* [x] leaflet.utfgrid.js

jwalgran:
We may want to merge [When upgrading from Leaflet 0.6.4, resolve source code patch](https://github.com/OpenTreeMap/otm-core/issues/776) into this issue.

RickMohr:
I started working on this with R&D time -- see https://github.com/RickMohr/otm-core/commits/leaflet-1.0

Issue #776, opened 2013-10-21
When upgrading from Leaflet 0.6.4, resolve source code patch


RickMohr:
When you select a tree and click "Edit Details" on the sidebar, Leaflet crashes in `_movePopup` (src/layer/marker/Marker.Popup.js).

I applied the fix suggested in https://github.com/Leaflet/Leaflet/issues/1983 --
a Leaflet developer's comment suggests the issue is resolved in master, but the suggested fix doesn't appear there and there's no link to the actual fix.

So when we upgrade Leaflet, check if the issue is fixed by running this plunker (modified from the one in the above issue) with the new version of Leaflet:

http://plnkr.co/edit/oF4g3M

If the plunker works, just use the new Leaflet. If not, look deeper and consider reapplying the fix.

jwalgran:
Leaflet 1.0 was released today.

Issue #578, opened 2013-10-09
Do we want the "Leaflet" link on our maps?


RickMohr:
Currently there's a "Leaflet" link at lower right on our maps.

RickMohr:
I tweaked the CSS so the "Leaflet" link no longer overlaps the Google attribution.

PR #2970, opened 2017-02-01
Upgrade to Leaflet 1.0

RickMohr:
Adjust for Leaflet map pane`z-index` changes (now 200-700 rather than small integers)
- Search box on `create` page now at 900
- `.map` is the root Leaflet map element. Add `z-index: 1` (overriding Leaflet's 400) so dropdowns above it will overlay it correctly.
- Adjust z-index and position of modeling controls / info displays (they should really all be Leaflet controls, but that's outside the scope of this PR)

Put "searched boundary" polygon between base map and tiles in the map layer stacking order
* Add `base-map` pane to contain base map layer
* Add `searched-boundary` pane to contain searched boundary polygon
* Simplify code creating ESRI base map (workarounds no longer needed)

Deal with all Leaflet plugins
- `npm`
  - `esri-leaflet`
    - Still works at current version (~1.0.2). Latest is 2.0.7, but no reason to update since we don't actually use it. We could now reduce its download size, see https://github.com/OpenTreeMap/otm-core/issues/2966
  - `leaflet-pip`
    - Had version ~0.1.1
    - Now fetching from `npm` and updated to latest, 1.0.0
- `vendor`
  - `leaflet.draw.js`
    - Had version 0.2.2
    - Now fetching from `npm` and updated to latest, 0.4.9
- `shim`
  - `leaflet.bing.js`
    - Still works fine, though I'm not sure where ours came from
    - There are also these:
      - https://github.com/mertemin/leaflet-bing
      - https://github.com/gmaclennan/leaflet-bing-layer
  - `leaflet-editable-polyline.js`
    - Abandoned
    - Crashes with Leaflet 1.0
    - Removed in favor of `leaflet-draw`
  - Google Maps tiles
    - Were using `leaflet.google.js` (https://gist.github.com/crofty/2197042)
    - Switched to `Leaflet.GridLayer.GoogleMutant` from `npm`
      - Solves layer sync problem -- tree dots and basemap now pan together!!
      - Doesn't work on IE10 (but that's only 0.8% of market; already losing 4.3% IE9)
  - `leaflet.utfgrid.js`
    - replaced shim with latest version 0.2.1

Also:
  - Green infrastructure polygon editing now uses `leaflet-draw` (whose vertex boxes are BIG).
  - Add "Terrain" option, now available on Google base maps
  - `maxNativeZoom` is honored by Leaflet 1.0, so the canopy layer on `patreemap` no longer disappears at high zoom levels
  - Tweak margin on tiny "Leaflet" button at lower right of map so it doesn't overlap Google Map attribution
  - Removed `fixZoomLayerSwitch` workaround

Testing - verify that all Leaflet maps work right
  - Main map page
      - Search dropdowns (regular and advanced) correctly cover the map
      - You can still add and move a tree and bioswale
      - `patreemap` canopy layers stack and toggle correctly; canopy shows at max zoom
      - The polygon shown after searching by zip code is correctly beneath the tree dots
      - Tree dots and basemap now pan together
  - Mini map on plot detail page
  - "Create" step 3
  - Modeling
      - "distribution" polygons work right -- add, select, view area, edit

Connects #2812
Connects #776
Connects #578


opentreemap/treemap/templates/treemap/partials/map_add_resource.html
```diff
@@ -52,7 +52,7 @@
                 <a href="javascript:;" class="close cancelBtn small visible-xs-block">×</a>
             </div>
             <div class="add-step-content">
-                <div class="alert alert-info">Move the dots on the blue polygon to define the borders of your {{ term.Resource.singular.lower }}.
+                <div class="alert alert-info">"Move the squares on the blue polygon to define the borders of your {{ term.Resource.singular.lower }}.
```
jwalgran: Looks like it

assets/css/vendor/leaflet.css
```diff
@@ -240,55 +298,38 @@
...
 .leaflet-control-layers {
-	box-shadow: 0 1px 7px rgba(0,0,0,0.4);
-	background: #f8f8f9;
-	-webkit-border-radius: 5px;
-	        border-radius: 5px;
+	box-shadow: 0 1px 5px rgba(0,0,0,0.4);
+	background: #fff;
+	border-radius: 5px;
 	}
 .leaflet-control-layers-toggle {
```
RickMohr: I don't see this issue -- try `scripts/manage.sh collectstatic`.

(Note that without branding colors webpack bundles the images directly but with branding colors the dynamically-compiled SASS refers to the static image location.)

jwalgran: Running `collectstatic` resolved the issue. Thanks.

opentreemap/opentreemap/settings/default_settings.py
```diff
@@ -406,7 +406,7 @@
 
 # We have... issues on IE9 right now
 # Disable it on production, but enable it in Debug mode
-IE_VERSION_MINIMUM = 9 if DEBUG else 10
+IE_VERSION_MINIMUM = 9 if DEBUG else 11
```
jwalgran: I tested this with BrowserStack and saw the redirect working in IE 10.

We have a hard-coded version "10" in our `unsupported.html` template that we should change as well.

https://github.com/OpenTreeMap/otm-core/blob/77f662725fc4bd3ff15e8a6817f9799757db514f/opentreemap/treemap/templates/treemap/unsupported.html

RickMohr: Thanks for remembering! Fixed.

Issue #2975, opened 2017-02-06
Polygon outline shown incorrectly after modify + save + modify


RickMohr:
1. Go to a bioswale detail page
1. Edit, Move Location, drag polygon handles, Save
1. Edit, Move Location

Expected: handles show new polygon shape
Actual: handles show old polygon shape

PR #2976, opened 2017-02-06
Show polygon outline correctly after modify + save + modify

RickMohr:
* Remove marker dragging workaround now that we're on Leaflet 1.0
* Remove `InlineEditForm.onSaveAfter` in favor of the existing `InlineEditForm.saveOkStream`
* In `geometryMover.js` rename `onSaveAfter` methods to `onSaveOk`
* In `geometryMover.js` replace `onCancel` with `isCancel` argument to `disable` (this will be important when showing the area of a polygon as it is modified)
* Refactor `geometryMover.js` so that `onSaveOk` is called before `disable`, fixing #2975

Connects #2975

### DIFF
--- a/opentreemap/treemap/js/src/lib/geometryMover.js
+++ b/opentreemap/treemap/js/src/lib/geometryMover.js
@@ -19,13 +19,12 @@ function init(obj, options) {
             $editLocationButton.show();
         } else { // in display mode
             $editLocationButton.hide();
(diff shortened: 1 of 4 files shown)


RickMohr:
@maurizi this is ready for a look.

maurizi:
This is working well in all the cases I tested, and the code changes look good 👍

Issue #2905, opened 2016-12-20
Display area when drawing or editing the polygon for a green infrastructure resource


dboyer:
1. When the user draws a polygon in step 3 of adding a rain garden or bioswale, display the area of the polygon in the right side bar as it is drawn on the map.

2. When the user edits the polygon on the detail page, show the area updating as they adjust the polygon.

PR #2977, opened 2017-02-07
As GSI polygon is modified, display its area

RickMohr:
* Use `leaflet-draw` "edit vertex" events for notifications when a polygon vertex is moved
* Add `getPolygonDisplayArea` utility function (previously in modeling code)
* `polylineEditor.js` exports `areaStream`, with current area as user modifies polygon
* `polylineEditor.js` retains `initialArea` so display can be restored if polygon editing is canceled
* Use `areaStream` to show area in "Resource Detail" table
* Use `areaStream` to show area in "Add Resource" sidebar

Testing:
* On resource detail page:
  * area should update as you edit the polygon
  * area should revert when you cancel editing or "Cancel Move Location"
  * area should not revert when you save
  * multiple edit/save cycles should work correctly
* When adding a resource, area should update as you edit the polygon

Connects #2905

### DIFF
--- a/opentreemap/treemap/js/src/lib/utility.js
+++ b/opentreemap/treemap/js/src/lib/utility.js
@@ -129,6 +129,15 @@ exports.offsetLatLngByMeters = function(latLng, dx, dy) {
     return { lat: newLat, lng: newLng };
 };
 
+// TODO: Respect instance units configuration.
+// https://github.com/OpenTreeMap/otm-addons/issues/326
+exports.getPolygonDisplayArea = function(poly) {
+    var latLngs = poly.getLatLngs()[0],
+        areaSqMeters = L.GeometryUtil.geodesicArea(latLngs),
+        areaSqFeet = areaSqMeters * 10.7639;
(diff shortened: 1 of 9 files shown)


RickMohr:
@maurizi, thanks for the comments. Dealing with the small difference in server-side vs. client-side area estimates is actually harder than it looks on the surface. I looked at some real client bioswales and the estimates were either identical or off by 1 sq ft. If it's deemed important we can tackle it in a future PR.

opentreemap/treemap/js/src/lib/utility.js
```diff
@@ -129,6 +129,14 @@ exports.offsetLatLngByMeters = function(latLng, dx, dy) {
     return { lat: newLat, lng: newLng };
 };
 
+// TODO: Respect instance units configuration.
```
RickMohr: We already had https://github.com/OpenTreeMap/otm-addons/issues/326 for modeling -- I noted that it also applies to GSI polygons, and added a link in the comment.

Issue #2982, opened 2017-02-09
UI test fails intermittently


RickMohr:
The test might succeed or fail with the same code, so there's probably a timing issue.

```
FAIL: test_edit_trees_on_map (treemap.tests.ui.uitest_map.MapTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/app/core/opentreemap/treemap/tests/ui/uitest_map.py", line 198, in test_edit_trees_on_map
    self.assertEqual(tree.diameter, 32.0)
AssertionError: 124.0 != 32.0
```

PR #2988, opened 2017-02-14
Attempt to fix intermittent test failure

RickMohr:
The failure suggests that when typing "32" for diameter and hitting "Save", the new diameter is sometimes not saved. My only thought is that the "save" event is handled before the diameter calculator's JS code has finished updating everything. So, wait for the circumference value to be displayed before continuing.

Also fixed an unrelated oversight in `browse_to_instance_url`.

Connects #2982


maurizi:
Merging this for @RickMohr since he's on vacation

PR #2998, opened 2017-02-20
Move JS files used only on main map page to their own directory

RickMohr:
Also adjust webpack code to create aliases for all folders in `js/src`, not just `lib`


RickMohr:
Included in #3001 -- closing.

PR #2999, opened 2017-02-21
Works Management Permissions

RobinIsTheBird:
Add instance permission and associated data migration

--

Connects to OpenTreeMap/otm-addons#1444
Depends on OpenTreeMap/otm-addons#1450

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -100,6 +100,7 @@ def add_species_to_instance(instance):
 
 
 PERMISSION_VIEW_EXTERNAL_LINK = 'view_external_link'
+PERMISSION_ACCESS_WORKS_MANAGEMENT = 'access_works_management'
 
 
 # Don't call this function directly, call plugin.get_instance_permission_spec()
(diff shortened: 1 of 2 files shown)


RobinIsTheBird:
@azavea-bot rebuild

opentreemap/treemap/instance.py
```diff
@@ -112,6 +113,12 @@ def get_instance_permission_spec():
                              'of a tree or map feature'),
             'default_role_names': [Role.ADMINISTRATOR, Role.EDITOR],
             'label': _('Can View External Link')
+        },
+        {
+            'codename': PERMISSION_ACCESS_WORKS_MANAGEMENT,
+            'description': _('Can view or modify Works Management entities'),
+            'default_role_names': [Role.ADMINISTRATOR],
+            'label': _('Can Access Works Management')
```
RickMohr: Capitalizing as "Can access Works Management" would be more consistent with other permissions like "Can delete any tree". (Likewise maybe we should change to "Can view External Link".)

RobinIsTheBird: I would expect the Permission to appear on the Roles page for any instance that has Works Management enabled.

Since, for now, we're making Works Management universally enabled for all instances in the Development stack, it would follow that the Roles page should display it universally in that stack.

And since the default permission agreed on in the slack channel is `administrator` only, I'd expect to see it granted only to the Administrator role unless you change it.

What were you expecting, @RickMohr?

(Happy to change "access" to lower case).

RickMohr: Sorry if I wasn't clear. Just summarizing the correct behavior I was seeing. All good.

RobinIsTheBird: I now see that I capitalized the first letter of all the words because it's friend, `view_external_link` capitalizes all the words: `'Can View External Link'`.

Leaving them both as is for now. If we want greater consistency, between instance permissions and model permissions, that seems like a separate issue.

PR #3009, opened 2017-02-25
Update node-sass and sass-loader versions

jwalgran:
NOTE: I am opening this as one option for dealing with `webpack_loader` exceptions on all page loads after a transitive JS dependency was updated to no longer be compatible with node 0.12.x.

----

After updating our app servers to use node 4.x instead of node 0.12.x, webpack_loader began raising errors on page requests.

```
WebpackError:
            ModuleBuildError in
            Module build failed: Error: Missing binding /opt/app/core/node_modules/node-sass/vendor/linux-x64-46/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 4.x
Found bindings for the following environments:
  - Linux 64-bit with Node 0.12.x
This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to build the binding for your current environment.
```

Attempting to follow the instructions and adding `npm rebuild node-sass` to the provisioning script did not resolve the error. Updating the version node-sass did.

---

Required by https://github.com/OpenTreeMap/otm-cloud/pull/349


maurizi:
Do we need to update our python `libsass` dependency in tandem with this?

maurizi:
See https://github.com/OpenTreeMap/otm-core/issues/2332

RobinIsTheBird:
@maurizi, Since we can build with these changes, my thoughts are no, we don't immediately _have_ to, but we should elevate the priority of doing so for a future sprint.

maurizi:
I'm mostly worried about hard-to-notice styling differences between the CSS generated by webpack and the CSS generated by `python-libsass` (for maps without branding and with branding, respectively).

We've had bugs like that before and have taken care since to keep the versions in tandem.

jwalgran:
I added a commit to update python-libsass to 0.11.2 so that the libsass versions on both the python and node sides match

screen shot 2017-02-27 at 12 56 26 pm

screen shot 2017-02-27 at 12 57 03 pm

maurizi:
Changes look good to me (I have not tested them).

PR #3010, opened 2017-02-27
[DO NOT MERGE] Pin versions of webpack tools that depend on loader-utils

jwalgran:
NOTE: I am opening this as one option for dealing with webpack_loader exceptions on all page loads after a transitive JS dependency was updated to no longer be compatible with node 0.12.x.

`loader-utils` dropped support for node 0.12.x. We can prevent loading the incompatible version of this transitive dependency by pinning the versions of the packages that depend on it.

---

##### Testing

loader-utils@1.0.2 should not appear in the `npm ls` output after provisioning.

```
vagrant@app:/opt/app/core$ npm ls | grep loader-utils
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ └─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ └─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
```

Django pages should load without error.

---


jwalgran:
I will reopen this against the release/2.15.0 branch.

PR #3011, opened 2017-02-27
[DO NOT MERGE] Pin versions of webpack tools that depend on loader-utils

jwalgran:
NOTE: I am opening this as one option for dealing with webpack_loader exceptions on all page loads after a transitive JS dependency was updated to no longer be compatible with node 0.12.x.

---

`loader-utils` dropped support for node 0.12.x. We can prevent loading the incompatible version of this transitive dependency by pinning the versions of the packages that depend on it.

---

##### Testing

loader-utils@1.0.2 should not appear in the `npm ls` output after provisioning.

```
vagrant@app:/opt/app/core$ npm ls | grep loader-utils
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ └─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ └─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
│ ├─┬ loader-utils@0.2.17
```

Django pages should load without error.

---


RobinIsTheBird:
I don't understand why Jenkins failed with the missing binding for Node 0.12.x (found binding for Node 4.x) error, when this PR does not depend on OpenTreeMap/otm-cloud#349, which is what switches to Node 4.8.0.

Also, I notice that the other two PRs are prefaced with "Do Not Merge" and this one is not. @jwalgran, is this your preferred choice?

jwalgran:
@azavea-bot rebuild

jwalgran:
No, I would prefer to upgrade to node 4.x.

I ran into the bindings issue on my VM too. I think it is because we do not destroy the PR builder VMs between runs. I manually destroyed the app VM for the core PR building job before triggering a rebuild of this PR.

jwalgran:
We have opted to upgrade our node version rather than pin to old dependency versions.

Issue #2992, opened 2017-02-17
Add endpoint for saving a user-drawn, anonymous boundary


jwalgran:
To support filtering map features by a user-drawn area, we have decided to store the geometry in the database and reference it by auto-generated ID in filter statements.

Our tiler and ecobenefit services already support filtering by boundaries saved to the database and referenced by ID. Part of resolving this issue involves research which of these two implementations is preferable:

1) Add an `is_ad_hoc` column to the existing boundary table, store user-drawn geometries the existing table with the new column set to `True`, and modify code that loads and displays named boundaries to ignore the user-drawn boundaries.

2) Store user-drawn geometries in a separate table and modify the filter syntax, tiler, and ecobenefit services to read from this table.

If we opt to create a new table, we will want the following attributes

```
geometry (multipolygon)
created_at (datetime, autogenerated)
```

The endpoint should accept a POST request with a GeoJSON argument or body, save the geometry, and return the database ID in a JSON response

```json
{ "id": 200 }
```

Anonymous users should be able to search by user-drawn geometry, so this endpoint must not be authenticated.

RobinIsTheBird:
@jwalgran, @maurizi, and @RickMohr, If we use the `treemap_boundary` table, could we just use `category='ad hoc' instead of adding an `is_ad_hoc` boolean column?

jwalgran:
Yes, we could use Category 

Pro:
- Ad-hoc boundaries don't have categories, so that field would be unused anyway.
Con: 
- Using a "Magic" constant in the data instead of an explicit "flag" column in the table schema.

In either case, we need to change the code that assigns boundaries to instances to ignore any ad-hoc boundaries. 

If storing ad-hoc boundaries in the existing boundaries is preferable to putting them in a separate table, I have a slight preference for adding an explicit `is_ad_hoc` column.

RobinIsTheBird:
Using the same `treemap_boundary` table as other boundaries, all we would have to change are:
- otm-core `Instance#boundary_thumbprint` property, to exclude ad hoc boundaries
- otm-addons `cloud_management.instance._init_instance` where it sets `instance.boundaries`, again to exclude ad hoc boundaries.

I recommend keeping ad hoc boundaries in the `treemap_boundary` table, and designating them by an empty `name` and empty `category`.

PR #3015, opened 2017-03-01
Anonymous (formerly ad hoc) boundary endpoint

RobinIsTheBird:
- Encapsulate anonymous boundary logic in the Boundary model
- Anonymous boundary creation view and url
- Can get geojson from an anonymous boundary that is not part
 of the instance.boundaries.
- Anonymous boundaries do not affect thumbprint.

--

Depends on OpenTreeMap/otm-addons#1452
Connects to #2992

### DIFF
--- a/opentreemap/treemap/views/misc.py
+++ b/opentreemap/treemap/views/misc.py
@@ -13,8 +13,8 @@ from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.contrib.gis.geos import Polygon
 from django.core.exceptions import ValidationError
-from django.http import HttpResponse, HttpResponseRedirect, Http404
-from django.shortcuts import render_to_response
+from django.http import HttpResponse, HttpResponseRedirect
+from django.shortcuts import render_to_response, get_object_or_404
 from django.template import RequestContext
 
 from stormwater.models import PolygonalMapFeature
@@ -109,10 +109,7 @@ def static_page(request, instance, page):
 
 
 def boundary_to_geojson(request, instance, boundary_id):
-    try:
-        boundary = Boundary.all_objects.get(pk=boundary_id)
-    except Boundary.DoesNotExist, Boundary.MultipleObjectsReturned:
-        raise Http404
+    boundary = get_object_or_404(Boundary.all_objects, pk=boundary_id)
     geom = boundary.geom
 
     # Leaflet prefers to work with lat/lng so we do the transformation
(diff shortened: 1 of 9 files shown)


opentreemap/treemap/views/misc.py
```diff
@@ -117,6 +118,18 @@ def boundary_to_geojson(request, instance, boundary_id):
...
 
+def add_anonymous_boundary(request):
+    request_dict = json.loads(request.body)
+    polygon = Polygon(request_dict.get('polygon', []))
+    b = Boundary()
+    b.geom = MultiPolygon(polygon)
+    b.name = ''
+    b.category = ''
+    b.sort_order = 1
+    b.save()
```
jwalgran: The `boundary_autocomple` method in `treemap/views/misc.py` filters based on the value of the `searchable` attribute, which defaults to `True`.

https://github.com/OpenTreeMap/otm-core/blob/caa9153a73cbebb1518a522bbb38fcba8727e252/opentreemap/treemap/views/misc.py#L124

We should set `searchable` to `False` for anonymous boundaries.

I would also like to discuss possibly using a custom manager to keep the logic that defines what "anonymous" means together in the model layer.

[NOTE: this is speculative, untested code]

```python
class BoundaryManager(models.GeoManager):
  def create_anonymous_boundary(self, coordinates):
    polygon = Polygon(request_dict.get('polygon', []))
    return self.create(geom=MultiPolygon(polygon),
                       name='',
                       category='',
                       sort_order = 1,
                       searchable=False)

  def anonymous(self):
    return self.get_queryset().filter(name='', category='')

  def named(self):
    return self.get_queryset().exclude(name='', category='')
```

maurizi: I also think that using a custom manager to encapsulate the logic for named vs. anonymous boundaries is a good idea.

Note that since Django 1.7 you can create a custom manager from a `QuerySet`, which allows the custom methods to be called before or after other ORM methods, instead of requiring that they be first:
```python
class BoundaryQuerySet(models.QuerySet):
    def create_anonymous_boundary(self, coordinates):
        polygon = Polygon(request_dict.get('polygon', []))
        return self.create(geom=MultiPolygon(polygon),
                           name='',
                           category='',
                           sort_order = 1,
                           searchable=False)

    def anonymous(self):
        return self.filter(name='', category='')

    def named(self):
        return self.filter(name='', category='')

...

class Boundary(models.Model):
    ...
    manager = GeoManager.from_queryset(BoundaryQuerySet)()

Issue #2996, opened 2017-02-20
Create a `works_management` Django application with a models file


jwalgran:
#### Models

- `WorkOrder`
  - `Auditable`, so we track state changes. Not `Authorizable`.
  - Has many `Task`s
  - Fields
     - FK to `Instance`
     - created_at (date, non-null, autogenerated)
     - created_by (non-null FK to `User`)
     - updated_at (date, non-null)
     - name (non-null)
- `Team`
  - Fields
    - Non-null FK to `Instance`
    - name (non-null)
- `Task`
  - `Auditable` so we track state changes. Not `Authorizable`.
  - `UDFModel` to allow custom fields
  - Fields
    - Non-null FK to `Instance`
    - Non-null FK to `MapFeature`
    - Nullable FK to `WorkOrder`
    - Nullable FK to `Team`
    - status (non-null choices: `Requested` (default), `Scheduled`, `Completed`, `Canceled`
    - created_at (date, non-null, autogenerated)
    - created_by (non-null FK to `User`)
    - updated_at (date, non-null)
    - requested_date (date without time)
    - scheduled_date (date without time)
    - closed_date (date without time)
    - office_notes (nullable text)
    - field_notes (nullable text)
  - "Canned" custom field
    - Action
      - Can’t delete field
      - Set of fixed choices that cannot be deleted (plant tree, remove tree, etc.) but also allows user-defined choices.

#### Notes

- Use `ugettext_lazy` for `Task.status` choices so they can be translated. Store the selected choice as a number to ease renaming and translation.
- If a `Task` is updated, and the `Task` has a `WorkOrder`, set the value of `updated_at` on both the `WorkOrder` and the `Task`.

RickMohr:
@RobinIsTheBird proposes that `requested_on`, `scheduled_on`, and `closed_on` will be clearer than `requested_date`, `scheduled_date`, and `closed_date`.

PR #3020, opened 2017-03-01
Create works_management app

kdeloach:
Create Django app, models, migrations, and unit tests for
works_management feature.

Connects #2996

### DIFF
--- a/opentreemap/works_management/models.py
+++ b/opentreemap/works_management/models.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import division
+
+from django.contrib.gis.db import models
+from django.utils.translation import ugettext_lazy as _
+from django.utils import timezone
+
+from treemap.models import MapFeature, User
+from treemap.audit import Auditable
+from treemap.udf import UDFModel
+from treemap.instance import Instance
+
+
+class Team(models.Model):
+    instance = models.ForeignKey(Instance)
+    name = models.CharField(max_length=255, null=False, blank=False)
+
+
+class WorkOrder(Auditable, models.Model):
+    instance = models.ForeignKey(Instance)
+    name = models.CharField(max_length=255, null=False, blank=False)
+    created_at = models.DateTimeField(auto_now_add=True)
+    created_by = models.ForeignKey(User)
+    updated_at = models.DateTimeField(auto_now=True)
+
+
+class Task(UDFModel, Auditable):
+    REQUESTED = 0
+    SCHEDULED = 1
+    COMPLETED = 2
+    CANCELED = 3
+
+    STATUS_CHOICES = (
+        (REQUESTED, _('Requested')),
+        (SCHEDULED, _('Scheduled')),
+        (COMPLETED, _('Completed')),
+        (CANCELED, _('Canceled')),
+    )
+
+    instance = models.ForeignKey(Instance)
+    map_feature = models.ForeignKey(MapFeature)
+    work_order = models.ForeignKey(WorkOrder, null=True, related_name='tasks')
+    team = models.ForeignKey(Team, null=True)
+
+    office_notes = models.TextField()
+    field_notes = models.TextField()
+
+    status = models.IntegerField(
+        choices=STATUS_CHOICES,
+        default=REQUESTED)
+
+    requested_on = models.DateField()
+    scheduled_on = models.DateField()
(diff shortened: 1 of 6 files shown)


jwalgran:
There is some lint

```
$ ./scripts/check.sh
+ vagrant ssh app -c 'cd /opt/app/core/opentreemap; flake8 --exclude migrations,opentreemap/settings/local_settings.py *'
works_management/urls.py:1:1: F401 'django.conf.urls.include' imported but unused
works_management/urls.py:1:1: F401 'django.conf.urls.url' imported but unused
Connection to 127.0.0.1 closed.

File: opentreemap/works_management/tests.py [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+
+
+def make_plot(instance, user):
+    p = Point(0, 0)
+    plot = Plot(geom=p, instance=instance)
+    plot.save_with_user(user)
+    return plot
+
+
+class WorksManagementTests(OTMTestCase):
```
jwalgran: I would like to add these two tests.

```python
    def test_cannot_call_save_on_workorder(self):
        # WorkOrder is Auditible and should only allow calling save_with_user.
        w = make_work_order(self.instance, self.user)
        with self.assertRaises(UserTrackingException):
            w.save()

    def test_cannot_call_save_on_task(self):
        # Task is Auditible and should only allow calling save_with_user.
        t = make_task(self.instance, self.user,
                      make_plot(self.instance, self.user))
        with self.assertRaises(UserTrackingException):
            t.save()
```

The WorkOrder test fails. I would expect `save` function in Auditible to be called, but it is not. 

https://github.com/OpenTreeMap/otm-core/blob/746322311ef4df3e550dfdcfd81f9848a5202c53/opentreemap/treemap/audit.py#L615-L618

If we cannot figure out the inheritance issue, the simple solution is to copy the `save` function into the WorkOrder model.

```
    def save(self, *args, **kwargs):
        raise UserTrackingException(
            'All changes to %s objects must be saved via "save_with_user"' %
            (self._model_name))
```

opentreemap/works_management/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Team(models.Model):
+    instance = models.ForeignKey(Instance)
+    name = models.CharField(max_length=255)
+
+
+class WorkOrder(models.Model, Auditable):
+    instance = models.ForeignKey(Instance)
+    name = models.CharField(max_length=255)
```
kdeloach: It should be non-nullable by default, right? But I'll set `null=False` and `blank=False` just to be safe.

opentreemap/works_management/models.py
```diff
@@ -0,0 +1,84 @@
...
+        (COMPLETED, _('Completed')),
+        (CANCELED, _('Canceled')),
+    )
+
+    instance = models.ForeignKey(Instance)
+    map_feature = models.ForeignKey(MapFeature)
+    work_order = models.ForeignKey(WorkOrder, null=True, related_name='tasks')
+    team = models.ForeignKey(Team, null=True)
+
+    name = models.CharField(max_length=255)
```
kdeloach: Thanks, must have been a copy & paste error.

opentreemap/works_management/models.py
```diff
@@ -0,0 +1,84 @@
...
+        },
+    }
+
+    def save_with_user(self, user, *args, **kwargs):
+        """
+        Update WorkOrder fields when Task is saved.
+        """
+        if self.work_order:
+            self.work_order.updated_at = timezone.now()
+            self.work_order.save()
```
kdeloach: I wonder why this doesn't trigger an exception in the unit test? I'll investigate.

opentreemap/works_management/tests.py
kdeloach: Good call. I was just thinking the same thing after your last comment.

kdeloach: The solution is to make sure we inherit from Auditable before Model:

```
-class WorkOrder(models.Model, Auditable):
+class WorkOrder(Auditable, models.Model):
```

kdeloach:
Updates:

* Fix lint
* Fix `Auditable` inheritance issue on `WorkOrder` model
* Unit tests to make sure `save_with_user` is enforced

opentreemap/works_management/models.py
jwalgran: :man_facepalming: I should not comment before sufficient caffeine.

kdeloach:
Thanks.

Issue #3022, opened 2017-03-01
How should we declare protected UDFs?


kdeloach:
Django models that inherit from `UDFModel` may define custom fields using the `udf_settings` property:

```
class Task(UDFModel):
    udf_settings = {
        'Action': {
            'iscollection': False,
            'defaults': {
                'type': 'choice',
                'choices': [
                    'Plant Tree',
                    'Remove Tree',
                    'Other action',
                ],
            }
        },
    }
```

For Works Management tasks, we want the ability to define "protected" choices (for "actions"), which cannot be deleted. Presumably, these are also protected against renaming. How should we update the `udf_settings` syntax to support this?

Here are some ideas:

1. `isprotected` property

```
udf_settings = {
    'Action': {
        'iscollection': False,
        'isprotected': ['Plant Tree', 'Remove Tree'],
        'defaults': {
            'type': 'choice',
            'choices': [
                'Plant Tree',
                'Remove Tree',
                'Other action',
            ],
        }
    },
}
```

2. Enhance `choices` to support metadata tuples

```
udf_settings = {
    'Action': {
        'iscollection': False,
        'defaults': {
            'type': 'choice',
            'choices': [
                ('Plant Tree', 'protected'),
                ('Remove Tree', 'protected'),
                'Other action',
            ],
        }
    },
}
```

3. Enhance `choices` to support a metadata class

```
udf_settings = {
    'Action': {
        'iscollection': False,
        'defaults': {
            'type': 'choice',
            'choices': [
                UDFSettingsChoice('Plant Tree', can_delete=False),
                UDFSettingsChoice('Remove Tree', can_delete=False),
                'Other action',
            ],
        }
    },
}
```

maurizi:
We don't just want to protect against deleting/renaming of choice fields, but also the field itself (the existing `udf_settings` code does not do this).

With that in mind, here is my preference:
```python
udf_settings = {
    'Action': {
        'iscollection': False,
        'is_protected': True,
        'defaults': {
            'type': 'choice',
            'protected_choices': [
                'Plant Tree',
                'Remove Tree',
            ],
            'choices': [
                'Other action',
            ],
        }
    },
}
```

kdeloach:
Thanks, I hadn't considered that fields and choices both need to be protected. I like how this syntax looks compared to the other proposed options. I'll implement this feature based on Mike's example unless anyone else has a better idea.

jwalgran:
[...]

PR #3036, opened 2017-03-06
Implement protected UDF definitions and choices

kdeloach:
## Overview

Adds a new field `is_protected` to the `UserDefinedFieldDefinition` Model which prevents anyone from deleting the UDF. A new property has also been added to the "choice" UDF called `protected_choices` which prevents renaming and deletion of the UDF choice.

The syntax implemented is based on feedback from #3022.

Connects #2996

## Notes

For this iteration, I decided to disable renaming and deleting protected UDFs & choices for **everyone**. In the future, it might make sense to waive this restriction for admins and superusers.

This *should* work out of the box with collection UDFs (all unit tests pass) but I have not created unit test cases specifically to verify this.

## Testing Instructions

* Run migrations
* Execute UDF test suite

```
./scripts/manage.sh test treemap.tests.test_udfs
```

* Create a protected UDF choice

```
> ./scripts/manage.sh shell
from treemap.udf import UserDefinedFieldDefinition
udf = UserDefinedFieldDefinition.objects.filter(id=366)[0] # PhillyTreeMap "Tree Condition" UDF
udf.add_choice('Protected Choice', is_protected=True)
```

* Verify that protected UDF choice is visible and editable on plot detail page
* Verify that UDF search works with protected UDF choice

### DIFF
--- a/opentreemap/treemap/lib/udf.py
+++ b/opentreemap/treemap/lib/udf.py
@@ -70,6 +70,8 @@ def _parse_params(params):
 
     if udf_type in ('choice', 'multichoice'):
         datatype['choices'] = params.get('udf.choices', None)
+        datatype['protected_choices'] = \
+            params.get('udf.protected_choices', None)
 
     datatype = json.dumps(datatype)
 
--- a/opentreemap/treemap/migrations/0044_userdefinedfielddefinition_is_protected.py
+++ b/opentreemap/treemap/migrations/0044_userdefinedfielddefinition_is_protected.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('treemap', '0043_works_management_access'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='userdefinedfielddefinition',
+            name='is_protected',
+            field=models.NullBooleanField(default=False),
+        ),
+    ]
--- a/opentreemap/treemap/templatetags/form_extras.py
+++ b/opentreemap/treemap/templatetags/form_extras.py
@@ -247,14 +247,11 @@ def _is_udf(model, udf_field_name):
             udf_field_name in model.udf_field_names)
 
 
-def _udf_dict(model, field_name):
-    matches = [field.datatype_dict
-               for field in model.get_user_defined_fields()
-               if field.name == field_name.replace('udf:', '')]
-    if matches:
-        return matches[0]
-    else:
-        raise Exception("Datatype for field %s not found" % field_name)
+def _get_udf(model, field_name):
+    for field in model.get_user_defined_fields():
+        if field.name == field_name.replace('udf:', ''):
+            return field
+    raise Exception("UDF '%s' not defined on model" % field_name)
 
 
 # Should a blank choice be added for choice and multichoice fields?
@@ -284,12 +281,13 @@ def field_type_label_choices(model, field_name, label=None,
         if choices and field.null:
             choices = [{'value': '', 'display_value': ''}] + choices
     else:
-        udf_dict = _udf_dict(model, field_name)
+        udf = _get_udf(model, field_name)
+        udf_dict = udf.datatype_dict
         field_type = udf_dict['type']
         label = label if label else udf_field_name
         if 'choices' in udf_dict:
             choices = [{'value': value, 'display_value': value}
-                       for value in udf_dict['choices']]
+                       for value in udf.all_choices]
             if add_blank == ADD_BLANK_ALWAYS or (
                 add_blank == ADD_BLANK_IF_CHOICE_FIELD
                     and field_type == 'choice'
--- a/opentreemap/treemap/tests/test_udfs.py
+++ b/opentreemap/treemap/tests/test_udfs.py
@@ -1306,6 +1306,61 @@ class UdfDeleteTest(OTMTestCase):
         self.assertEquals(1, len(updated_instance.mobile_api_fields))
 
 
+class ProtectedUDFTestCase(OTMTestCase):
+    def setUp(self):
+        User._system_user.save_base()
+
+        self.instance = make_instance()
+        create_stewardship_udfs(self.instance)
+        self.user = make_commander_user(self.instance)
+
+        set_write_permissions(self.instance, self.user,
+                              'Plot', ['udf:Test choice'])
+
+        self.udf = UserDefinedFieldDefinition.objects.create(
+            instance=self.instance,
+            model_type='Plot',
+            datatype=json.dumps({'type': 'choice',
+                                 'protected_choices': ['a'],
+                                 'choices': ['b', 'c']}),
+            iscollection=False,
+            is_protected=True,
+            name='Test choice')
+
+        p = Point(0, 0)
+        self.plot = Plot(geom=p, instance=self.instance)
+        self.plot.save_with_user(self.user)
+
+    def test_cannot_delete_protected_udf(self):
+        with self.assertRaises(AuthorizeException):
+            self.udf.delete()
+
+    def test_can_pick_protected_udf_choice(self):
+        # Assert that ValidationError is not raised
+        self.plot.udfs['Test choice'] = 'a'
+        self.plot.save_with_user(self.user)
+
+    def test_cannot_modify_protected_udf_choice(self):
+        with self.assertRaises(ValidationError):
+            self.udf.add_choice('a')
+        with self.assertRaises(ValidationError):
+            self.udf.delete_choice('a')
+        with self.assertRaises(ValidationError):
+            self.udf.update_choice('a', 'z')
+
+    def test_protected_udf_choice_order(self):
+        self.assertEqual(['a', 'b', 'c'], self.udf.all_choices)
+
+    def test_invalid_udf_definition(self):
+        # Ensure that UDFs cannot be created with *only* protected choices.
+        # The "choices" key must always exist.
+        body = {'udf.name': 'cool udf',
+                'udf.model': 'Plot',
+                'udf.type': 'choice',
+                'udf.protected_choices': ['a']}
+        self.assertRaises(ValidationError, udf_create, body, self.instance)
+
+
 class UdfCRUTestCase(OTMTestCase):
     def setUp(self):
         User._system_user.save_base()
--- a/opentreemap/treemap/udf.py
+++ b/opentreemap/treemap/udf.py
@@ -402,11 +402,9 @@ class UserDefinedFieldDefinition(models.Model):
                     _("Can't modify protected choice '%(choice)s'") % {
                         'choice': old_choice_value}]})
 
-        if new_choice_value:
-            choices = [c if c != old_choice_value else new_choice_value
-                       for c in choices]
-        else:
-            choices = [c for c in choices if c != old_choice_value]
+        choices = self._list_replace_or_remove(choices,
+                                               old_choice_value,
+                                               new_choice_value)
 
         datatype['choices'] = choices
         return datatype
--- a/opentreemap/treemap/udf.py
+++ b/opentreemap/treemap/udf.py
@@ -313,6 +313,11 @@ class UserDefinedFieldDefinition(models.Model):
     required:
 
     'choices' - An array of choice options
+
+    Specify choices that may not be modified with the 'protected_choices'
+    key (optional):
+
+    'protected_choices' - An array of choice options
     """
     datatype = models.TextField()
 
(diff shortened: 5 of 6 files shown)


opentreemap/treemap/tests/test_udfs.py
```diff
@@ -1306,6 +1306,46 @@ def test_delete_cudf_deletes_mobile_api_field_group(self):
...
+
+    def test_cannot_modify_protected_udf_choice(self):
+        def delete_choice():
+            self.udf_def.delete_choice('a')
+
+        def update_choice():
+            self.udf_def.update_choice('a', 'z')
+
+        self.assertRaises(ValidationError, delete_choice)
+        self.assertRaises(ValidationError, update_choice)
```
maurizi: Another test in this style I'd like to see:
```python
self.assertRaises(ValidationError, lambda: self.udf_def.add_choice('a'))

Comment (kdeloach) [inline opentreemap/treemap/tests/test_udfs.py]: Had this originally but it exceeds the maximum line length. Couldn't think of a cleaner way to format this better across multiple lines.

File: opentreemap/treemap/migrations/0044_userdefinedfielddefinition_is_protected.py [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('treemap', '0043_works_management_access'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='userdefinedfielddefinition',
+            name='is_protected',
+            field=models.BooleanField(default=False),
```
kdeloach: Thanks, good catch.

kdeloach: Changed to `NullBooleanField`.

kdeloach:
I assumed that the only way to add protected choices would be to update the UDF definition in Python. If the ability to add protected choices through the admin is a requirement, then I should update this to allow admins & superusers to be able to rename/delete them too.

I'll go through the UDF code again and see if there are any other places where it makes sense to merge choices with protected choices.

opentreemap/treemap/tests/test_udfs.py
maurizi: I meant more another case to test, I don't care about the formatting.

FWIW, I usually use the decorator version of `assertRaises`:
```python
with self.assertRaises(AnException) as e:
    call_the_function(bad_data)

self.assertEquals(e.message, 'You gave bad data')

Comment (kdeloach) [inline opentreemap/treemap/tests/test_udfs.py]: Good idea. I'll add that test and update the formatting.

**Comment by maurizi** [comment]
> I assumed that the only way to add protected choices would be to update the UDF definition in Python. If the ability to add protected choices through the admin is a requirement, then I should update this to allow admins & superusers to be able to rename/delete them too.

As we discussed in-person, we are only letting protected choices be added or updated through Python, but we still want them to be visible in the UI once they've been added.

**Comment by kdeloach** [comment]
Updates:

* `add_choice` now raises a `ValidationError` if the new choice value exists in `protected_choices`
* added new properties to UDF model: `choices`, `protected_choices`, and `all_choices`

These new properties will be helpful in displaying choices and protected choices in a unified way for the frontend. They were less useful than imagined as an abstraction for combining choices, because in the places where we need to check against the protected choices list, only the UDF datatype dict is available instead of the class instance.

File: opentreemap/treemap/udf.py [hunk 3]
```diff
@@ -358,17 +369,24 @@ def _validate_and_update_choice(
...
-            choices = [c for c in choices if c != old_choice_value]
+        if old_choice_value in protected_choices:
+            raise ValidationError({
+                'datatype': [
+                    _("Can't modify protected choice '%(choice)s'") % {
+                        'choice': old_choice_value}]})
+
+        choices = self._list_replace_or_remove(choices,
+                                               old_choice_value,
+                                               new_choice_value)
```
kdeloach: This change isn't related to the protected UDFs work. I just noticed that this logic was already encapsulated in an existing helper function and thought it would be a good idea to use it here too.

kdeloach:
I'm going to take a pass at using these properties in the application views.

kdeloach:
I pushed a change to make protected UDF choices visible and editable on the plot detail page. That change seems to have also impacted the search form as well. Updated this issue with more testing instructions.

opentreemap/treemap/udf.py
```diff
@@ -635,16 +661,19 @@ def _validate_single_datatype(self, datatype):
             if choices is None:
                 raise ValidationError(_('Missing choices key'))
 
-            for choice in choices:
+            protected_choices = datatype.get('protected_choices', [])
+            all_choices = choices + protected_choices
```
maurizi: This should use the `all_choices` property.

opentreemap/treemap/udf.py
```diff
@@ -857,11 +911,15 @@ def clean_value(self, value, datatype_dict=None):
 
         elif 'choices' in datatype_dict:
             def _validate(val):
-                if val not in datatype_dict['choices']:
+                choices = datatype_dict['choices']
+                protected_choices = datatype_dict.get('protected_choices', [])
+                all_choices = choices + protected_choices
```
maurizi: This should use the `all_choices` property.

opentreemap/treemap/udf.py
```diff
@@ -358,17 +369,24 @@ def _validate_and_update_choice(
                 {'datatype': [_("Can't change choices "
                                 "on a non-choice field")]})
 
-        if old_choice_value not in datatype['choices']:
+        choices = datatype['choices']
+        protected_choices = datatype.get('protected_choices', [])
+        all_choices = choices + protected_choices
```
maurizi: This should use the `all_choices` property.

opentreemap/treemap/templatetags/form_extras.py
```diff
@@ -288,8 +288,11 @@ def field_type_label_choices(model, field_name, label=None,
         field_type = udf_dict['type']
         label = label if label else udf_field_name
         if 'choices' in udf_dict:
+            choices = udf_dict['choices']
+            protected_choices = udf_dict.get('protected_choices', [])
+            all_choices = protected_choices + choices
```
maurizi: This should use `UserDefinedFieldDefinition.all_choices` - if we switch from using `_udf_dict` on line 287 to getting the UDF itself, we can do that.

opentreemap/treemap/udf.py
kdeloach: Using properties here breaks collection UDFs.

opentreemap/treemap/udf.py
kdeloach: Using properties here breaks collection UDFs.

opentreemap/treemap/udf.py
kdeloach: Using properties here breaks collection UDFs as well.

opentreemap/treemap/templatetags/form_extras.py
kdeloach: Refactored this to retrieve the UDF field directly instead of the dict and it seems to work.

opentreemap/treemap/udf.py
maurizi: Ah, makes sense. Can we go with your suggestion before about making a helper function that takes a datatype (so we keep logic on which choices go first in one location):

```python
class UserDefinedFieldDefinition(...):
    @property
    def all_choices(self):
        return UserDefinedFieldDefinition._all_choices(self.datatype_dict)

    @classmethod
    def all_choices(cls, datatype):
        if datatype['type'] in {'choice', 'multichoice'}:
            return datatype.get('protected_choices', []) + datatype.get('choices', [])
        return None

Comment (kdeloach) [inline opentreemap/treemap/udf.py]: Good idea. We'll still need a way to obtain `choices`, `protected_choices`, and `all_choices` for both UDFs and Collection UDFs (for `_validate_and_update_choice`). So should we have 6 functions in total to accomplish this?

Comment (maurizi) [inline opentreemap/treemap/udf.py]: > So should we have 6 functions in total to accomplish this?
I think it is OK for `_validate_and_update_choice` to reach into the datatype dictionary when it needs to (it is a private helper method after all), so I don't think we need separate helper functions for `choices` vs. `protected_choices` that support collection UDFs. The goal is more to ensure that logic around which field goes first when you combine the lists is located in one place, so that we don't have things in different orders depending on the code path, and to keep things outside the UDFD class from reaching into the datatype dictionary unnecessarily.

Comment (jwalgran) [inline opentreemap/treemap/udf.py]: If it helps, we do not need to support protected choices for collection UDFs.

Comment (kdeloach) [inline opentreemap/treemap/udf.py]: The problem (not that it's much of a problem at all) is that UDFs and collection UDFs exercise the same code paths for validate, clean, etc. On the plus side, this means that protected choices should just work for collection UDFs, although I have not thoroughly tested it on collections.

File: opentreemap/treemap/udf.py [hunk 8]
```diff
@@ -630,10 +674,7 @@ def _validate_single_datatype(self, datatype):
             raise ValidationError(_('invalid datatype'))
 
         if datatype['type'] in ('choice', 'multichoice'):
-            choices = datatype.get('choices', None)
-
-            if choices is None:
-                raise ValidationError(_('Missing choices key'))
```
kdeloach: I removed this validation error because it's now possible (and valid) for the "choices" key to be missing. For example, if you define "protected_choices" instead, you don't need "choices" (Ex. https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/works_management/models.py#L68). Also, there is another validation constraint which checks that there is at least 1 choice defined, which I think is sufficient. No unit tests failed as a result of this change.

opentreemap/treemap/udf.py
kdeloach: I added this classmethod and property.

kdeloach:
Think this is ready for another look.

kdeloach:
May need help verifying the impact of this change on the API and mobile apps.

kdeloach:
CircleCI is also failing due to Webpack errors.

maurizi:
> May need help verifying the impact of this change on the API and mobile apps.

The code for the goes through `field_type_label_choices` in `form_extras.py`, so I think the mobile apps should be covered.

> CircleCI is also failing due to Webpack errors.

I restarted the build without cache, which should fix that error.

opentreemap/treemap/udf.py
jwalgran: This line in `_validate_and_update_choice` will raise a `KeyError` if the datatype does not have a `choices` key.
https://github.com/OpenTreeMap/otm-core/pull/3036/files#diff-5232da98c01c519f4805c4259ad42fe1R390

It would be safer if we kept choices as a required field and used and empty list in the case where all choices are protected.

kdeloach: Good catch. I'm going to add some test coverage around this and update the WM model so that the UDF definition contains an empty choices field.

opentreemap/treemap/lib/udf.py
```diff
@@ -70,6 +70,8 @@ def _parse_params(params):
 
     if udf_type in ('choice', 'multichoice'):
         datatype['choices'] = params.get('udf.choices', None)
+        datatype['protected_choices'] = \
+            params.get('udf.protected_choices', None)
```
kdeloach: This was updated so that we could test UDF definitions with `protected_choices` in unit tests.

opentreemap/treemap/tests/test_udfs.py
```diff
@@ -1306,6 +1306,61 @@ def test_delete_cudf_deletes_mobile_api_field_group(self):
         self.assertEquals(1, len(updated_instance.mobile_api_fields))
 
 
+class ProtectedUDFTestCase(OTMTestCase):
+    def setUp(self):
+        User._system_user.save_base()
+
+        self.instance = make_instance()
+        create_stewardship_udfs(self.instance)
```
kdeloach: The `udf_create` helper raises a `ValidationException` (something to do with mobile API fields) if the instance doesn't have stewardship UDFs. Even though we don't need stewardship UDFs for this test case, adding them resolved this error. Not sure if this is by design or not.

opentreemap/treemap/udf.py
kdeloach: I have reverted this change and added an extra unit test.

kdeloach:
Pushed one more change to restore a validation constraint that I removed and cleaned up the unit tests a little bit.

maurizi:
Continued :+1:

kdeloach:
What changes do we need to fix the CircleCI build check?

maurizi:
> What changes do we need to fix the CircleCI build check?

Assuming that this pull request is rebased on the latest `develop` so that it has https://github.com/OpenTreeMap/otm-core/commit/41200c083c23426b02b6bc899a9e6e23d8b0e333 rerunning the job without cache (which I just did) should fix it.

If it doesn't, don't worry about it too much - Jenkins is our "source of truth" for pull request happiness, not CircleCI, as CircleCI only tests the open source code and is generally less stable.

kdeloach:
Thanks for the feedback.

Issue #3030, opened 2017-03-05
Reduce size of handles on polygon for drawing GSI feature


dboyer:
Tested on staging in Chrome on a Macbook.

When working on smaller rain gardens or bioswales, the large boxes overlap each other and can be difficult to manipulate.

![screen shot 2017-03-04 at 9 57 53 pm](https://cloud.githubusercontent.com/assets/516764/23584127/a3ae022e-0125-11e7-9d9b-cd7de31409bd.png)

RickMohr:
Those big squares are Leaflet Draw's default icon, but that can be overridden. If somebody (@lederer?) provides a different icon (smaller? not square?) we can use it to replace the big square.

They may have chosen a large-sized default icon to support mobile use. But the situations where we need to create/edit polygons (adding a Bioswale, defining a custom search area, creating a modeling distribution) seem unlikely to be used on mobile.

jwalgran:
If our recently updaded Leaflet supports it then increasing the max zoom beyond the base map resolution may be a simple solution to this problem.

lederer:
Circles like those in the Draw Search Area mocks would be preferable. I'm
OOO today but can whip one up tomorrow, if it can't be done
programmatically in Leaflet.

RickMohr:
Looks like it can all be done with JS and CSS. Working on it...

RickMohr:
@jwalgran, it looks like we can't increase the max zoom until the next release of Leaflet. [`googleMutant`](https://www.npmjs.com/package/leaflet.gridlayer.googlemutant) (the Leaflet plugin we use to display google map tiles) says:

"The `maxNativeZoom` functionality introduced in v0.5.0 requires Leaflet >1.0.3. At the time of this writing, this means "using the master branch of Leaflet".

Currently, increasing the max zoom sort of works -- if you switch to the satellite layer at zoom 21 you will still see it at zoom 22. But if you zoom to 22 and then switch to the satellite layer you won't see it.

RickMohr:
Here's a tiny "Sidewalk Garden" (15 sq ft) at max zoom from `urbanforestmap`, on [production](https://www.opentreemap.org/urbanforestmap/features/2612758) and [staging](https://staging.opentreemap.org/urbanforestmap/features/2612758):

[...]

PR #3039, opened 2017-03-07
Change default vertex icons for drawing/editing polygons

RickMohr:
Leaflet Draw has different polygon vertex icons for touch screens and non-touch screens, and decides which to use based on Leaflet's `L.Browser.Touch`. But with current browsers and Leaflet 1.0.3, `L.Browser.Touch` is `true` for a browser that supports touch even when you're using a non-touch device. That's why we get giant vertex icons on desktop devices. Since for us polygon editing is unlikely to be done via touch, always use the non-touch icons.

Also change the default 8x8 square into a 10x10 circle (with help from CSS on .leaflet-editing-icon)

Connects #3030
Depends on OpenTreeMap/otm-addons#1459

Testing: 
* Create a Bioswale. When editing the area the vertices should be small circles.
* Go to its detail page, Edit, and Move Location. The vertices should be small circles.

### DIFF
--- a/assets/css/sass/partials/pages/_map.scss
+++ b/assets/css/sass/partials/pages/_map.scss
@@ -147,6 +147,10 @@ $light-to-dark-green: #ffffff 10%,
     }
 }
 
+.leaflet-editing-icon {
+    border-radius: 5px;  // Make 10x10 vertex icon into a circle
+}
+
 @media #{$screen-xs} {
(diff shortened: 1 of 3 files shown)


RickMohr:
@azavea-bot rebuild

RickMohr:
@azavea-bot rebuild

PR #3041, opened 2017-03-08
Fix polygon area calculation

jwalgran:
The `getPolygonDisplayArea` function began returning zero when passed the results of calling `getLatLngs` on a boundary polygon. The `getLatLngs` method was producing a doubly-nested array and the `geodesicArea` function only works on a 1-dimensional array. I fixed `getPolygonDisplayArea` by making it recursively descend through an arbitrary number of nested arrays and sum the results of calling `geodesicArea` on any "leaf" array it finds.

---

##### Testing

- Selecting a modeling distribution area created by loading a boundary should show a popup with the correctly calculated area.
- The area should still be correctly calculated when editing a GSI polygon.

I tested the modeling case on PhillyTreeMap by selecting the city boundary which reports 3,819,786,060 sq/ft

screen shot 2017-03-07 at 11 32 21 pm

Wikipedia says that the area of Phildadelphia is 141.6 sq miles, which is 3.94758e+9 square feet.

---

Connects to https://github.com/OpenTreeMap/otm-modeling/issues/146

### DIFF
--- a/opentreemap/treemap/js/src/lib/utility.js
+++ b/opentreemap/treemap/js/src/lib/utility.js
@@ -141,8 +141,20 @@ exports.makeZoomLatLngQuery = function(zoomLatLng) {
 // TODO: Respect instance units configuration.
 // https://github.com/OpenTreeMap/otm-addons/issues/326
 exports.getPolygonDisplayArea = function(poly) {
-    var latLngs = poly.getLatLngs()[0],
-        areaSqMeters = L.GeometryUtil.geodesicArea(latLngs),
+    function totalAreaInMeters(collection) {
+        if (_.isArray(collection[0])) {
+            return _.chain(collection)
+                    .map(totalAreaInMeters)
+                    .reduce(function(sum, num) {
+                        return sum + num;
+                    })
+                    .value();
+        } else {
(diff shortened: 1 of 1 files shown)


opentreemap/treemap/js/src/lib/utility.js
```diff
@@ -141,8 +141,20 @@ exports.makeZoomLatLngQuery = function(zoomLatLng) {
...
-        areaSqMeters = L.GeometryUtil.geodesicArea(latLngs),
+    function totalAreaInMeters(collection) {
+        if (_.isArray(collection[0])) {
+            return _.chain(collection)
+                    .map(totalAreaInMeters)
+                    .reduce(function(sum, num) {
+                        return sum + num;
+                    })
+                    .value();
+        } else {
```
RickMohr: Could write this more concisely as
```js
            return _.reduce(collection, function(sum, geom) {
                return sum + totalAreaInMeters(geom);
            }, 0);
```
... but it's plenty clear as-is.

jwalgran:
The build failure was on test is a different repo. It passed on my development machine. This is most likely a result of the PR builder checking out develop branches instead of release branches.

Issue #3000, opened 2017-02-22
Previous tree selection persists in sidebar after cancelling "Add a Tree"


RickMohr:
1. Select a tree
1. Click "Add a tree" (the selected tree's popup and marker disappear)
1. Click the "X" to cancel adding the tree

Expected: sidebar shows no selected tree
Actual: sidebar shows previously-selected tree

RickMohr:
Proposed solution: 
* Merge `editTreeDetailsMode` into `browseTreesMode` (they already share an inline edit form)
* When deactivating `browseTreesMode`, clear the currently-selected tree (and its sidebar info)

RickMohr:
While we're working on mode transitions I'd like to also address [Robin's comment](https://github.com/OpenTreeMap/otm-core/pull/3012#pullrequestreview-24589439) about unifying where we put code for mode transitions. These changes take significant testing ... let's do it once instead of twice.

PR #3042, opened 2017-03-08
Mode change cleanup

RickMohr:
Functional changes:
* Hide "Details" sidebar when switching out of browse or edit mode
* Don't update URL when adding a tree or resource from the treemap page. The only reason for having `m=addTree` in the URL is so we can initiate adding a tree or resource from another page. There's no value in updating the URL when adding a tree or resource from the treemap page, so don't.

Simplify and clarify `modes.js` and mode modules
* Refactor activation methods to use options instead of named arguments
* Pass `activateBrowseTreesMode` to "add" modes rather than `onClose` function
* Cleaner mode transitions for mobile
* Remove logic to show accordion buttons -- they were never hidden anyway
* Use `dom.` pattern rather than passing selectors in `options`
* Require `plotMarker.js` rather than passing it in `options`
* Remove unused `require`s and `var`s

Testing -- both desktop and mobile

Browse and edit trees
* Select, deselect, and edit trees
  * The "Details" sidebar should behave correctly
  * Saving or cancelling edits should not ask "are you sure"?
* If a tree is selected, it should be deselected (and the "details" sidebar cleared) when you start adding a tree or drawing or editing a custom area.

Add trees and resources
* Add a tree and a resource
* While adding
  * Changing to a different mode (e.g. draw a polygon or add a different kind of resource) should ask "are you sure"?
  * Finishing or cancelling should not ask "are you sure"?
  * The URL should not change

Connects #3000

### DIFF
--- a/opentreemap/treemap/js/src/mapPage/addMapFeature.js
+++ b/opentreemap/treemap/js/src/mapPage/addMapFeature.js
@@ -14,12 +14,17 @@ var $ = require('jquery'),
     geocoderResultsUi = require('treemap/lib/geocoderResultsUi.js'),
     enterOrClickEventStream = require('treemap/lib/baconUtils.js').enterOrClickEventStream,
     otmTypeahead = require('treemap/lib/otmTypeahead.js'),
+    plotMarker = require('treemap/lib/plotMarker.js'),
     config = require('treemap/lib/config.js');
+
+var dom = {
+    exploreMapHeaderLink: '.navbar li.explore-map',
+    addFeatureHeaderLink: '.navbar li[data-feature=add_plot]'
+};
 
 function init(options) {
     var mapManager = options.mapManager,
-        plotMarker = options.plotMarker,
-        onClose = options.onClose || $.noop,
+        activateBrowseTreesMode = options.activateBrowseTreesMode,
         clearChildEditControls = options.clearEditControls || $.noop,
         sidebar = options.sidebar,
         $sidebar = $(sidebar),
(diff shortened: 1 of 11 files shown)


RickMohr:
waffle

maurizi:
Another problem I encountered in mobile responsive mode:

1. Tap on a tree icon
2. Open the tree info by tapping on popup bar at the bottom
3. Enter "Quick Edit" mode
4. Tap on the "Move Tree" button in the upper-left of the map
3. Enter add tree mode by tapping "Add Tree" at the top. When it asks if you want to leave, say yes
4. Exit add tree mode by tapping the "x" on the right of the header
5. The "Move Tree" button is still visible, even though it should only be present when in "Quick Edit" mode

maurizi:
As we discussed, we should probably _clear_ the contents of the accordion when exiting browse trees mode, not just collapse it, since 1) you can always reopen it on desktop and 2) mobile responsive view ignores the collapsed/shown CSS.

opentreemap/treemap/js/src/mapPage/addMapFeature.js
```diff
@@ -353,6 +353,10 @@ function init(options) {
         $(dom.addFeatureHeaderLink).removeClass("active");
         $(dom.exploreMapHeaderLink).addClass("active");
 
+        // Clear classes that control display for mobile
+        $('body').removeClass('add-feature hide-search open');
+        $('#feature-panel').toggleClass('expanded with-map');
```
maurizi: Should this `toggleClass` be `hideClass`?

PR #3047, opened 2017-03-09
Fix UI test

RickMohr:
When you're adding a tree we no longer put `m=addTree` in the URL. So check whether the "Add Tree" sidebar is visible.


maurizi:
:+1:

Issue #3049, opened 2017-03-09
Work order numbers and task numbers start from "1" for each instance


jwalgran:
Compound primary key for WorkOrder and Task (each instance has its own work order sequence starting from 1)

PR #3054, opened 2017-03-13
Create sequential Work Order numbers and Task numbers per instance

kdeloach:
## Overview

Adds a new reference number field on WorkOrder and Task models, which
is sequential (starting from 1), and is distinct per instance.

Migrations have been split up into 3 parts:
1. Add new fields (nullable)
2. Set default values on null fields
3. Add unique key and non-nullable constraint

Connects #3049

## Testing Instructions

* Run migrations
* Run unit tests

```
> ./scripts/manage.sh test works_management.tests
```

### DIFF
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -771,6 +771,38 @@ class Instance(models.Model):
         if errors:
             raise_errors(errors)
 
+    @transaction.atomic
+    def get_next_task_sequence(self, how_many=1):
+        """
+        Return next sequence value and increment counter by `how_many`.
+
+        Arguments:
+        how_many - Amount of sequence values to allocate
+        """
+        if how_many < 1:
+            raise ValueError('how_many must be >= 1')
+        qs = Instance.objects.filter(id=self.id).select_for_update()
+        old_value = qs[0].task_sequence_number
+        new_value = F('task_sequence_number') + how_many
+        qs.update(task_sequence_number=new_value)
+        return old_value
+
+    @transaction.atomic
+    def get_next_work_order_sequence(self, how_many=1):
+        """
+        Return next sequence value and increment counter by `how_many`.
+
+        Arguments:
+        how_many - Amount of sequence values to allocate
+        """
+        if how_many < 1:
+            raise ValueError('how_many must be >= 1')
+        qs = Instance.objects.filter(id=self.id).select_for_update()
+        old_value = qs[0].work_order_sequence_number
+        new_value = F('work_order_sequence_number') + how_many
+        qs.update(work_order_sequence_number=new_value)
+        return old_value
+
     def save(self, *args, **kwargs):
         self.full_clean()
 
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -782,7 +782,7 @@ class Instance(models.Model):
         if how_many < 1:
             raise ValueError('how_many must be >= 1')
         qs = Instance.objects.filter(id=self.id).select_for_update()
-        old_value = qs[0].task_sequence_number
+        old_value = qs.values_list('task_sequence_number', flat=True)[0]
         new_value = F('task_sequence_number') + how_many
         qs.update(task_sequence_number=new_value)
         return old_value
@@ -798,14 +798,22 @@ class Instance(models.Model):
         if how_many < 1:
             raise ValueError('how_many must be >= 1')
         qs = Instance.objects.filter(id=self.id).select_for_update()
-        old_value = qs[0].work_order_sequence_number
+        old_value = qs.values_list('work_order_sequence_number', flat=True)[0]
         new_value = F('work_order_sequence_number') + how_many
         qs.update(work_order_sequence_number=new_value)
         return old_value
 
     def save(self, *args, **kwargs):
-        self.full_clean()
+        self.full_clean(exclude=['task_sequence_number',
+                                 'work_order_sequence_number'])
 
         self.url_name = self.url_name.lower()
+
+        if self.id:
+            # Prevent Django from overwriting these fields with stale data.
+            # These values are managed by get_next_task_sequence and
+            # get_next_work_order_sequence.
+            self.task_sequence_number = F('task_sequence_number')
+            self.work_order_sequence_number = F('work_order_sequence_number')
 
         super(Instance, self).save(*args, **kwargs)
--- a/opentreemap/treemap/instance.py
+++ b/opentreemap/treemap/instance.py
@@ -245,6 +245,9 @@ class Instance(models.Model):
                                         null=True, blank=True)
     eco_rev = models.IntegerField(default=_DEFAULT_REV)
 
+    task_sequence_number = models.IntegerField(default=1, null=True)
+    work_order_sequence_number = models.IntegerField(default=1, null=True)
+
     eco_benefits_conversion = models.ForeignKey(
         'BenefitCurrencyConversion', null=True, blank=True)
 
--- a/opentreemap/works_management/models.py
+++ b/opentreemap/works_management/models.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals
 from __future__ import division
 
 from django.contrib.gis.db import models
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from django.utils import timezone
 
@@ -25,6 +26,24 @@ class WorkOrder(Auditable, models.Model):
     created_by = models.ForeignKey(User)
     updated_at = models.DateTimeField(auto_now=True)
 
+    # This value comes from `instance.work_order_sequence_number`
+    reference_number = models.IntegerField()
+
+    class Meta:
+        unique_together = ('instance', 'reference_number')
+
+    def clean(self):
+        if not self.reference_number:
+            raise ValidationError({
+                'reference_number': [_('Reference number is required.')]})
+
+    def save_with_user(self, user, *args, **kwargs):
+        """
+        Update WorkOrder fields when Task is saved.
+        """
+        self.full_clean()
+        super(WorkOrder, self).save_with_user(user, *args, **kwargs)
+
 
 class Task(UDFModel, Auditable):
     REQUESTED = 0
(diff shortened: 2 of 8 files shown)


opentreemap/works_management/models.py
```diff
@@ -21,10 +22,28 @@ class Team(models.Model):
...
+        unique_together = ('instance', 'reference_num')
+
+    def get_next_reference_num(self):
+        """
+        Return next sequential reference number for instance.
+        """
+        agg_result = Instance.objects.filter(id=self.instance.id) \
+            .annotate(max_value=Max('workorder__reference_num'))[0]
+        max_value = agg_result.max_value or 0
+        return max_value + 1
```
maurizi: In general, it's a good idea to avoid the practice of requesting a value from the database, manipulating it, and saving it to the database.  It leads to race conditions where multiple workers may see the same max value.

A way to avoid the race condition is to get and manipulate the maximum number in the database, which will handle concurrency issues for us. Django has support for this with [Query Expressions](https://docs.djangoproject.com/en/1.8/ref/models/expressions/). Another would be to lock the `Task` / `WorkOrder` table during the transaction using [`select_for_update()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#select-for-update).

kdeloach: I couldn't find any examples of how to use query expressions to create new records with an aggregate value. And I don't think `select_for_update` will work because there's nothing to select until we create the record initially.

This code assumes that objects are created within a transaction. I'm curious to see what happens when multiple workers try to insert tasks at the same time.

Alternatively, we could create 2 new fields on `Instance` to store the "next" values, and use `select_for_update` to safely modify that each time we create a new WorkOrder/Task.

maurizi: > I'm curious to see what happens when multiple workers try to insert tasks at the same time.

This is simple, it is the same thing that happens _outside_ of a transaction: you'll get an `IntegrityError` from the `unique_together` constraint you added :smile: .

Also, `select_for_update` won't catch a race condition when creating the first task for a map, but it can in all other cases. From the Django docs:
> All matched entries will be locked until the end of the transaction block, meaning that other transactions will be prevented from changing or acquiring locks on them.

So if `save_with_user` is wrapped in a `@transaction.atomic` decorator and a call to `Task.objects.filter(instance=self.instance).select_for_update()` is called before the query to get the max value, any concurrent workers should block if necessary while attempting to acquire a lock for the all the pre-existing `Task` rows.

kdeloach: Thanks for the info. I'm not really worried about concurrency issues but now that I think about it we should make sure this works for bulk create operations. I'm going to test this now.

What do you think about creating a general `InstanceSequence` model instead of this? In case we need to add other sequences like this in the future?

jwalgran: We could avoid using an aggregate query by adding `workorder_sequence` and `task_sequence` columns to `treemap_instance` table. Atomically incrementing these would behave in a similar way to using lower level Postgres sequences.

kdeloach: Good idea. I'll change this.

kdeloach:
This has been updated and is ready for another look.

Changes:
* Sequence numbers are now tracked on the `Instance` model (thanks @maurizi  & @jwalgran)
* The `save` method on `Instance` now accepts optional keyword arguments for tasks and work orders (thanks @RobinIsTheBird)  which will:
  * Bulk assign `reference_number` field on models
  * Update the instance sequence field in the DB
  * Refresh instance fields to prevent committing stale data
* Migrations have been split up so they can (theoretically) be applied incrementally to work with blue/green deployments

opentreemap/works_management/models.py
kdeloach: Thanks for the feedback. Updated.

opentreemap/treemap/instance.py
```diff
@@ -768,9 +771,39 @@ def raise_errors(errors):
...
+        Keyword Arguments:
+
+        tasks - List of Task models
+        work_orders - List of WorkOrder models
+        """
         self.full_clean()
 
         self.url_name = self.url_name.lower()
 
+        if tasks:
```
jwalgran: The instance `save` method does not feel like the right place to put this operation. Could you explain the design constraints that would push us toward this approach rather than something like:


- Add an `atomic` helper method to `Instance` to get the next `n` reference numbers and update the `task_sequence_number` field appropriately (`n` would default to 1).
- Decorate `Task.save_with_user` as `atomic` and have it use its `instance` reference to call the helper and fetch the next sequence number.
- Add an atomic `Task.bulk_save_with_user` class method for when you want to create a set of tasks in a loop.

kdeloach: This logic could very easily be moved to a utility function or classmethod. I thought putting it here would prevent a data consistency error, but I just realized that's not true. For example:

```py
# User 1
instance1 = Instance.objects.get(id=1)
# User 2
instance2 = Instance.objects.get(id=1)

# User 1
t = make_task(instance1, ...)
self.assertEqual(1, t.reference_number)
self.assertEqual(2, instance1.task_sequence_number)

# User 2
instance2.save()

# AssertionError: 2 != 1
instance3 = Instance.objects.get(id=1)
self.assertEqual(2, instance3.task_sequence_number)
```

As long as these sequence fields exist on the `Instance` model there is the potential for data consistency errors. For example, using a helper method to allocate sequence numbers inside a transaction:

```py
# User 1
instance = get_instance() # task_sequence_number=1

# User 2
instance = get_instance() # task_sequence_number=1
n = task_sequence_increment(instance, 20) # task_sequence_number=21
tasks[0].reference_number = n
tasks[1].reference_number = n + 1
# etc.

# User 1
instance.save() # task_sequence_number=1; Whoops!
```

Do you think this is a real issue we should be concerned about?

If we could define "unmanaged" fields in Django, that could help avoid this problem. Another alternative could be to use Redis?

jwalgran: Based on the small sample of potential customers we have talked to multiple people creating work orders at the same time would not be a common occurrence. In the interest of standing up the feature sooner rather than later, we could defer integrity error protection.

I was unable to find documentation on Django unmanaged fields. I could only find documentation on unmanaged models.

I think the "unmanaged field" philosophy is correct, though. The sequence numbers will always need to be fetched and updated directly through queryset functions, not via the `Instance` model properties.

I don't think we need to involve another datastore (Redis) to solve the problem, just change the way we are interacting with the Postgres database in this situation.

jwalgran: This `Instance` helper looks like it may be a good starting point

```python
@transaction.atomic
def get_next_task_sequence(self, n=1):
    if n < 1:
        raise ValueError('n must be >= 1')
    qs = Instance.objects.select_for_update()
    next_number = qs.get(id=self.id).work_order_sequence_number
    qs.filter(id=self.id).update(
        work_order_sequence_number=F('work_order_sequence_number') + n)
    if 1 == n:
        return next_number
    else:
        return range(next_number, next_number + n)
```

```
In [1]: i1 = Instance.objects.get(url_name='phillytreemap')

In [2]: i2 = Instance.objects.get(url_name='phillytreemap')

In [3]: i1.get_next_task_sequence()
Out[3]: 1

In [4]: i2.get_next_task_sequence()
Out[4]: 2

In [5]: i1.get_next_task_sequence(n=10)
Out[5]: [3, 4, 5, 6, 7, 8, 9, 10, 11, 12]

In [6]: i2.get_next_task_sequence()
Out[6]: 13
```

kdeloach: I like this approach. I think the risk of someone overwriting instance fields with stale data is low. In that unlikely event, it would trigger `IntegrityErrors` due to the unique constraint. Since there isn't any exception handling around this case, I'm wondering if it makes sense to get rid of the unique constraint?

maurizi: > Since there isn't any exception handling around this case, I'm wondering if it makes sense to get rid of the unique constraint?

I would say to keep the unique constraint - we can always show a "something went wrong, try again" toast to the user in that case.

kdeloach: Cool.

kdeloach:
This is ready for another look.

Changes:

* Remove sequence number logic from save method
* Create helper methods to increment sequence counters
* Unit test `bulk_create` and demonstrate how to assign many reference numbers

Finally, it should be noted that saving an instance after allocating sequence numbers is a mistake, and should be avoided at all costs. It's unclear how to prevent this. But we should be aware of it. For example:

```py
task = make_task()
task.reference_number = instance.get_next_task_sequence()
task.save_with_user(user)
instance.save()  # <--
# Whoops! Sequence numbers overwritten with stale data.
# The next Task we insert could trigger an IntegrityError.

```

There is a feature in the works to add support for "readonly" fields in Django. See https://code.djangoproject.com/ticket/21454 and https://github.com/django/deps/pull/36.

kdeloach:
Pushed a commit which prevents sequence numbers from being overwritten with stale data based on Mike's suggestions. So now...

```py
task = make_task()
task.reference_number = instance.get_next_task_sequence()
instance.save()  # No problem!
```

opentreemap/works_management/models.py
```diff
@@ -74,11 +94,19 @@ class Task(UDFModel, Auditable):
...
     }
 
+    class Meta:
+        unique_together = ('instance', 'reference_number')
+
     def save_with_user(self, user, *args, **kwargs):
         """
         Update WorkOrder fields when Task is saved.
         """
+        if not self.reference_number:
```
RobinIsTheBird: Should this validation be in a `clean` method? (Same for WorkOrder)

opentreemap/works_management/tests.py
```diff
@@ -74,15 +79,82 @@ def test_work_order(self):
...
+
+    def test_task_sequence_bulk_create(self):
+        tasks = [make_task(self.instance, self.user, self.plot, save=False)
+                 for _ in range(10)]
+
+        value = self.instance.get_next_task_sequence(len(tasks))
+        for i in range(len(tasks)):
+            tasks[i].reference_number = value + i
+
+        Task.objects.bulk_create(tasks)
```
RobinIsTheBird: What if `WorkOrder` and `Task` each had a custom `Manager` that overrode `bulk_create` to do the three lines of code above?

kdeloach:
I will make the new fields on `Instance` nullable and then move all the non-null/unique constraint migrations to a separate PR so that we can deploy them when we're ready.

opentreemap/works_management/tests.py
kdeloach: My only hesitation to that idea is that setting these reference numbers requires updating the `Instance` model. I don't know if it's a best practice to have Django models update other models behind the scenes.

opentreemap/works_management/models.py
kdeloach: Good idea.

kdeloach:
Updated with these changes:
* New fields on `Instance` are nullable (created issue #3067 to make them non-nullable)
* Move validation logic from `save` to `clean` method
* Call `full_clean` when saving WM models (and update model attributes to allow blank field values)

kdeloach:
Rebased onto develop.

kdeloach:
Great. Thanks for the feedback.

PR #3055, opened 2017-03-14
anonymous boundary for custom area

RobinIsTheBird:
Here is what it does.

Polygons are completed by user events:
- `mapPage/drawAreaMode.onDrawComplete`
- `mapPage/editAreaMode.saveArea`

When polygons are completed, `locationSearchUI.completePolygon` POSTs
to the backend `anonymous_boundary` url, and adds the polygon to the map.

The stream created by `anonymous_boundary` tells `drawAreaMode` to
set some state and activate browse-trees mode.

After some mapping, it also gets plugged into a bus, which `mapPage`
feeds into `boundarySelect.init`, after some merges and transformations.

`mapPage` also feeds it into a stream that triggers backend searches.

TODO: ajax errors from `anonymous_boundary` are presently ignored.

When `anonymous_boundary` succeeds, it returns `{"id": <boundary id>}`.
This is processed as follows:
- `onValue`,
  * inject the id into the hidden location value dom element
  * tell `boundarySelect` not to zoom on layer change
- send the boundary id on the bus that feeds into `boundarySelect`, which
  * requests geojson from the backend given the id
  * parses the returned geojson
  * displays it on the map
  * sends a notice on another stream, which `mapPage` previously passed
    to `locationSearchUI`, so the latter can remove the draw layer,
    with no "blink"
- on another stream that it feeds, trigger the backend search.

Polygons are cleared by user events:
- clicking the reset button, triggering `searchBar.resetStream`
- clicking the "Clear" control in the Location box

Either way, it calls `locationSearchUI.clearCustomArea`, which
- tells `boundarySelect` it's safe to zoom again
- sets the controls back to standard
- sets the hidden location value to empty
- pushes events out on the `locationChangeBus` where it goes through
  the same pipeline as `anonymous_boundary` success

The backend had to be changed, too.
- the `anonymous_boundary` endpoint has to transform the geometry.
- `search.py` has to use `Boundary.all_objects`

--

Connects to #2994

### DIFF
--- a/opentreemap/treemap/js/src/lib/mapPage.js
+++ b/opentreemap/treemap/js/src/lib/mapPage.js
@@ -53,6 +53,15 @@ module.exports.init = function (options) {
             geocodeEvents);
 
     if (options.shouldUseLocationSearchUI) {
+        // When loading the page from a URL with a BOUNDARY_ID search,
+        // we must wait for the location typeahead to populate the
+        // "Search by Location" input box (by fetching the boundary name
+        // from the server).
+        //
+        // That triggers an event on searchBar.programaticallyUpdatedStream,
+        // and we call locationSearchUI.onSearchChanged,
+        // which uses the presence of text in the "Search by Location" input box
+        // to distinguish between named and anonymous boundaries.
         builtSearchEvents
             .merge(searchBar.programmaticallyUpdatedStream)
             .onValue(locationSearchUI.onSearchChanged);
--- a/opentreemap/treemap/js/src/lib/mapPage.js
+++ b/opentreemap/treemap/js/src/lib/mapPage.js
@@ -13,7 +13,8 @@ var $ = require('jquery'),
     urlState = require('treemap/lib/urlState.js'),
     SearchBar = require('treemap/lib/searchBar.js'),
     config = require('treemap/lib/config.js'),
-    boundarySelect = require('treemap/lib/boundarySelect.js');
+    boundarySelect = require('treemap/lib/boundarySelect.js'),
+    locationSearchUI = require('treemap/mapPage/locationSearchUI.js');
 
 $.ajaxSetup(require('treemap/lib/csrf.js').jqueryAjaxSetupOptions);
 
@@ -21,10 +22,16 @@ module.exports.init = function (options) {
     // init mapManager before searchBar so that .setCenterWM is set
     var zoomLatLngOutputStream = mapManager.createTreeMap(options);
 
+    var locationChangedEvents = locationSearchUI.init({
+            map: mapManager.map
+        });
+
     // When there is a single geocode result (either by an exact match
     // or the user selects a candidate) move the map to it and zoom
     // if the map is not already zoomed in.
-    var searchBar = SearchBar.init();
+    var searchBar = SearchBar.init({
+        anotherNonGeocodeObjectStream: locationChangedEvents
+    });
 
     searchBar.geocodedLocationStream.onValue(_.partial(onLocationFound, mapManager));
 
@@ -35,6 +42,9 @@ module.exports.init = function (options) {
                 .map('.search'),       // get search string
             searchBar.resetStream.map({})
         );
+    searchBar.resetStream.onValue(function () {
+        locationSearchUI.clearCustomArea();
+    });
 
     var geocodeEvents = searchBar.searchFiltersProp
         .sampledBy(searchBar.geocodedLocationStream);
@@ -43,6 +53,8 @@ module.exports.init = function (options) {
             triggeredQueryStream,
             searchBar.filterNonGeocodeObjectStream,
             geocodeEvents);
+
+    builtSearchEvents.onValue(locationSearchUI.onSearchChanged);
 
     triggeredQueryStream.onValue(searchBar.applySearchToDom);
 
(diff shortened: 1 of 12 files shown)


opentreemap/treemap/js/src/mapPage/locationSearchUI.js
```diff
@@ -41,11 +51,19 @@ function clearLocationInput() {
...
+                      !!searchEvent.filter['mapFeature.geom'] &&
+                      searchEvent.filter['mapFeature.geom'].IN_BOUNDARY) ||
+                      null;
     if (text) {
-        showControls(dom.controls.searched);
         $(dom.locationSearched).html(text);
+        showControls(dom.controls.searched);
+    } else if (boundaryId !== null) {
+        $(dom.locationValue).val(String(boundaryId));
+        showControls(dom.controls.customArea);
```
RickMohr: Currently there's a bug -- if you search for a custom boundary and then reload the page, we think you have a custom area. But the logic here should suffice, since there should be `text` in the custom boundary case.

I suspect it's a sequencing issue -- that this code is called before `Search.applyFilterObjectToDom`.

opentreemap/treemap/js/src/mapPage/locationSearchUI.js
```diff
@@ -1,9 +1,15 @@
...
-    _ = require('lodash');
+    _ = require('lodash'),
+    Bacon = require('baconjs'),
+    reverse = require('reverse'),
+    config              = require('treemap/lib/config.js'),
+    BU = require('treemap/lib/baconUtils.js'),
+    boundarySelect = require('treemap/lib/boundarySelect');
 
 var dom = {
+    locationValue: '#boundary',
```
RobinIsTheBird: This element is shared between custom area and search by location. It's consistent with the other `location*` dom property names.

opentreemap/treemap/js/src/mapPage/locationSearchUI.js
```diff
@@ -60,12 +78,46 @@ function setPolygon(newPolygon) {
...
+    });
+
+    // Map id from anonymousBoundaryStream into the data structure
+    // that locationChangeBus expects.
+    var builtSearchStream = Bacon.combineTemplate({
+        display: null,
+        filter: {
+            'mapFeature.geom': {IN_BOUNDARY: anonymousBoundaryStream}
+        }
+    });
```
RobinIsTheBird: This suggestion results in the issue we saw before where `editAreaMode` shows the edit tooltip but not the markers. I don't know why yet. Investigating.

There's another bug here. Without your suggested changes, when editing a polygon in `editDrawMode`, if you click "Search" instead of "Done", although the location search box returns to the "_CUSTOM AREA_" display, clicking the pencil icon no longer gets into edit area mode at all.

Investigating that, too.

opentreemap/treemap/js/src/lib/mapPage.js
```diff
@@ -13,18 +13,25 @@ var $ = require('jquery'),
...
+    var locationChangedEvents = locationSearchUI.init({
+            map: mapManager.map
+        });
+
     // When there is a single geocode result (either by an exact match
     // or the user selects a candidate) move the map to it and zoom
     // if the map is not already zoomed in.
-    var searchBar = SearchBar.init();
+    var searchBar = SearchBar.init({
+        anotherNonGeocodeObjectStream: locationChangedEvents
```
RobinIsTheBird: The suggestion as offered doesn't work, but prompted a fix in `searchBar.init`, which you'll see in the next iteration.

opentreemap/treemap/js/src/mapPage/locationSearchUI.js
RobinIsTheBird: Sequencing issue, but also required additional logic here.

opentreemap/treemap/js/src/mapPage/locationSearchUI.js
```diff
@@ -60,12 +80,43 @@ function setPolygon(newPolygon) {
...
+            boundarySelect.shouldZoomOnLayerChange(false);
+        });
+
+    // `customAreaSearchBus` prompts `searchBar` to call `Search.buildSearch`,
+    // which rebuilds the search stringfrom scratch.
+    //
+    // That must happen after an anonymous boundaryhas been created,
+    // and its id set in`dom.locationValue`,
+    // which supplies the value to `Search.buildSearch`.
+    customAreaSearchBus.plug(identifiedBoundaryStream);
```
RickMohr: Missing spaces -- "stringfrom", "boundaryhas", "in`dom"

opentreemap/treemap/js/src/lib/mapPage.js
```diff
@@ -44,6 +52,15 @@ module.exports.init = function (options) {
             searchBar.filterNonGeocodeObjectStream,
             geocodeEvents);
 
+    if (options.shouldUseLocationSearchUI) {
```
RickMohr: Understanding what's happening with `programmaticallyUpdatedStream` and why is tricky. How about a comment here like:

When loading the page from a URL with a `BOUNDARY_ID` search we must wait for the location typeahead to populate the "Search by Location" input box (by fetching the boundary name from the server). That triggers an event on `searchBar.programaticallyUpdatedStream` and we call `locationSearchUI.onSearchChanged`, which uses the presence of text in the "Search by Location" input box to distinguish between named and anonymous boundaries.

Issue #3024, opened 2017-03-02
Friendlier location search


RickMohr:
Change what happens if you hit ENTER or click "Search" without having chosen an autocomplete option:
* If there is an autocomplete "hint", proceed as if user had chosen that option
* If there is no autocomplete option (e.g. when you type "xyzzy"), show a modal saying "There were no results matching your location search"

RickMohr:
Front end changes:
* Simplify `searchBar`, `geocoderResultsUI`, and possibly others

Back end changes:
* Remove `geocode.views._geocode_without_magic_key`
* Don't return choices for which `isCollection == true` (e.g. Beaches)

PR #3058, opened 2017-03-15
Friendlier and simpler location search

RickMohr:
Simplify back end `geocode` endpoint
* Require a magic key (i.e. an autocomplete choice)
* Return a single result, not multiple candidates
* Return location as lat/long

`geocoderUi.js`:
* Rename from `geocoderInvokeUi.js`
* Handle typeahead(s)
  * Trigger a search when typeahead item chosen (so you don't have to click "Search")
  * Autocomplete when search initiated (so you can e.g. type a few characters and hit ENTER)
* Handle creating the `enterOrClickEventStream`
  * Previously it was created separately by each of 3 client modules
  * Handle ENTER key from the input of each specified typeahead
* Client-side handling of "location not found" errors (i.e. unable to autocomplete)

Simplify clients of location search
* Refactor/simplify `searchBar.js`
* Simplify create instance, add tree/resource
* Remove `geocoderResultsUi.js`
* geocoded locations now in lat/long

Also:
* Show errors using toast instead of alerts
* In `otmTypeahead.js`, omit choices representing more than one match (e.g. "Beaches")
* Remove special formatting of "Search by Location" area when a location search is in effect. Simplifies the UI and the code, and allows modifying a successful search.
* Rename `searchBar.filterNonGeocodeObjectStream` to `searchBar.filtersStream`
* Fix bug in `typeahead.jquery.js`
* Update geocoder test to use ESRI suggestion engine before geocoding

Testing:
* Exercise all location typeaheads:
  * Treemap page
  * Plot detail page
  * Add tree/resource
  * Modeling page
  * Create instance
* Choosing a typeahead item (boundary or address, or species) triggers a search
* Hitting ENTER:
  * autocompletes before searching (boundary or address, or species)
  * shows error toast if there's no match (e.g. type "laskdjf")
* You can search for an address, edit the address, and hit ENTER to search again
* Hitting the "X" in "Search by Location" clears the current location search (but preserves the rest of the search)
* Hitting "Reset" clears all search items, including the current location search
* Typing "Beaches" shows no autocompletions

Connects #3024


assets/css/sass/partials/_layout.scss
```diff
@@ -321,9 +321,6 @@ body {
     background-color: $secondary-color;
     color: white;
     display: none;
-    overflow: hidden;
```
RickMohr: They were necessary to show an ellipsis when we were displaying the current location search specially, but aren't needed for other uses of `.location-search-alternate`.
image

opentreemap/geocode/tests.py
```diff
@@ -22,40 +27,57 @@ def setUp(self):
...
+            'https://geocode.arcgis.com/arcgis/rest/services/'
+            'World/GeocodeServer/suggest?' +
+            urlencode({
+                'f': 'json',
+                'searchExtent': json.dumps(dict({
+                    'spatialReference': {'wkid': 102100}
+                }, **extent)),
+                'text': '340 n 12th st philadelphia'
+            }))
+        result = json.loads(requests.get(url).content)
```
RickMohr: Seems like a tradeoff. With your suggestion the test would run a bit faster, but is vulnerable to changes in the ESRI service. As-is, if ESRI changes their magic key generation formula the test won't break.

jwalgran: I agree. The test is already making a geocode call, so our test suite would still break if the ESRI services were offline.

opentreemap/treemap/js/src/lib/geocoderUi.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
```
RickMohr: @jwalgran, I fixed (1) and (3) but could not reproduce (2). Ready for another look.

jwalgran: I confirmed that the bad address searches are fixed.

I am still able reproduce the "select boundary with down arrow" problem. I am seeing slightly different behavior now. It correctly selects and zooms to the 3rd region, but then immediately selects and zooms to the first region. The back button history has both regions.  I captured a GIF of it.

otm-region-arrow-key-select-behavior


Possibly related: if I use the mouse to select a region from the autocomplete list, on staging, I get the solid color overlay

screen shot 2017-03-21 at 1 42 02 pm

On this branch I do not

screen shot 2017-03-21 at 1 42 22 pm

jwalgran: I just tried this in Firefox and I did not see the error, but I did see that the search by location box no longer has the dark background.

jwalgran: I reread the PR description and I see that this removing the dark background is intentional

RickMohr: Entered as #3073

PR #3068, opened 2017-03-20
Fix IE11 issues

RickMohr:
* Update browser detection regex
* Include promise polyfill
* Declare IE11 compatibility in HTML header so that `indexOf` is defined (http://stackoverflow.com/a/41830202)
* Require IE 11 even in debug mode
* Update tests

Connects OpenTreeMap/otm-clients#353

### DIFF
--- a/opentreemap/opentreemap/middleware.py
+++ b/opentreemap/opentreemap/middleware.py
@@ -5,7 +5,8 @@ from django.conf import settings
 import logging
 logger = logging.getLogger(__name__)
 
-_ie_version_regex = re.compile(r'MSIE\s+([\d]+)')
+# http://stackoverflow.com/a/30907476
+_ie_version_regex = re.compile(r'(MSIE\s+|Trident.*rv[ :])([\d]+)')
 
 CONTENT_TYPE_PASS_THROUGHS = ('application/json', 'text/csv')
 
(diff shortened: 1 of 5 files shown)


opentreemap/treemap/templates/base.html
```diff
@@ -12,6 +12,7 @@
     <meta charset="utf-8">
     <title>{% block title %}OpenTreeMap{% block instance_title %}{% endblock %}{% block page_title %}{% endblock %}{% endblock title %}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
jwalgran: The MSDN [Specifying legacy document modes](https://msdn.microsoft.com/en-us/library/jj676915(v=vs.85).aspx) page makes this statement about `<meta http-equiv="X-UA-Compatible" content="IE=edge">`

> Note that this is functionally equivalent to using the HTML5 doctype. 

Seeing that we have a `<!DOCTYPE html>` at the top of our template, I removed the `X-UA-Compatible` tag and did not see any console errors or map problems when browsing the map page in IE 11 (via https://www.browserstack.com)

RickMohr: My IE11 had an obscure "compatibility mode" checked. `meta` tag removed.

Issue #3048, opened 2017-03-09
Add a priority field to the Works Management Task model


jwalgran:
- `priority` should be an integer choice field with a fixed set of choices. 
- We want the choices to be in order so that we can use a `min` database function to easily aggregate tasks and find the most urgent priority.
- The choice labels should be High, Medium, Low

jmorrison1847:
I vote for replacing "Normal" with "Routine" since I see that word a lot in inventories that I look at. But the general thought of 3 buckets makes total sense to me.

jmorrison1847:
Sorry--missed the "Comment" button and accidentally closed it.

RickMohr:
I think we should consider "High", "Medium", and "Low" because they're easy to understand and give customers more flexibility in how to categorize things. For example I might want to prioritize a task as "High" priority but I might not consider it "Urgent".

dboyer:
I also agree that we should stick with as general as possible and would prefer High, Medium, Low or 1, 2, 3 rather than something like Urgent or Immediate.

jwalgran:
There is a consensus around High, Medium, and Low, so I updated the issue description.

PR #3069, opened 2017-03-20
Add priority field to WM Task model

kdeloach:
## Overview

Connects #3048 

## Notes

Since WM has not been deployed to production I decided not to make the priority field nullable.

## Testing Instructions

 * Run migrations

### DIFF
--- a/opentreemap/works_management/migrations/0006_task_priority.py
+++ b/opentreemap/works_management/migrations/0006_task_priority.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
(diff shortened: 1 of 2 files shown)


kdeloach:
Not sure I understand why Jenkins build is failing.

maurizi:
> Not sure I understand why Jenkins build is failing.

The output from the Jenkins console has "Connection to 127.0.0.1 closed by remote host." which I _think_ generally means it was shutdown by the host system due to resource pressure.  Rerunning the build often succeeds when this happens.

kdeloach:
Cool. It failed more than once but it looks like the problem fixed itself.

kdeloach:
Migrations need to run on staging for this.

PR #3072, opened 2017-03-22
Make Task scheduled_on and closed_on nullable.

jwalgran:
These dates are not required when a task is initially created.


jwalgran:
This change will be part of a larger PR.

Issue #3070, opened 2017-03-21
Small improvements to polygon editing UI


RickMohr:
@lederer says:

Basically all minor style and text changes. Feel free to bounce the PR to me for final tweaks. All icons should be available in fontello (font awesome) except the draw/polygon icon, which I can provide as SVG, draggable into fontello.

https://marvelapp.com/3j5eid8/screen/25874908

RickMohr:
This was completed.

PR #3074, opened 2017-03-22
Polygon ui tweaks

RickMohr:
* Use updated Fontello files with custom polygon icon
* Tweak UI as specified (https://marvelapp.com/3j5eid8/screen/25874912 and discussion with Scott)
* Remove "hamburger" icons in search wells
* Use small polygon vertex icons (as we do elsewhere in the app)

Connects #3070


RickMohr:
waffle

Issue #3051, opened 2017-03-09
Add endpoint to create Tasks from a map feature search


jwalgran:
Accepts search JSON to define the map features for which `Tasks` will be created along with field values collected from the modal form (#3050)

The endpoint will associate the created `Task`s with either a new `WorkOrder` or an existing work order.

`search.py` can already convert query JSON to a result set.

Coordinate the development of this with #3050 to make sure the data collected by the form matches what the endpoint expects. It is not strictly blocked by #3050 because the endpoint can be unit tested.

RobinIsTheBird:
I assume that only plots (with or without trees) are considered, and stormwater resources are ignored. Yes?

jwalgran:
We decided to exclude non-plot map features from the work order process.

PR #3075, opened 2017-03-23
create tasks from search

RobinIsTheBird:
Migrations:
- Make `required_on` and `closed_on` nullable
- Make `Task.work_order` required
- Give the note fields a default value
- Merge
- Resolve that django forms make choice fields required
  even if the model field has a default, no matter what blank is

Other apps:
- Change the 404 decorator to receive additional global errors
  along with field errors
- Provide for model names in the form of <app_name>.<ModelName>
  in `treemap.util`, `treemap.udf`, and `treemap.instance`
- Safe generic utility to get models by app name
- A new setting to interface `treemap` to other apps in a
  more loosely coupled way
- Make UDFs safer by not instantiating the model
- Add `clz` kwarg to `treemap.instance.get_or_create_udf`,
  for cases where the caller already has the model class

Works management:
- Move ref number reservation into `save_with_user`
- Use a `ModelForm` to perform field value validation in the view
- Bulk create with user to produce an audit trail
- Correctly packaged errors on bad regular fields,
  bad udf values, and bad searches

--

Connects to #3051

### DIFF
--- a/opentreemap/opentreemap/settings/default_settings.py
+++ b/opentreemap/opentreemap/settings/default_settings.py
@@ -9,6 +9,17 @@ FEATURE_BACKEND_FUNCTION = None
 USER_ACTIVATION_FUNCTION = None
 INSTANCE_PERMISSIONS_FUNCTION = 'treemap.instance.get_instance_permission_spec'
 
+ENABLEABLE_FEATURES = {
+    'gsi': {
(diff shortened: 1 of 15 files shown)


RobinIsTheBird:
Will amend the commit subject line before merge, to remove "WIP".

Issue #3063, opened 2017-03-17
Polygonal search tool should not be visible on mobile


maurizi:
It isn't visible on the map view, but is on the advanced search view.

![localhost-6060-phillytreemap-map--z 18-40 02121--75 20617 nexus 5x](https://cloud.githubusercontent.com/assets/4432106/24050675/fb351426-0b05-11e7-96fd-d03a1c4ab4e2.png)

RickMohr:
I don't see this problem on staging or on my machine. @maurizi could you please double check on your browser?

maurizi:
I can reproduce this on both Firefox and Chrome in responsive design mode (and also on Chrome for Android)
 1. Go to staging.opentreemap.org/phillytreemap/map
 2. Start responsive design mode
 2. Open the advanced search menu and observe that there is no icon
1. Refresh the page
 2. Open the advanced search menu and observe that there is a search icon

It is only hidden if the page started out big and then was shrunk... very odd

RobinIsTheBird:
In this case, does "mobile" mean `$screen-xs    : "(max-width: 767px)";` or `$screen-ltmd  : "(max-width: 991px)";`?

maurizi:
`$screen-xs` - thats the cut-off where we show the advanced search as a fullscreen view instead of in dropdown menus

PR #3077, opened 2017-03-27
Polygonal search tool not available on xs mobile

RobinIsTheBird:
Easier to add an `!important` than to figure out where `style="display: inline"` is being assigned to the dom element and make it stop doing that.

--

Connects to #3063

### DIFF
--- a/assets/css/sass/partials/_header.scss
+++ b/assets/css/sass/partials/_header.scss
@@ -201,7 +201,11 @@
             width: calc(100% - #{$mobile-search-button-group-width});
         }
         .draw-area {
-            display: none;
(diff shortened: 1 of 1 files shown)


assets/css/sass/partials/_header.scss
```diff
@@ -201,7 +201,7 @@
             width: calc(100% - #{$mobile-search-button-group-width});
         }
         .draw-area {
-            display: none;
+            display: none !important;
```
maurizi: Can we add a comment here similar to your PR description:
```
// JS add/removes 'display: inline' for desktop users, using '!important' to
// override that for mobile users without having to make the JS conditional on screen size.

PR #3079, opened 2017-03-27
Revert works management

jwalgran:
This was created by browsing the commit history to identify the works management related pull requests, then reverting those pull requests, from newest to oldest, by running

```
git revert -m 1 {{sha1 of the merge commit}}
```

---

##### Testing

There have not been any non-works mangagement migrations mixed with the works management migrations, so it is possible to check out `develop` and unapply the works managemnt migrations by migrating `treemap 0042`. Dependencies should take care of reversing the works management migrations. There should not be a need to migrate `works_management zero`

There should be no remaining references to works management in the application.

The polygonal search feature and other map page functionality should still work correctly.

---

Required by https://github.com/OpenTreeMap/otm-addons/pull/1475
Connects to https://github.com/OpenTreeMap/otm-cloud/issues/359


opentreemap/treemap/decorators.py
```diff
@@ -168,12 +168,9 @@ def run_and_catch_validations(*args, **kwargs):
             return req(*args, **kwargs)
         except ValidationError as e:
             if hasattr(e, 'message_dict'):
+                message_dict['fieldErrors'] = e.message_dict
                 message_dict['globalErrors'] = [_(
                     'One or more of the specified values are invalid.')]
-                if 'globalErrors' in e.message_dict:
-                    message_dict['globalErrors'] += \
-                        e.message_dict.pop('globalErrors')
-                message_dict['fieldErrors'] = e.message_dict
```
maurizi: This may be worth keeping: "Change the 404 decorator to receive additional global errors
  along with field errors" from https://github.com/jwalgran/otm-core/commit/ab457ea78c45f41b226d5e4bc85032aea271cdba

jwalgran:
@azavea-bot rebuild

opentreemap/treemap/decorators.py
jwalgran: That does look like it may be a keeper. @RobinIsTheBird can you confirm this? Would you want to include anything else in a commit that restores these lines?

RobinIsTheBird: I'll check. I've been verifying operation. Found a bug but determined that it was also in staging, so unrelated to this PR.

RobinIsTheBird: Nothing else is worth preserving. Only useful for starting a new module.

Issue #3083, opened 2017-03-28
Cannot search from detail page


dboyer:
Tested on staging in Chrome on a Macbook.

1. Add a tree
2. Go to tree's detail page
3. While on detail page, enter an address in the location search
4. Click Search
5. Nothing happens

The same error occurs when trying to search by species. The Advanced search bar also does not drop-down when searching from a detail page.

dboyer:
Looks good on staging now.

PR #3085, opened 2017-03-29
Fix search from detail page

RobinIsTheBird:
Dereferencing an undefined `options` variable resulted in a
JS stacktrace.

--

Connects to #3083


jwalgran:
This needs to be moved over and merged into the `release/2.16.0` branch.

maurizi:
> This needs to be moved over and merged into the release/2.16.0 branch.

I changed the base so this will merge into the release branch.

Issue #3087, opened 2017-03-30
Address search with accented character fails


RickMohr:
1) https://www.opentreemap.org/bup
2) In "Search by Location" paste "Calle Hernán Cortés, Muñoz, San Luis Potosí, MEX" and click the matching autocomplete.

Expected: Map zooms to the address
Actual: "There were no results matching your search"

PR #3088, opened 2017-03-30
Support address search with accented characters

RickMohr:
OMGEO calls `urllib.urlencode`, which calls `str()` on the address and errors out.

Solution is to encode the address using UTF-8.

Connects #3087


RobinIsTheBird:
Looks good to me, fwiw.

PR #3091, opened 2017-04-04
Custom Fields management page rework

RobinIsTheBird:
Remove a now unused udf property.
Add option to inlineEditForm.js to allow override of the
function to get data to send to the server.
Styling.

--

Connects to OpenTreeMap/otm-addons#1121
Depends on OpenTreeMap/otm-addons#1478

### DIFF
--- a/opentreemap/treemap/js/src/lib/inlineEditForm.js
+++ b/opentreemap/treemap/js/src/lib/inlineEditForm.js
@@ -90,7 +90,7 @@ exports.init = function(options) {
             $("table[data-udf-id] .placeholder").css('display', 'none');
         },
 
-        getDataToSave = function() {
+        getDataToSave = options.getDataToSave || function() {
             var data = FH.formToDictionary($(form), $(editFields), $(displayFields));
 
             // Extract data for all rows of the collection,
--- a/assets/css/sass/partials/pages/_admin.scss
+++ b/assets/css/sass/partials/pages/_admin.scss
@@ -926,6 +926,9 @@
             }
         }
     }
+    .choice-item {
+        margin-bottom: 4px;
+    }
 }
 
 #save-udf-panel ul {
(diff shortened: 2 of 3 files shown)


RobinIsTheBird:
@azavea-bot rebuild

opentreemap/treemap/js/src/lib/inlineEditForm.js
```diff
@@ -90,7 +90,7 @@ exports.init = function(options) {
             $("table[data-udf-id] .placeholder").css('display', 'none');
         },
 
-        getDataToSave = function() {
+        getDataToSave = options.getDataToSave || function() {
```
RickMohr: Can you use `options.onSaveBefore` rather than a new mechanism?
* Here's where it's [implemented](https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/js/src/lib/inlineEditForm.js#L187)
* Here's [an example](https://github.com/OpenTreeMap/otm-addons/blob/develop/cloud_management/js/src/plan.js#L45) of using it

RobinIsTheBird: @RickMohr, I _am_ using `options.onSaveBefore`, for the confirmer.

It doesn't appear to be intended for gathering data, since `getDataToSave` is called unconditionally, and will not work with the data for this page.

But I'll go look again.

RobinIsTheBird:
Jenkins failed because it wasn't using the `otm-addons` branch.

RobinIsTheBird:
@azavea-bot rebuild

opentreemap/treemap/js/src/lib/inlineEditForm.js
RickMohr: Sorry about that premature comment. I see.

Issue #3092, opened 2017-04-04
Stuck in edit mode in user-drawn polygonal search


dboyer:
Tested in staging in Chrome on a Macbook

1. Draw polygon
2. Search results and ecobenefits change as soon as polygon is closed
3. Click Edit pencil icon
4. Drag a corner of the polygon
5. Click search
6. Polygon returns to original shape rather than keeping the edited shape and check mark icon in location box becomes a pencil again
7. Moving mouse over polygon shows the help text to drag or move a corner
8. There are no handles to move a corner and no editing is possible
9. Clicking the pencil icon does nothing
10. I can click the trash can icon to delete the polygon or click reset

RickMohr:
@dboyer, some behavior choices if you hit "Search" while editing a custom area, in order of increasing difficulty:
1. Abort your edit session and search with the unedited custom area.
1. Save your edit session and search with the updated custom area.
1. Ask the user what they want to do.

I think most users would expect (2) -- probably they didn't realize they were supposed to hit the check mark to save the edited area. But (2) is a bit tricky. (1) is easy. (3) is rather involved.

jwalgran:
Is disabling search while in polygon edit mode a viable/desirable option?

RickMohr:
If we go with (1) that's a reasonable option. We'd also have to disable the species autocomplete since that can trigger a search.

PR #3094, opened 2017-04-06
Disable search while editing custom area

RickMohr:
* Change "Search" button from `a` to `button` to allow easy disabling
* Disable/enable searching on activate/deactivate of "edit area" mode
* Tweak alignment of top controls

Connects #3092


RickMohr:
Search during draw does no harm. We could consider disabling it for consistency's sake if somebody thinks it's important.

Issue #3029, opened 2017-03-04
Stewardship search with date does not include date when searching


dboyer:
Tested on staging (testmap0301) in Chrome on a Macbook

1. Add stewardship entries for "watered" (default tree stewardship category) and "pruneddd" (modified tree stewardship category) with several different dates on several different trees
2. Search for trees watered between Feb 26 and March 2
3. Results say 2 trees, 0 empty planting sites
4. No tree dots are highlighted on map
5. Export search results and see that the trees have entries for watered on Jan 11 and Feb 17

I found similar results when searching pruneddd. I tested the above steps on production and did not see the same results.

dboyer:
The issue happens on both tree and planting site stewardship searches.

RobinIsTheBird:
duplicated in Chrome on Ubuntu, same staging treemap.

RobinIsTheBird:
This is not a regression. I easily reproduced it in production.

dboyer:
If it's useful, on production map testmap0301 I saw date based stewardship working successfully as described:

1. Added two trees with a stewardship action of Pruneddd - one on January 30, 2017 and one on February 18, 2017. 

2. Searched tree care with the following terms: I am looking for trees that have been Pruneddd between 1/1/2017 and 1/31/2017.

3. Search results returned one result as expected

4. Changed the date parameters to 1/1/2017 and 3/1/2017

5. Search returned two results as expected

RickMohr:
@dboyer, to make it fail you need to have a Pruneddd entry that is not in the date range and a non-Pruneddd Stewardship entry that is in the date range.

Should return no results, but returns one.

jwalgran:
We decided that this should not block the 2.15.0 release so I removed it from the milestone.

RobinIsTheBird:
@dboyer and @RickMohr, to make it fail, all you need is a Pruneddd entry that is in the date range, and a Watered entry that is not in the date range, on the _same tree_.

PR #3101, opened 2017-04-11
Stewardship with date search fix

RobinIsTheBird:
Changes
------------
Searches that include multiple queries on a single collection UserDefinedFieldDefinition need to be grouped together into a single subquery.

The previous search created a separate subquery for each one, and so incorrectly matched entities that had one UserDefinedCollectionValue that matched the action, another that matched the date, but none that matched both.

This fix necessitated test changes.

Now also includes a JS fix that threw, and prevented the frontend from submitting a search.

Testing Notes
------------------
Note that the fix does not cause the search cache to be flushed, so you will see the same wrong results if you execute the exact same query.

Anonymous boundary search is an easy way to game the cache. Just change the boundary a little each time.

I tried disabling the search cache in settings, but it resulted in an exception that prevented search from completing. I did not investigate the exception because it appeared to be out of scope.

--

Connects to #3029

### DIFF
--- a/opentreemap/treemap/search.py
+++ b/opentreemap/treemap/search.py
@@ -6,6 +6,7 @@ from __future__ import division
 from json import loads
 from datetime import datetime
 from functools import partial
+from itertools import groupby, chain
 
 from django.db.models import Q
 
@@ -122,7 +123,6 @@ class FilterContext(Q):
 
         super(FilterContext, self).__init__(*args, **kwargs)
 
-    # TODO: Nothing uses add, is it necessary?
     def add(self, thing, conn):
         if thing.basekeys:
             self.basekeys = self.basekeys | thing.basekeys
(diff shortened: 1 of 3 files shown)


opentreemap/treemap/search.py
```diff
@@ -170,18 +170,82 @@ def create_filter(instance, filterstr, mapping):
...
+def _parse_query_dict(query_dict, mapping):
+    # Separate the collection udf queries from the scalar queries,
+    # handle them distinctly, and then combine back together.
+
+    # A search for {
+    #    'udf:plot:<udfd>.action': ...,
+    #    'udf.plot:<udfd>.date': ...}
+    # should match IFF a plot with a UserDefinedCollectionValue
+    # matches both the action and the date parts.
+    # By correlary, it should NOT match a plot that has
```
maurizi: `correlary` :arrow_right: `corollary`

PR #3102, opened 2017-04-12
Management pages in "General" category

RickMohr:
* Restore test of `/management` URL now that `site_config` is present

Also add "Units" page.

Only the final two commits are relevant.

Required by OpenTreeMap/otm-addons#1488

### DIFF
--- a/opentreemap/manage_treemap/views/management.py
+++ b/opentreemap/manage_treemap/views/management.py
@@ -5,12 +5,20 @@ from __future__ import division
 
 import locale
 
+from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.urlresolvers import reverse
+from django.core.validators import URLValidator, validate_email
 from django.shortcuts import redirect
 from django.utils.translation import ugettext as _
 
+from manage_treemap.views import update_instance_fields
 from opentreemap.util import json_from_request, dotted_split
 from otm_comments.views import get_comments
+from treemap.images import save_image_from_request
+from treemap.lib import COLOR_RE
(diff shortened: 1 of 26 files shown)


opentreemap/manage_treemap/views/management.py
```diff
@@ -13,15 +14,19 @@
 from django.utils.translation import ugettext as _
```
jwalgran: The closed source version of this module imports `ugettext_lazy` for translating the `green_infrastructure_labels` content. I don't see any references to `green_infrastructure_labels` outside of this module, and it is only used once within this module, so it may be easier to just remove `green_infrastructure_labels` as a variable rather than import 2 versions of the translation function.

RickMohr: Great subtle find. Here `green_infrastructure_labels` is used as a value for `label_dict`. The other value of `label_dict` uses `ugettext` (`_`) for translation, so I think it's safe to leave as-is. `ugettext` is recommended for views rather than `ugettext_lazy` since different view loads may require different languages.

RickMohr:
Rebased on `develop`

opentreemap/manage_treemap/views/management.py
jwalgran: It is my understanding that because `green_infrastructure_labels` is defined at the module level rather than inside a view function using `ugettext` will cause the text to be translated just once when the module is loaded and we would need to use `ugettext_lazy` or move the declaration inside of a view function to ensure that the text is translated on request.

RickMohr: Now I see! Fixed.

Issue #3043, opened 2017-03-08
Map does not resize in mobile responsive view when entering "Move Tree" mode from Quick Edit


maurizi:
This causes the Google logo to be in the middle of the map.

PR #3103, opened 2017-04-12
Fix map height during move tree on mobile

maurizi:
We needed to call `map.invalidateSize()` after changing the map height

Connects to #3043


jwalgran:
Merging for Mike, who is out of the office.

PR #3108, opened 2017-04-18
Defer choice delete until Save

RobinIsTheBird:
Styling for additional error message.

--

Connects to OpenTreeMap/otm-addons#1482
Is required by OpenTreeMap/otm-addons#1494


RickMohr:
+1

PR #3109, opened 2017-04-18
Defer choice delete until Save

RobinIsTheBird:
Changes
-------
- Added docstrings in the endpoint explaining what data is expected
- Removed markup for the delete choice confirmer from the template
- Removed markup that cued the JavaScript to pop up the delete choice
 confirmer
- Added markup for a trash div to hold pending deletes
- Removed a JavaScript TODO for the Save Confirmer in favor of an issue,
 OpenTreeMap/otm-addons#1493
- Replaced JavaScript code dealing with the confirmer with code
 to stage deletions to a trash div
- Related JavaScript changes for the content of the Save Confirmer,
 and fixing the dom up after Save or Cancel

--

Connects to OpenTreeMap/otm-addons#1482

### DIFF
--- a/opentreemap/manage_treemap/js/src/udfs.js
+++ b/opentreemap/manage_treemap/js/src/udfs.js
@@ -22,16 +22,17 @@ var dom = {
 
     // Trash can on the choice item in Edit mode
     choiceDelete: '[data-action="delete"]',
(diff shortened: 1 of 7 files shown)


RickMohr:
Works!
+1 once tests pass.

Issue #1386, opened 2014-05-22
The paging UI on the photo review page does not react when moderation actions change the number of pages


jwalgran:
If there are two pages and you reject enough photos so that there is only one page full, the paging UI still shows 2 pages.

maurizi:
This is no longer an issue

PR #3117, opened 2017-04-21
Default static pages

RobinIsTheBird:
Changes
-------
- Data migration to add default Resources, About, and FAQ pages
  to instances that don't have them
- Styling for the `<dl>` in the About page

Testing Notes
-------------
Before running the migration, find both instances that do and
that do not have static pages with these names defined.

After running the migration, make sure that
- instances that did not have these pages now do
- the three pages have the right content
- the three pages are formatted nicely
- instances that already had custom versions of these pages
  still have their custom pages

--

Connects to #1386

### DIFF
--- a/assets/css/sass/partials/pages/_content.scss
+++ b/assets/css/sass/partials/pages/_content.scss
@@ -23,7 +23,7 @@
       &.title {
         list-style: none;
         margin-top: 10px;
-        
+
         a {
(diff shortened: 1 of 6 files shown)


RobinIsTheBird:
Static page editor issue is OpenTreeMap/otm-addons#1506

PR #3119, opened 2017-04-26
Remove code related to modeling prioritization

RickMohr:
Connects OpenTreeMap/otm-addons#1503
Required by OpenTreeMap/otm-addons#1508


jwalgran:
`webpack.common.config` has a reference to boostrap-slider that we will also want to remove

https://github.com/OpenTreeMap/otm-core/blob/c8f63f5fd09f43d26a639179b16af50089fd6be7/webpack.common.config.js#L52

jwalgran:
`urlState.js` has a reference to prioritization that it looks like we should remove.

https://github.com/OpenTreeMap/otm-core/blob/c8f63f5fd09f43d26a639179b16af50089fd6be7/opentreemap/treemap/js/src/lib/urlState.js#L20

RickMohr:
Fixed those two oversights before merging.

PR #3123, opened 2017-05-10
Add app to model projected tree growth and mortality

RickMohr:
Connects OpenTreeMap/otm-addons#1504
Depends on OpenTreeMap/otm-addons#1516
Depends on OpenTreeMap/otm-cloud#376


opentreemap/opentreemap/settings/#local_settings.py#
```diff
@@ -0,0 +1,257 @@
+import os
```
RickMohr: Oops. Gone.

opentreemap/modeling/run_model/land_use.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
```
RickMohr: Good point. Gone!

RickMohr:
Right. Squashed.

Issue #3115, opened 2017-04-20
Update core JS libraries, test thoroughly, enter any issues


RickMohr:
* bacon
* lodash
* ramda

We decided not to update JQuery at this time

PR #3128, opened 2017-05-16
Update lodash and ramda to latest versions.

maurizi:
I attempted updating Bacon.js as well, but had to revert those changes, as it caused significant regressions with no simple fixes. See https://github.com/OpenTreeMap/otm-core/issues/3129 for more details.

Connects to #3115

### DIFF
--- a/opentreemap/treemap/js/src/lib/utility.js
+++ b/opentreemap/treemap/js/src/lib/utility.js
@@ -143,7 +143,7 @@ exports.makeZoomLatLngQuery = function(zoomLatLng) {
 exports.getPolygonDisplayArea = function(poly) {
     function totalAreaInMeters(collection) {
         if (_.isArray(collection[0])) {
-            return _.chain(collection)
+            return _(collection)
                     .map(totalAreaInMeters)
                     .reduce(function(sum, num) {
                         return sum + num;
--- a/opentreemap/treemap/js/src/lib/utility.js
+++ b/opentreemap/treemap/js/src/lib/utility.js
@@ -147,8 +147,7 @@ exports.getPolygonDisplayArea = function(poly) {
                     .map(totalAreaInMeters)
                     .reduce(function(sum, num) {
                         return sum + num;
-                    })
-                    .value();
+                    });
         } else {
             return L.GeometryUtil.geodesicArea(collection);
         }
--- a/opentreemap/importer/js/src/lib/status.js
+++ b/opentreemap/importer/js/src/lib/status.js
@@ -39,7 +39,7 @@ function initTypeaheads() {
             $hidden = $c.find(dom.resolver.species.hidden);
 
         // make sure this is a species popover
-        if (R.every(R.not(_.isEmpty), [$input, $hidden])) {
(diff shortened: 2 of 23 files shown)


RobinIsTheBird:
Smoke testing, first.

With:
```
$ scripts/git.sh sb
===== otm-cloud =====
## develop...origin/develop
===== otm-core =====
## update-js-libs...maurizi/update-js-libs
===== otm-addons =====
## update-js-libs...maurizi/update-js-libs
===== otm-tiler =====
## develop...origin/develop
===== otm-ecoservice =====
## develop...origin/develop
```
I navigated to `http://localhost:6060/rgs-161123-01/map/`, which came up with a gray empty map panel.

The console shows
```
baconUtils.js:32 Uncaught TypeError: R.complement is not a function
    at eval (baconUtils.js:32)
    at Object.<anonymous> (base.js:687)
    at __webpack_require__ (base.js:586)
    at fn (base.js:110)
    at eval (export.js:4)
    at Object.<anonymous> (base.js:2089)
    at __webpack_require__ (base.js:586)
    at fn (base.js:110)
    at eval (base.js:9)
    at Object.<anonymous> (base.js:1993)
udfcSearch.js:65 Uncaught TypeError: _(...).map(...).map(...).flatten(...).uniq(...).map(...).fromPairs is not a function
    at eval (udfcSearch.js:65)
    at Object.<anonymous> (base.js:1533)
    at __webpack_require__ (base.js:586)
    at fn (base.js:110)
    at eval (searchBar.js:18)
    at Object.<anonymous> (base.js:1527)
    at __webpack_require__ (base.js:586)
    at fn (base.js:110)
    at eval (mapPage.js:13)
    at Object.<anonymous> (base.js:1521)
```
I updated `develop`, checked `update-js-libs` back out, and got the same thing.

RobinIsTheBird:
So far I've been unable to break anything in `otm-core`. Found an independent problem in `otm-addons`.

Tomorrow, I'll test `otm-website` apart from the management plan page, and modeling and prioritization.

So :+1:, but probably don't want to check it in until the other two pass inspection.

RobinIsTheBird:
I have to rescind my thumbs up. Found a problem that didn't come up until I tested modeling.

opentreemap/treemap/js/src/lib/utility.js
```diff
@@ -143,7 +143,7 @@ exports.makeZoomLatLngQuery = function(zoomLatLng) {
 exports.getPolygonDisplayArea = function(poly) {
     function totalAreaInMeters(collection) {
         if (_.isArray(collection[0])) {
-            return _.chain(collection)
+            return _(collection)
                     .map(totalAreaInMeters)
                     .reduce(function(sum, num) {
                         return sum + num;
```
RobinIsTheBird: `reduce` here returns a scalar floating point number, which has no `value` method. So the call to `value` on line 151 gets a `TypeError`.

Reproduce by navigating to Plan, adding a scenario, adding one or more trees, and clicking to draw the distribution area. Click to make some points. Try to close the polygon. `TypeError`.

RobinIsTheBird: The new push fixes it.

PR #3134, opened 2017-05-18
Use cached plot count in public instances GeoJSON view

maurizi:
Fetching every map's tree/plot count was too slow for use in the new marketing site (taking 2-10 seconds).

Since we already cache plot count, I switched to getting the `plot_count` from the cache, and entirely removed the `tree_count`.

Required by https://github.com/OpenTreeMap/otm-website/pull/178

Connects to OpenTreeMap/otm-website#176


maurizi:
@RobinIsTheBird I fixed the unit tests.

PR #3141, opened 2017-05-31
Fix skipped test

RobinIsTheBird:
Have to override `FEATURE_BACKEND_FUNCTION=None`
in order to `add_map_feature_types`.

Fixed this in order to be able to write a test for the real issue
that also needs to use `add_map_feature_types`.

--

Connects to OpenTreeMap/otm-addons#1394


RobinIsTheBird:
Putting this fix in a separate PR because it doesn't directly relate to the issue it was blocking.

RobinIsTheBird:
@azavea-bot rebuild

PR #3144, opened 2017-06-05
ecobenefits totals

RobinIsTheBird:
`treemap.lib.format_benefits` now calculates `'benefits_total_currency'`
correctly, and also adds its formatted results to
`rslt['benefits']['all']['totals']`.

>`format_benefits` now omits non-annual currency from total currency based on a `dict` in `BenefitCategory`. I hope that the code and variable names make this self-explanatory, but if they do not, it's fair to ask for a comment in the code.

`treemap/partials/eco_benefits.html` adjusted accordingly.

Total currency calculation no longer needed in `treemap.views.tree`.

Added a very basic resource ecobenefits test,
and a test for total currency for both plots and resources,
which failed before correcting `format_benefits`.

--

Connects to OpenTreeMap/otm-addons#1394

### DIFF
--- a/opentreemap/treemap/ecobenefits.py
+++ b/opentreemap/treemap/ecobenefits.py
@@ -30,6 +30,14 @@ class BenefitCategory(object):
 
     GROUPS = (ENERGY, STORMWATER, AIRQUALITY, CO2, CO2STORAGE)
 
+    is_annual_table = {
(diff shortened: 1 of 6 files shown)


RobinIsTheBird:
Fixed TreeMapLA `map_feature`.

Issue #3143, opened 2017-06-01
Add versioning to branded CSS


maurizi:
Files going through Webpack are getting a "version", but our CSS is being built by Django at run-time for maps with custom colors

We should add a cache-buster value to the URL for this file (based on date? etag? commit?) so that we can add a far-futures header to this URL in `nginx`.

PR #3148, opened 2017-06-21
Add checksum to URL for both static and dynamic SASS bundles

maurizi:
We had mistakenly omitted using the `[chunkhash]` in the Webpack SASS bundle.
Adding it in allows us to use the bundle name as a cache-buster URL parameter for the dynamically computed SASS bundle.

Required by https://github.com/OpenTreeMap/otm-cloud/pull/382

Connects to #3143

### DIFF
--- a/opentreemap/treemap/templates/instance_base.html
+++ b/opentreemap/treemap/templates/instance_base.html
@@ -7,10 +7,13 @@
 
 {% block application_css %}
 
+{% get_files 'js/treemap/base' 'css' as css_files %}
 {% if request.instance.scss_variables %}
-  <link id="application-css" href="{% url 'scss' %}?{{ request.instance.scss_query_string }}" rel="stylesheet">
+<!-- The Webpack bundle name is added as a query parameter for cache busting since
+     it includes a checksum of the bundle, letting us set far-future expires cache headers -->
+  <link id="application-css" rel="stylesheet"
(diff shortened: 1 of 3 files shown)


maurizi:
@azavea-bot rebuild

opentreemap/treemap/templates/instance_base.html
```diff
@@ -7,10 +7,13 @@
 
 {% block application_css %}
 
+{% get_files 'js/treemap/base' 'css' as css_files %}
 {% if request.instance.scss_variables %}
-  <link id="application-css" href="{% url 'scss' %}?{{ request.instance.scss_query_string }}" rel="stylesheet">
+<!-- The Webpack bundle name is added as a query parameter for cache busting since
+     it includes a checksum of the bundle, letting us set far-future expires cache headers -->
+  <link id="application-css" rel="stylesheet"
+        href="{% url 'scss' %}?{{ request.instance.scss_query_string }}&url={{ css_files.0.name|urlencode }}">
```
maurizi: I think we're OK on both counts.

The `{% if ... %}` block above checks for `request.instance.scss_variables`, so `scss_query_string` won't be blank.

`css_files.0.` is what we've been using to get the Webpack CSS bundle URL ever since we switched to Webpack, and from inspecting the `webpack-stats.json` file there is only ever one entry per bundle.

Issue #3084, opened 2017-03-28
Replying to comment posts on last comment


dboyer:
This is on production

1. Add a comment
2. Add a second comment
3. Click reply on the first comment
4. Enter a comment and click Post
5. Comment appears on a thread for the second comment rather than the first

RobinIsTheBird:
The problem is on the backend

dboyer:
I'm fine with blocking this and updating django-threadedcomments as part of our general python dependency updates to see if that solves the issue.

jwalgran:
The bug persists after updating to the latest versions of our comment libraries https://github.com/OpenTreeMap/otm-core/commit/bbecd2a2d39160e93816670c2d587e0a8e016e5e

PR #3151, opened 2017-06-28
Add template tag to sequence comments in thread order

jwalgran:
In my analysis of the template tags in both `django-contrib-comments` and `django-threadedcomments` (the older versions we use and also the latest code in Github) I did not see any code that put comments in a parent-then-children sequence.

---

##### Testing

The test procedure in #3084 should produce correct results.

---

Connects to #3084


jwalgran:
@azavea-bot rebuild

PR #3155, opened 2017-06-29
Fix geocode test to be compatible with python-omgeo 2.0.0

jwalgran:
The updated ESRI World Geocoding Service returns a slightly different lat/lng for the old Azavea office address.

screen shot 2017-06-29 at 11 23 35 am

---

Connects to https://github.com/OpenTreeMap/otm-cloud/issues/384


jwalgran:
Quick merge to get successful hotfix branch build.

PR #3160, opened 2017-07-10
Skip tree creation if tree UDF values are empty

jwalgran:
We previously added code to protect against the iOS app always sending values for all UDF fields. While this protective code does skip over the values, it only runs after the presence of a `tree:` field in the request triggers the instantiation of a `Tree`, which then gets saved.

In this commit I have added a check to skip field value processing entirely and bypass the creation of a new `Tree` in the situation where the `Tree` does not already exist and the field value is empty.

---

##### Testing

The new unit test will fail if run with the `develop` branch and passes on this branch.

On an instance that has a choice custom field on Tree, adding an empty planting site via the iOS app works correctly.

---

Connects to https://github.com/OpenTreeMap/otm-clients/issues/382

### DIFF
--- a/opentreemap/treemap/tests/test_views.py
+++ b/opentreemap/treemap/tests/test_views.py
@@ -597,13 +597,21 @@ class PlotUpdateTest(OTMTestCase):
         set_write_permissions(self.instance, self.user,
                               'Plot', ['udf:Test choice', 'udf:Test col'])
         set_write_permissions(self.instance, self.user,
-                              'Tree', ['udf:Test col'])
+                              'Tree', ['udf:Test choice', 'udf:Test col'])
 
         self.choice_field = UserDefinedFieldDefinition.objects.create(
             instance=self.instance,
             model_type='Plot',
             datatype=json.dumps({'type': 'choice',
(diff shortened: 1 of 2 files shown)


opentreemap/treemap/tests/test_views.py
```diff
@@ -668,6 +668,20 @@ def test_creates_new_plot_and_tree(self):
         created_plot_update.current_tree().delete_with_user(self.user)
         created_plot_update.delete_with_user(self.user)
 
+    def test_does_not_create_tree_if_tree_field_value_is_an_empty_string(self):
+        plot = Plot(instance=self.instance)
+
+        update = {'plot.geom': {'x': 4, 'y': 9},
+                  'plot.readonly': False,
+                  'tree.udf:Test choice': ''}
```
jwalgran: Addressed in 79b44df

jwalgran:
Thanks for the much improved diff. Rebased on top of the latest `develop` and added a fixup commit.

jwalgran:
Waiting to merge until we fix the health check endpoint

Issue #3162, opened 2017-07-11
Ability to cancel map invitations


jwalgran:
_From @hadleystein1 on January 13, 2017 21:42_

The ability to delete map invitations in cases where the email address is accidentally typed incorrectly.

![screen shot 2017-01-13 at 4 41 14 pm](https://cloud.githubusercontent.com/assets/15219245/21946616/2e405c32-d9af-11e6-965e-8873de88ad3f.png)

Make sure that accepting an invitation that no longer exists fails gracefully.

_Copied from original issue: OpenTreeMap/otm-addons#1408_

PR #3168, opened 2017-07-17
Add ability to remove user invitations to the Users management page

maurizi:
Connects to #3162


opentreemap/manage_treemap/templates/manage_treemap/partials/user_roles.html
```diff
@@ -19,6 +19,7 @@ <h3 class="management-header">{% trans "Invited tree map users" %}</h3>
                 <th data-sort>
                     <a href="{{ invite_sort_url }}&invite_sort={{ 'admin'|reverse_if_current:invite_sort }}">{% trans "Management Access" %}</a>
                 </th>
+                <th data-class="edit">Action</th>
```
jwalgran: We should wrap this with `trans`.

I can make this column disappear with the following steps:

- Go to the users management page
- Click "Edit"
- Click "Cancel"

maurizi: Fixed in 448bbf60e951580ba72b0fe40ba1341e13652de3

Issue #3163, opened 2017-07-11
Registration and authentication workflow enhancements


jwalgran:
### Registration

- [x] Edit the registration form to require entering the email address twice and add a validation to verify that they two entries match.

- [x] Add “you will not be able to make any edits to the map until your account is confirmed” to the “Account Created” page shown after the account registration form is successfully submitted.

### Authentication

- [x] If logging in fails due to an inactive account, show a prominent link to trigger sending a confirmation email. Send the email when the link is clicked and redirect to a confirmation page.

### Password Reset

- [x] If a user attempts to reset the password for an inactive account, show a message with a link to to trigger sending a confirmation email. Send the email when the link is clicked and redirect to a confirmation page. (Same endpoint and confirmation page uses when requesting a new confirmation email from the login page)

jwalgran:
Labeled "high" because this new issue supersedes other issues labeled "high."

PR #3174, opened 2017-07-26
Registration and authentication workflow enhancements

maurizi:
Connects to #3163

### DIFF
--- a/opentreemap/treemap/templates/treemap/resend_activation_email.html
+++ b/opentreemap/treemap/templates/treemap/resend_activation_email.html
@@ -1,0 +1,40 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<div class="content login-signup">
+  <div class="well login-signup-panel">
+    <legend>{% trans "Resend Activation Email" %}</legend>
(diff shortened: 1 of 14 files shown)


opentreemap/treemap/templates/treemap/resend_activation_email.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <form method="post" action="{% url 'resend_activation_email' %}">
+      {% csrf_token %}
+      <label>
+          {% trans "Username" %}
+      </label>
+      <div class="row">
+          <div class="col-md-9">
+              {% if username %}
+                  <input type="text" name="username" required value="{{ username }}" />
```
jwalgran: A username like `jwalgran+2017-07-26` is loaded into this form as `jwalgran 2017-07-26`. This is likely due `+` being interpreted as a space when used in a URL.

maurizi: Fixed in https://github.com/OpenTreeMap/otm-core/pull/3174/commits/332f0729ed8a05742af21a2f03b39c21a6f72590

PR #3177, opened 2017-08-02
Use ESRI credentials when geocoding stored points/addresses

maurizi:
This is required to be compliant with the ESRI terms of service.

Some of our uses of geocoding are "use and discard" and we don't need to pass the  `forStorage` flag, so I made it an option to `geocodeStream`. All uses of `reverseGeocodeStream` can be stored in a database, so I made it always pass `forStorage`

Required by https://github.com/OpenTreeMap/otm-addons/pull/1540
Depends on https://github.com/OpenTreeMap/otm-cloud/pull/401

Connects to OpenTreeMap/otm-cloud#309

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ pep8==1.4.6 # rq.filter: ==1.4.6
 Pillow==3.3.1
 psycopg2==2.6.2                          # http://initd.org/psycopg/docs/news.html
 python-dateutil==2.5.3                   # https://github.com/dateutil/dateutil/blob/master/NEWS
-git+https://github.com/maurizi/python-omgeo.git@authenticate-esri-wgs
+python-omgeo==3.0.0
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -30,7 +30,7 @@ pep8==1.4.6 # rq.filter: ==1.4.6
 Pillow==3.3.1
 psycopg2==2.6.2                          # http://initd.org/psycopg/docs/news.html
 python-dateutil==2.5.3                   # https://github.com/dateutil/dateutil/blob/master/NEWS
-python-omgeo==2.0.0
+git+https://github.com/maurizi/python-omgeo.git@authenticate-esri-wgs
```
maurizi: TODO: Switch this to a new release once https://github.com/azavea/python-omgeo/pull/40 is merged

maurizi:
@azavea-bot rebuild

Issue #3145, opened 2017-06-07
Species search finds tree of wrong species


RickMohr:
On [this page](https://www.opentreemap.org/phillytreemap/map/?z=20/40.02581/-75.21212&q=%7B%22species.id%22%3A%7B%22IS%22%3A%2272560%22%7D%7D) note that the search is for Allegheny Serviceberry but the dark green dot is a Japanese tree lilac.

(A friend posted to be on the lookout for edible Juneberries / Serviceberries so I tried to see where they were on `phillytreemap`.)

maurizi:
Ok, I've got a handle on what's going on here.

There are _11_ trees at this exact same location, each with a different species (presumably testing data?).
![localhost-6060-phillytreemap-map--z 20-40 02581--75 21212 q 7b 22species id 22 3a 7b 22is 22 3a 2272560 22 7d 7d](https://user-images.githubusercontent.com/4432106/27293479-85f65520-54e4-11e7-87dd-59f13d2d2bc6.png)

![screenshot from 2017-06-19 11-45-08](https://user-images.githubusercontent.com/4432106/27293586-e1cfd11e-54e4-11e7-9a96-78f03d6f852e.png)

Our tile requests for PNGs include the search query, but our UTF Grid requests do not, so that users can still select grayed-out dots and see a popup.  The tiler _must_ pick a single feature to place in each location in the grid, and seems to have arbitrarily picked the Japanese Lilac.

This problem is compounded by the "hide-at-zoom" tiler optimization, as it will group trees in _different_ locations into the same spot on the map at higher zoom levels, which is also reflected in the UTF grid, regardless of the search.

I can see several possibilities for fixing this:
 - Have the existing UTF grid layer be filtered by the current search. This will prevent clicking on grayed-out dots, and will also harm search performance (twice the number of queries).
 - Add a new UTF grid layer above the unfiltered layer when searches are present, and forward misses from the filtered layer to the unfiltered layer.  This is more complex, and has the same performance problems as the above option, but does allow clicking the grayed-out trees.
 - Remove the UTF Grid entirely in favor of AJAX requests to the Django backend. This is simpler, and more performant when performing the search, but is slower to respond to a map click, and prevents us from showing a hand cursor when hovering over a tree dot.

dboyer:
[...]

PR #3181, opened 2017-08-08
Disambiguate in map popup for multiple features at the same location

maurizi:
Uses the same distance calculation used in the "Nearby Trees" sidebar
on the plot detail page (but also includes non-Plot map features)

Connects to #3145

### DIFF
--- a/opentreemap/treemap/js/src/mapPage/browseTreesMode.js
+++ b/opentreemap/treemap/js/src/mapPage/browseTreesMode.js
@@ -9,14 +9,17 @@ var $ = require('jquery'),
     buttonEnabler = require('treemap/lib/buttonEnabler.js'),
     config = require('treemap/lib/config.js'),
     format = require('util').format,
-    plotMarker = require('treemap/lib/plotMarker.js');
+    plotMarker = require('treemap/lib/plotMarker.js'),
+    webMercatorToLeafletLatLng = require('treemap/lib/utility').webMercatorToLeafletLatLng;
 
 var dom = {
         sidebarContent: '#map-info',
         mapFeatureAccordion: '#map-feature-accordion',
         treeDetailAccordion: '#tree-detail',
         fullDetailsButton: '#full-details-button',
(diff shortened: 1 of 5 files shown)


jwalgran:
Looks good. I will give it an interactive test when I am done with release testing.

opentreemap/treemap/js/src/mapPage/browseTreesMode.js
```diff
@@ -49,11 +52,19 @@ function init(options) {
...
         clickedIdStream = utfEventStream.map('.data.' + config.utfGrid.mapfeatureIdKey),
-        accordionHtmlStream = BU.fetchFromIdStream(clickedIdStream,
+        pagingIdStream = $('body').asEventStream('click', dom.popupPagingButtons)
+            .doAction(changePopupSelection)
+            .doAction(toggleAccordion, '')
+            .debounce(300)
+            .map(getIdFromPopup),
+        idStream = Bacon.mergeAll(clickedIdStream, pagingIdStream),
+
+        accordionHtmlStream = BU.fetchFromIdStream(idStream,
```
jwalgran: I ran into a strange bug when trying to use quick edit to "unstack" trees on top of each other.

- Create an instance.
- Add a tree with a species.
- Add a tree with a different species directly on top of the first.
- Refresh the map page.
- Select the tree stack and navigate to the second item in the popup.
- Use Quick Edit to move the tree further than 20ft away.
- Save and refresh the page

Expected: The popup for each dot shows details for a single tree
Actual: The popup for the original dot shows both items. The popup for the moved dot shows just the one item. The plot count does not increase, so it does not look like things are being duplicated.

maurizi: Good catch! This was caused by the `etag` header not including the list of nearby features into the hash (which I didn't notice because I had disabled my browser cache).

Fixed in https://github.com/OpenTreeMap/otm-core/pull/3181/commits/41be44fc321693a61be1b6ae56aeba53f87b92da

Issue #3175, opened 2017-07-27
TravisCI jobs failing to install PostGIS


maurizi:
Output from a failed job:
```
psql -c "CREATE EXTENSION postgis;" -U postgres -d ci

ERROR:  could not access file "$libdir/postgis-2.3": No such file or directory

The command "psql -c "CREATE EXTENSION postgis;" -U postgres -d ci" failed and exited with 1 during .
```

PR #3182, opened 2017-08-09
Fix Travis CI builds

maurizi:
Connects to #3175

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,14 @@ services:
   - redis-server
   - postgresql
 
+dist: trusty
+
+addons:
+  postgresql: "9.5"
+  apt:
+    packages:
+      - postgresql-9.5-postgis-2.3
+
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,7 +49,7 @@ before_script:
 script:
   - flake8 --exclude migrations,opentreemap/settings/local_settings.py,*.pyc opentreemap
   - npm run check
-  - testem ci -l Firefox
```
jwalgran: The `test` script in `package.json` runs `testem ci` and our travis config was running `testem ci -l Firefox`. Are we intentionally dropping the `-l Firefox`?

maurizi: It wasn't my intention, but there isn't really a need to have the `-l Firefox` anyways.

It _was_ there because we had issues w/ PhantomJS on Travis hanging indefinitely, but that seems like it is no longer an issue.

jwalgran: :+1:

Issue #3172, opened 2017-07-19
Registration page has styling issues on small screens


maurizi:
![localhost-6060-accounts-register- 2](https://user-images.githubusercontent.com/4432106/28391050-6686b1de-6caa-11e7-803f-266c3af30b08.png)

PR #3193, opened 2017-08-17
Fix styling issues for registration page on small screens

maurizi:
localhost-6060-accounts-register-

Connects to #3172


jwalgran:
Looks good to me.

PR #3200, opened 2017-08-28
Fix styling issues on management pages

maurizi:
- Make management navigation sidebar scrollable
 - Fix padding on last link collapsible groups
 - Prevent map title from being visible when the sidebar is collapsed

Connects to OpenTreeMap/otm-addons#1460

### DIFF
--- a/assets/css/sass/partials/pages/_admin.scss
+++ b/assets/css/sass/partials/pages/_admin.scss
@@ -64,9 +64,14 @@
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


maurizi:
> I see the scroll fix working but when I collapse the sidebar I still see a sliver of the instance name (Chrome and Firefox on macOS)
Ah, I forgot to test on a map with a long enough name! Fixed in 95fb4ce

Issue #2920, opened 2017-01-06
Species dropdowns omit "Other part of name" from scientific name 


RickMohr:
1) Create a map outside the US
2) In "Search by Species", type "marina arbutus"

Expected: scientific name displayed as "Arbutus var. marina"
Actual: scientific name displayed as "Arbutus"

This means two species differing only in "Other part of name" are indistinguishable.

PR #3201, opened 2017-08-28
Add 'other_part_of_name' (Variety) to species scientific name

maurizi:
To test:

Search for 'Cockspur Hawthorn' on an international map. On the `develop` branch there should two identical entries, and on this branch one entry should include the variety as part of the scientific name.

Connects to #2920


jwalgran:
@azavea-bot rebuild

Issue #1123, opened 2014-02-19
Search box address not brought over from plot detail to map page


steventlamb:
After entering an address and selecting a geocoded location from the plot detail page, you are brought to the map page, but the address you entered is lost.

jmorrison1847:
1. Go to a plot detail page in Philly Tree Map (https://www.opentreemap.org/phillytreemap/features/366764/)
![screen shot 2016-12-20 at 11 55 26 am](https://cloud.githubusercontent.com/assets/12401491/21359784/500e26f8-c6ab-11e6-9b4e-b82b38e2c632.png)

2. Enter an address (990 Spring Garden Street, Philadelphia) and pick the geocoded option from the dropdown menu
![screen shot 2016-12-20 at 11 55 46 am](https://cloud.githubusercontent.com/assets/12401491/21359789/565e29fe-c6ab-11e6-8c61-575daaf124d0.png)

3. Get sent to the approximate address on the map page but without the dropped pin and without the information stored in the search box
![screen shot 2016-12-20 at 11 55 54 am](https://cloud.githubusercontent.com/assets/12401491/21359890/b2c8a62e-c6ab-11e6-99d8-e716c9e0855b.png)

PR #3203, opened 2017-08-29
Store geocoded address in URL when searching and show on page load

maurizi:
Adds a new `a` parameter to the URL which is set when there is a
non-boundary value in the location typeahead.

On page load on the map page, if the `a` parameter is present in the URL
and no boundary filter is present, we geocode the address and display a map
pin (which will also recenter the map on the pin).

Connects to #1123

### DIFF
--- a/opentreemap/treemap/js/src/lib/geocoderUi.js
+++ b/opentreemap/treemap/js/src/lib/geocoderUi.js
@@ -21,14 +21,20 @@ module.exports = function (options) {
             inputs: inputs,
             button: options.searchButton
         }),
+
+        manualUpdates = new Bacon.Bus(),
+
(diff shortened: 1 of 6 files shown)


jwalgran:
I ran into a geocode issue when change the species search after loading the page with an `a=` param.

- On the map page, search for an address.
- Refresh the page. The address and red pin appear correctly.
- Select a species

Expected: No error. The red pin stays on the map.
Actual: "Location not found" toast. The red pin is removed from the map.

maurizi:
> I ran into a geocode issue when change the species search after loading the page with an `a=` param.

Fixed in https://github.com/OpenTreeMap/otm-core/pull/3203/commits/1b65bcf942b7dcc14a0875032840d504dcef7d7f

Issue #2714, opened 2016-08-16
After editing the shape of a polygonal map feature the Last Update section is unclear


jwalgran:
Presumably, this is showing the new are. Is it? If it is, it is most likely showing the raw, internal value, which would be confusing to those who have changed the display units.

<img width="206" alt="screen shot 2016-08-16 at 2 00 01 pm" src="https://cloud.githubusercontent.com/assets/17363/17715636/eab3c0b4-63b9-11e6-9ab5-170134bcfaff.png">

PR #3212, opened 2017-09-04
Calculate area in edits sidebar more accurately and use instance units

maurizi:
Connects to #2714


maurizi:
> What do you think about adding class method that allows use to specify a custom field_display_name

I added this.

Issue #2930, opened 2017-01-12
Plot detail: Users are invited to share a photo when they shouldn't be


RickMohr:
In both of these cases:
1) Uploading a photo for a tree in a private treemap
2) After an image upload error (e.g. uploading a file that isn't an image)

The "Would you like to share this photo?" panel is raised. It should not be raised.

dboyer:
This also happens when uploading a photo that is too large.

PR #3217, opened 2017-09-07
Don't show the tree photo sharing dialog if it is not appropriate

jwalgran:
There are three tree photo upload situations where we do not want to show a social media sharing dialog after POSTing the image to the server.

- The instance is private.
- The image file was invalid.
- The image file was too big.

Private instances are not handled by not rendering the dialog markup on the page unless the instance is public.

Invalid images are handled by checking the markup returned from the response for the presence of an error element. This is not ideal, but is easier that changing the way failures are reported vs. loading updated carousel markup.

Large images are intended to be blocked by Nginx before reaching Django. If they reach Django they will trigger a validation error, which will be handled the same way invalid images are handled.

---

##### Testing

- Create 2 instances, 1 public and 1 private
- Verify that uploading a valid photo to the public instance shows the share dialog.
- Verify that uploading a valid photo to the private instance does not show the share dialog.
- Verify that uploading a non-image to the public instance shows the share dialog.
- Verify that uploading a too large image to the public instance shows the share dialog. 
[large.png.zip](https://github.com/OpenTreeMap/otm-core/files/1285635/large.png.zip)

---

Connects to #2930


jwalgran:
@azavea-bot rebuild

PR #3218, opened 2017-09-07
Add automatically generated migrations

maurizi:
I'm not sure why these weren't created earlier, but they seem to be valid migrations.


maurizi:
@azavea-bot rebuild

Issue #3224, opened 2017-10-17
Search results do not match parameters when searching by unit


dboyer:
1. On testmap1017, I entered three trees with diameters between 3 and 8 inches
2. Completed an advanced search for trees between 3 and 8 inches and received 3 search results
3. Went to Units management page and changed diameter to cm
4. Refreshed search and refreshed map page
5. Clicked on a tree dot, went to tree detail page, saw diameter displaying as cm
6. Searched for trees with a diameter between 3 and 8 cm
7. Returned 3 search results, although only one tree has a diameter between 3 and 8 cm

PR #3225, opened 2017-10-18
Update universal rev after changing instance units.

jwalgran:
## Overview

Calculating ecobenefits for a large number of trees can be expensive, so we cache the results based on query string arguments and the `universal_rev` of the instance. After changing the units of a field, the same query string can produce different results so we must increment the `universal_rev` to bust the cache.

Connects #3224 

## Testing Instructions

 * Create a new instance with the URL name `unit-eco-test` via http://localhost:6060/create
 * Add three trees with the following diameters
   * 3.1in
   * 5in
   * 7.9in
 * Click the "Advanced" button above the "Search" button and in the "Tree" down down enter a range of 3in to 8in.
 * Click "Search" and verify that "3 Trees" appears in the search result
screen shot 2017-10-18 at 10 46 16 am

 * Go to the units management page http://localhost:6060/unit-eco-test/management/units/
   * Click "Edit"
   * Change the units for Tree Diameter to `centimeters`
   * Open the the dev tools network panel.
   * Click Save. Verify that the request succeeds and the URL includes `?update_universal_rev=1`
 * Click "Explore Map" in the header.
 * Run the same advanced search for tree diameter, entering 3 and 8 into the range boxes (the unit label will now be cm.
 * Click "Search" and verify that only "1 Tree" is shown in the search results.
 * Go to the "Basic Information" management page http://localhost:6060/unit-eco-test/management/site-config/
   * Click "Edit"
   * Set a value in the "Contact Email" field
   * Open the the dev tools network panel.
   * Click "Save". Verify that the request does not include an `update_universal_rev` query parameter.

### DIFF
--- a/opentreemap/manage_treemap/views/__init__.py
+++ b/opentreemap/manage_treemap/views/__init__.py
@@ -21,10 +21,15 @@ def update_instance_fields_with_validator(validation_fn):
 
 def update_instance_fields(request, instance, validation_fn=None):
     json_data = json_from_request(request)
-    return _update_instance_fields(json_data, instance, validation_fn)
+    # The update is a PUT request but the query string param appears in the GET
+    # dict
+    should_update_universal_rev = 'update_universal_rev' in request.GET
+    return _update_instance_fields(json_data, instance, validation_fn,
+                                   should_update_universal_rev)
 
 
-def _update_instance_fields(json_data, instance, validation_fn=None):
+def _update_instance_fields(json_data, instance, validation_fn=None,
+                            should_update_universal_rev=False):
     error_dict = {}
     for identifier, value in json_data.iteritems():
         model, field_name = dotted_split(identifier, 2, maxsplit=1)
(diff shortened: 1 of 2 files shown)


opentreemap/manage_treemap/views/__init__.py
```diff
@@ -47,6 +51,8 @@ def _update_instance_fields(json_data, instance, validation_fn=None):
         try:
             validate_is_public(instance)
             instance.save()
+            if should_update_eco_rev:
+                instance.update_eco_rev()
```
maurizi: You should also update the `universal_rev` here: see https://github.com/OpenTreeMap/otm-core/blob/develop/opentreemap/treemap/instance.py#L237-L241 for an explanation

jwalgran:
@mmcfarland I added a fixup to implement a comment from Mike.

opentreemap/manage_treemap/views/__init__.py
jwalgran: Updated. Thanks.

opentreemap/manage_treemap/views/__init__.py
```diff
@@ -21,10 +21,14 @@ def f(*args, **kwargs):
 
 def update_instance_fields(request, instance, validation_fn=None):
     json_data = json_from_request(request)
-    return _update_instance_fields(json_data, instance, validation_fn)
+    # The update is a PUT request but the query string param appears in the GET dict
```
mmcfarland: Looks like you're getting dinged by the linter for the length of this line.

PR #3227, opened 2018-01-05
Fix README format

Larry850806:


jwalgran:
Thanks for your contribution.

Issue #1672, opened 2014-09-30
No option to change password from user profile


maurizi:
The is a link to "reset password", but it requires clicking a link in an email.

If you are already logged in, you should be able to change your password by providing the old one.

PR #3235, opened 2018-02-24
Added option to change password from user profile

bryant100000:
Connects to #1672


jwalgran:
Thanks for your contribution to OpenTreeMap.

PR #3236, opened 2018-02-24
made change in user.py code to make user profile not to be case sensitive

vmanish19:
connects to #1742


jwalgran:
Thank you for the contribution. We need to wait on merging this until we resolve some other case sensitivity related issues.

Issue #2450, opened 2016-01-14
Tree diameter validation message is not shown directly below the diameter field


jwalgran:
It is below a "need help..." link, which may be confusing.

<img width="326" alt="screen shot 2016-01-14 at 1 07 33 pm" src="https://cloud.githubusercontent.com/assets/17363/12336441/e24f18de-bac1-11e5-8489-fcbd9ff92483.png">

PR #3237, opened 2018-02-24
Fixed issue diameter value is not showing directly below diameter field.

vmanish19:
connects to #2450. When adding tree error message will be displayed above video link and an error message will display after video link in edit tab.


jwalgran:
Thank you for the contribution.

PR #3239, opened 2018-02-25
Store current base map in local storage

vmanish19:
connect to #832. Using local storage API


jwalgran:
Thank you for the contribution.

PR #3243, opened 2018-03-10
Update moment to 2.19.3

jwalgran:
## Overview

Incorporates bug fixes.

Connects https://github.com/OpenTreeMap/otm-cloud/issues/432

### Setup

* `vagrant provision app` to install the updated dependency.
* Create an instance via http://localhost:6060/create

### Moment

* Add a tree and choose the "continue editing" option to move to the detail edit page
* Enter a "Date Planted" value and save the tree
* Advanced Search for a date planted range and verify that the tree is matched.
* Edit and safe the Date Planted and verify that the value is correct after a refresh.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "leaflet.gridlayer.googlemutant ": "~0.4.3",
     "leaflet.locatecontrol": "^0.62.0",
     "lodash": "^4.17.4",
-    "moment": "~2.11.2",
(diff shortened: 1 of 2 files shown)


jwalgran:
Matt is not available. There are no actual code changes, and I ran through the manual tests without and problems so I feel we can rely on release QA as a substitute review.

PR #3244, opened 2018-03-22
 Incorporate the basemap provider into selected basemap storage

jwalgran:
## Overview

OTM supports multiple basemap providers, each with different basemap names. When we store the selected basemap, we need to namespace that selection under the provider name.

Connects https://github.com/OpenTreeMap/otm-cloud/issues/438

## Testing Instructions

- Browse the map page of an existing instance
- Change the basemap to satellite
- Refresh the page and verify that the basemap change persists
- Browse `/create` and work through the instance creation wizard
- Verify that the Bing basemap correctly appears on the 3rd page of the create wizard

### DIFF
--- a/opentreemap/treemap/js/src/lib/MapManager.js
+++ b/opentreemap/treemap/js/src/lib/MapManager.js
@@ -146,7 +146,7 @@ MapManager.prototype = {
             bounds = options.bounds,
             map = L.map(options.domId),
             type = options.type,
-            basemapMapping = getBasemapLayers(type);	
+            basemapMapping = getBasemapLayers(type);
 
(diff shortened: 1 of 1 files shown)


jwalgran:
Blocked on a fix for https://github.com/OpenTreeMap/otm-cloud/issues/438

dboyer:
@jwalgran  I finally had a chance to check this and I was able to change the basemap on an existing map and then create a new map. If it matters, it looks like the basemap change impacts any map you have open in your browser. I had an old map set to satellite layer in one tab and my newly created map set to streets layer in another tab. When I refreshed the tab with the old map, it switched to the streets layer.

Issue #3245, opened 2018-04-10
Date filter on Tree Care advanced search behaving strangely


jmorrison1847:
![apr-10-2018 19-02-07](https://user-images.githubusercontent.com/12401491/38588054-dfec9292-3cf1-11e8-93e7-71ba11fe9a52.gif)
When you update the date filter for the terminal date, it changes the original date. You can test it out on https://www.opentreemap.org/greenprintmaps/map/.

jmorrison1847:
Seems to be the same for all maps

dboyer:
I am also seeing this issue. It renders the feature fairly unusable so it would be helpful to get in a fix.

PR #3246, opened 2018-05-03
Use strict mode when using moment.isValid to check date formatting.

jwalgran:
### Overview

Dates on the advanced search toolbar were being set incorrectly because the parsing a short date with the long date format was being interpreted as valid. Enabling "strict mode" makes the date parsing resolves this.

Connects #3245 

### Demo

Before
screen shot 2018-05-03 at 11 42 53 am

After
screen shot 2018-05-03 at 11 43 27 am

### Testing

 * Visit production https://www.opentreemap.org/phillytreemap/map/
 * Open the advanced search menu, click on Tree Care, and select any values for the first 3 dropdowns
 * Select May 1 in the date pickers and verify the bad behavior
 * Repeat the process with a local instance and verify that setting May 1 in the date pickers works.


jwalgran:
@azavea-bot rebuild

Issue #3250, opened 2018-07-11
Map layer selector does not open on iPad


dboyer:
Reported by a client.

1. Open any tree map on an iPad. 
2. Click the map layer selector.
3. Nothing happens

A client reported this using the most recent iOS version. I saw the same issue on my ancient iPad running version 9.3.5.

jwalgran:
I reproduced this in the simulator with multiple iOS versions.

jwalgran:
When using a desktop browser, the layer selector appears on hover. We need to update event handling so that a click will also trigger the layer selector.

jwalgran:
I created a test page with the current OTM production Leaflet version (1.0.3) and the layer selector works in the iOS simulator.

```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="UTF-8"/>
        <title>Leaflet map</title>
        <link href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.0.3/leaflet.css" rel="stylesheet"/>
        <style>
         body {
             padding: 0;
             margin: 0;
         }
         html, body, #map {
             height: 100%;
             width: 100%;
         }
        </style>
    </head>
    <body>
        <div id="map"></div>
        <script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.0.3/leaflet-src.js"></script>
        <script>
         var map = L.map('map').setView([ 39.961318, -75.154188 ], 12);
         var layers = {
             'Light': L.tileLayer(
                 'https://cartodb-basemaps-a.global.ssl.fastly.net/light_all/{z}/{x}/{y}@2x.png', {
                     maxZoom: 18,
                 }),
             'Dark': L.tileLayer(
                 'https://cartodb-basemaps-a.global.ssl.fastly.net/dark_all/{z}/{x}/{y}@2x.png', {
                     maxZoom: 18,
                 })
         }
         layers['Light'].addTo(map);
         var layersControl = L.control.layers (layers, null, {
             autoZIndex: false
         }).addTo(map);
        </script>
    </body>
</html>
```

jwalgran:
I attached the Safari debugger to the iOS Simulator and set breakpoints inside the Leaflet code. It appears that click events on the layer selector are not making it to Leaflet. Breakpoints set on the attached handler functions are never hit.

jwalgran:
[...]

PR #3253, opened 2018-07-31
Monkey patch the Leaflet layer selector to work in Mobile Safari

jwalgran:
## Overview

Tapping on the layers control in Mobile Safari was not opening the layer selector. I was not able to reproduce this issue in a separate test application, only within OTM.

After much trial and error I discovered that changing this line
```javascript
  L.DomEvent.on(el, L.Draggable.START.join(' '), stop);
```
to
```javascript
  L.DomEvent.on(container, 'mousedown', stop);
```
in `DomEvent.disableClickPropagation` resolved the issue. Effectively, this prevents the `touchstart` event from being stopped.

I opted to implement this as a runtime patch applied within MapManager instead of a global patch to the Leaflet source because there are no other Leaflet behaviors that are known to be broken in Mobile Safari and I did not want to risk introducing a regression in some other event handling.

I minimized the chance of forgetting about this patch if/when we upgrade our version of Leaflet by adding a comment to package.json.

Connects #3250 

## Setup

 * Ensure that your development server is accessible from other machines on the network.
 * Obtain an iOS device for testing (hardware or simulator). Browse https://www.opentreemap.org/phillytreemap in Mobile Safari and verify that tapping on the layer selector does not work.

## Test

 * Run `vagrant provision app` and verify that it completes without error.
 * Browse an instance on your development server in Mobile Safari and verify that the layer selector works.

### DIFF
--- a/opentreemap/treemap/js/src/lib/MapManager.js
+++ b/opentreemap/treemap/js/src/lib/MapManager.js
@@ -30,6 +30,90 @@ require('leaflet.locatecontrol');
 
 var MapManager = function() {};  // constructor
 
+function monkeyPatchLeafletLayersControlForMobileSafari(layersControl) {
+    /*
+      WARNING: This method will likely break if Leaflet is upgraded.
+
+      Tapping on the layers control in Mobile Safari was not opening the layer
+      selector. I was not able to reproduce this issue in a separate test
+      application, only within OTM.
+
+      After much trial and error I discovered that changing this line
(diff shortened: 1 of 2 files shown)


jwalgran:
The comment in the dependency section of the `package.json` file did not work. I am going to try moving it up to the top level.

jwalgran:
I force pushed a commit that resolved the build failure.

jwalgran:
Thanks for the review.

Issue #3255, opened 2018-09-28
Attempting to restore a basemap layer selection fails if the basemap type is changed


jwalgran:
Storing basemap preferences was added in #3239.

Steps to reproduce:

* Browse an instance with the `google` basemap.
* Switch to satellite view.
* Update the instance `basemap_type` to `bing`.
* Refresh the map page

<img width="1680" alt="screen shot 2018-09-28 at 4 02 46 pm" src="https://user-images.githubusercontent.com/17363/46237234-fc669900-c337-11e8-9987-16c183bfd835.png">

PR #3256, opened 2018-09-29
Handle basemap preference not available

jwalgran:
Avoids an unhandled exception when the `basemap_type` of an instance has changed and a previous layer selection persisted in local storage no longer exists.

Connects #3255 

### Testing

* Browse an instance with the `google` basemap.
* Select the `Satellite` layer
* Refresh the page and verify that the `Satellite` layer is displayed
* Update the instance `basemap_type` to `bing`.
```
otm=# update treemap_instance set basemap_type = 'google', universal_rev = universal_rev + 1;
```
* Refresh the map page
* Verify that the street layer loads and no error is logged in the JS console.


jwalgran:
Merged early to expedite getting production working again.

PR #3262, opened 2018-11-06
Add client-side validation to prevent extra udf choice whitespace

jwalgran:
## Overview

Leading or trailing whitespace cannot be easily seen, but causes problems matching values. This change follows the pattern of other client-side choice validations.

Connects https://github.com/OpenTreeMap/otm-clients/issues/442

### Demo

screen shot 2018-11-06 at 11 21 45 am

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

* Visit http://localhost:6060/create and create an instance named `trailing-choice`
* Visit http://localhost:6060/trailing-choice/management/udfs/, click `Edit` and verify that an error message is displayed and the save button is disabled if an existing choice or a new choice has leading or trailing whitespace
* Verify that new and edited choices can still be saved if they do not have excess whitespace.

### DIFF
--- a/assets/css/sass/partials/pages/_admin.scss
+++ b/assets/css/sass/partials/pages/_admin.scss
@@ -904,6 +904,11 @@
                 display: block;
             }
         }
+        &.whitespace-error {
+            [data-item="whitespace"] {
+                display: block;
+            }
+        }
         &.blank-error {
             [data-item="blank"] {
                 display: block;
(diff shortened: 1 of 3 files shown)


jwalgran:
Thanks. I was focued on editing existing fields and overlooked the new field dialog.

I intentionally did not dig into validating existing fields that may have whitespace issues to keep the scope of the change small and because we are planning on cleaning up existing data via scripting.

jwalgran:
I added a FIXUP commit to properly disable the button on the new choice field dialog when there is a client side validation error.

jwalgran:
Thanks for the thorough review.

Issue #3265, opened 2018-12-21
Cannot delete a tree if the tree ID field is hidden.


jwalgran:
- Create an instance named `cant-delete-without-tree-id`
- Browse http://localhost:6060/cant-delete-without-tree-id/management/field-configuration/
- Click "Edit"
- Uncheck "ID" in the "Tree" section.
- Click "Save"
- Click "Add a tree." Complete the add tree process, filling in a species and diameter.
- Browse the detail page for the tree
- Click "Delete" and "Confirm Deletion"

Expected: The tree is deleted and the planting site remains
Actual: The API returns a 400 and the tree is not deleted.

This is where `plotDetail.js` is creating a delete URL by looking for the presence of the tree ID in the DOM.

https://github.com/OpenTreeMap/otm-core/blob/dc0428c468707a38ea41176204b527b1832e9a8d/opentreemap/treemap/js/src/lib/plotDetail.js#L31-L47

This is the root cause of https://github.com/OpenTreeMap/otm-clients/issues/439

PR #3266, opened 2018-12-21
Read tree ID from detail page tree section header rather than field

jwalgran:
## Overview

Instance managers can hide the tree ID field from view, which was breaking the delete logic.

Connects #3265 
Connects https://github.com/OpenTreeMap/otm-clients/issues/439
Depends on https://github.com/OpenTreeMap/otm-cloud/pull/467

## Testing Instructions

- Create an instance named `delete-without-tree-id-test`
- Browse http://localhost:6060/delete-without-tree-id-test/management/field-configuration/
- Click "Edit"
- Uncheck "ID" in the "Tree" section.
- Click "Save"
- Click "Add a tree." Complete the add tree process, filling in a species and diameter.
- Browse the detail page for the tree
- Click "Delete" and "Confirm Deletion". Verify that the tree is deleted without error.

### DIFF
--- a/opentreemap/treemap/js/src/lib/plotDetail.js
+++ b/opentreemap/treemap/js/src/lib/plotDetail.js
@@ -14,7 +14,6 @@ var $ = require('jquery'),
 
 var dom = {
     form: '#map-feature-form',
-    treeIdColumn: '#tree-id-column',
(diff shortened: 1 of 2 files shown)


jwalgran:
Thanks.

Issue #3264, opened 2018-12-21
Cannot show the tree ID field after it is hidden


jwalgran:
- Create an instance named `tree-id-hidden-forever`
- Browse http://localhost:6060/tree-id-hidden-forever/management/field-configuration/
- Click "Edit"
- Uncheck "ID" in the "Tree" section.
- Click "Save"
- Refresh the page and click "Edit"

Expected: The "ID" field is in shown in the edit list unchecked
Actual: The "ID" field does not appear

![2018-12-21 10 08 35](https://user-images.githubusercontent.com/17363/50354345-75752700-0508-11e9-978d-dae983ab22bb.gif)

PR #3267, opened 2018-12-21
Allow a hidden Tree ID field to be shown

jwalgran:
## Overview

The management page was allowing the Tree ID field to be hidden, but was not rendering the option in the hidden field list so that it could be shown again. We have corrected the `_should_show_field` predicate function and corrected the model argument passed to it so that the Tree ID field is correctly shown in the disabled field list.

Connects #3264 
Depends on https://github.com/OpenTreeMap/otm-cloud/pull/467

## Testing Instructions

- Create an instance named `tree-id-show-and-hide-test`
- Browse http://localhost:6060/tree-id-show-and-hide-test/management/field-configuration/
- Click "Edit"
- Uncheck "ID" in the "Tree" section.
- Click "Save"
- Refresh the page and click "Edit". Verify that "ID" appears in the "Tree" section.
- Check "ID" and click "Save"
- Refresh the page and verify that "ID" appears in the "Tree" section.

### DIFF
--- a/opentreemap/manage_treemap/views/fields.py
+++ b/opentreemap/manage_treemap/views/fields.py
@@ -9,7 +9,7 @@ from django.utils.translation import ugettext as _
 
 from opentreemap.util import json_from_request
 
-from treemap.models import Species
+from treemap.models import Species, Tree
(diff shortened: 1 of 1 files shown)


jwalgran:
Thanks.

PR #3269, opened 2019-01-03
Update urllib3 from 1.22 to 1.23

jwalgran:
## Overview

Includes bug fixes. A transitive dependency of [requests](http://python-requests.org/).

Connects https://github.com/OpenTreeMap/otm-cloud/issues/466

### Notes

Searching through the core, cloud, and addons code bases using the regex `import.*requests` returns just 2 usages.

 * A geocoding unit test.
 * An non-critical add-on utility for POSTing data to Salesforce 

## Testing Instructions

 * Reprovision and `pip freeze` to ensure that the updated version is installed.
 * Verify that the unit tests pass.


jwalgran:
Thanks.

PR #3270, opened 2019-01-09
Add django-recaptca to the user registration page

jwalgran:
## Overview

We are attempting to prevent spam account registrations by adding a robot test via Google's reCAPTCHA service and the `django-recaptcha` library.

Connects https://github.com/OpenTreeMap/otm-cloud/issues/461
Depends on https://github.com/OpenTreeMap/otm-cloud/pull/469

### Demo

screen shot 2019-01-08 at 7 14 43 pm

### Notes

From https://developers.google.com/recaptcha/docs/faq

>reCAPTCHA v2 is not going away! We will continue to fully support and improve security and usability for v2.

>reCAPTCHA v3 is intended for power users, site owners that want more data about their traffic, and for use cases in which it is not appropriate to show a challenge to the user.

## Testing Instructions

NOTE: This is currently deployed to staging: https://staging.opentreemap.org/accounts/register/

 * Log out and visit the registration page at `/accounts/register/`
 * Verify that the reCAPTCHA is shown and that the registration form cannot be submitted without checking the I am not a robot box.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ django-apptemplates==1.3
 django-contrib-comments==1.8.0
 django-js-reverse==0.7.3
 django-queryset-csv==1.0.0               # https://github.com/azavea/django-queryset-csv/commits/master
+django-recaptcha==1.4.0
 django-redis==4.8.0
 django-registration-redux==1.7
 # django-statsd-mozilla==0.3.16  # TODO: enable when compatible with Django > 1.8 https://github.com/django-statsd/django-statsd/issues/97
--- a/opentreemap/opentreemap/settings/default_settings.py
+++ b/opentreemap/opentreemap/settings/default_settings.py
@@ -277,6 +277,7 @@ INSTALLED_APPS = (
     'django.contrib.contenttypes',
     'django.contrib.sites',
     'django.contrib.auth',
+    'captcha',
     'treemap',
     'geocode',
     'api',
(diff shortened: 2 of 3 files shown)


jwalgran:
I will be working on fixing the CI failures

requirements.txt
```diff
@@ -14,6 +14,7 @@ django-apptemplates==1.3
 django-contrib-comments==1.8.0
 django-js-reverse==0.7.3
 django-queryset-csv==1.0.0               # https://github.com/azavea/django-queryset-csv/commits/master
+django-recaptcha==1.4.0
```
hectcastro: You may already know about this, but while I was Googling to look at the environment variables available for this plugin, I stumbled upon https://github.com/kbytesys/django-recaptcha3.

Based on what I read, still seems like v2 (appears to be what we're using) is a safe bet.

jwalgran: Yes. This looked like the most well used library with the smallest footprint and Google has no plans to shut off v2. I added the following quotes from the [FAQ](https://developers.google.com/recaptcha/docs/faq) to the NOTES section of the PR description

>reCAPTCHA v2 is not going away! We will continue to fully support and improve security and usability for v2.

>reCAPTCHA v3 is intended for power users, site owners that want more data about their traffic, and for use cases in which it is not appropriate to show a challenge to the user.

jwalgran:
Thanks.

PR #3275, opened 2019-10-14
Fix test suite issues related to OS upgrade

hectcastro:
When checking to determine the ecoservice status via `systemd`, ensure that Python decodes the command output as UTF-8 instead of ASCII because the `systemd` output contains non-ASCII values.

Also, alter the approach when making area and runoff reduction assertions. Aim to be within a small percentage of the expected value vs. matching the integer-parts
exactly.

Depends on https://github.com/OpenTreeMap/otm-cloud/pull/487

**Testing**

Please use https://github.com/OpenTreeMap/otm-cloud/pull/487 to review this PR.

### DIFF
--- a/opentreemap/importer/tests.py
+++ b/opentreemap/importer/tests.py
@@ -1151,7 +1151,7 @@ class TreeIntegrationTests(IntegrationTests):
 
         p2_geom = plot2.geom
         p2_geom.transform(4326)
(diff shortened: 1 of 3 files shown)


hectcastro:
@azavea-bot rebuild

mmcfarland (approved):
These assumptions look reasonable. The delta is extremely small and the use of adjacent values for a coordinate check has precedence in the codebase.

PR #3283, opened 2020-01-03
Update requirements for Python 3 compatibility

jwalgran:
## Overview

Update requirements for Python 3 compatibility 

This PR is targeted at a `python3` branch rather than `develop`. This will be the first in a set of PRs that convert the application to run exclusively under Python 3. Merging this PR will break the application until it is updated to use Python 3 syntax.

Required by https://github.com/OpenTreeMap/otm-cloud/pull/501
Connects https://github.com/OpenTreeMap/otm-cloud/issues/500

## Notes

Django 1.11.17 is the minimum version required to run under Python 3.
https://docs.djangoproject.com/en/1.11/faq/install/#what-python-version-can-i-use-with-django

`functools32` was a backport of a Python 3 included library.

The 4.2.0 version of kombu fixes a syntax error under Python 3.
https://github.com/celery/kombu/issues/841

wsgiref is part of the Python 3 standard library.

## Testing Instructions

This should be tested as part of https://github.com/OpenTreeMap/otm-cloud/pull/501

## Checklist

- [ ] `fixup!` commits have been squashed
- [ ] CI passes after rebase

### DIFF
--- a/opentreemap/manage.py
+++ b/opentreemap/manage.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3.7
 import os
 import sys
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ celery==4.1.0
 certifi==2017.7.27.1
 chardet==3.0.4
 # https://docs.djangoproject.com/en/1.10/releases/#id2
-Django==1.11.16  # rq.filter: >=1.11,<1.12
+Django==1.11.17  # rq.filter: >=1.11,<1.12
 django-apptemplates==1.3
 django-contrib-comments==1.8.0
 django-js-reverse==0.7.3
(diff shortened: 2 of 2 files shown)


jwalgran:
@rbreslow Requesting a review from you because of your involvement on previous Python 3 transitions.

jwalgran:
I just pushed a fixup that replaces a Python 2 backport in requirements.txt with the main version of the library that supports Python 3.

https://github.com/OpenTreeMap/otm-core/pull/3283/commits/fcd6fb6bb8523e53f787feed6c980d28f36ef0dc

fungjj92:
Do we plan on upgrading to the latest versions of all requirements either in this PR or a followup?

In [initial research on dependencies](https://github.com/OpenTreeMap/otm-cloud/issues/499#issuecomment-569288198), I had checked py3 version compatibility _in their latest releases_. To test our current versions with this PR I ran `caniusepython3` on the 3 requirements files and we're golden for py3 on all current versions, but not necessarily 3.7. For example, for `django-contrib-comments` we are on `1.8.0` which [includes testing through py3.5 but not for py3.7](https://pypi.org/project/django-contrib-comments/1.9.2/) until the latest version, `1.9.2`.

jwalgran:
Excellent question. I had focused this PR on just the things that were required to get the development provisioning working. I had planned that updating code and requirements would be the next steps we take in a follow up PR.

fungjj92 (approved):
Makes sense to me! I didn't check issues when I asked, but now  see that you had already made https://github.com/OpenTreeMap/otm-core/issues/3284

opentreemap/manage.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3.7
```
hectcastro, line 1: This change is connected to the changes and justification included in: https://github.com/OpenTreeMap/otm-cloud/pull/501/files#r363948469.

jwalgran: Thanks. I agree with the changes you made in OpenTreeMap/otm-cloud/pull/501

hectcastro (approved):
Going to merge these changes now. I discussed merging them with Justin yesterday afternoon. 👍

PR #3288, opened 2020-01-22
First pass at fixing tests from Py2->3 upgrade

fungjj92:
### Overview
This PR is a first go at fixing the failing tests from the Py2->3 upgrade.
The tests should be fixed for the `exporter` app 🎉 One down.

I touched tests / code across the django apps `api, exporter, importer, treemap`. Except for `exporter`, they continue to have errors. I'm not sure how the test runner selects the order of tests to run because it skipped around these 4 apps when running `./scripts/manage.py test --failfast`

Most of the changes thus far have been about unicode/byte - string translation.
The celery and Django-tinsel upgrades were to fix test-related errors, see commit https://github.com/OpenTreeMap/otm-core/commit/df782b1e40ab988e27bede383616c0587344a2c2

### Notes
In retrospect, I would have worked on one app's tests at a time. My advice to parallelize this work moving forward is just that:
`./scripts/manage.py test <app_name> --failfast`

It's impossible to linearly compare progress for various reasons. But anyway, we're making progress! 

Before:
```
Ran 1115 tests in 275.613s

FAILED (failures=19, errors=305, skipped=23)
```

And this is what remains after related-PRs 😳 
```
Ran 1115 tests in 331.760s

FAILED (failures=26, errors=125, skipped=23)
```
### Testing

In the app VM, pip install the new python dependencies:
`vagrant ssh app -c 'cd /opt/app/core && envdir /etc/otm.d/env pip3 install -r ./requirements.txt'`

Checkout the otm-addons companion branch https://github.com/OpenTreeMap/otm-addons/pull/1566

I suppose you'll want to test each of the apps that was touched and check if any still failling tests are a result of lines of codes touched in this PR.
`./scripts/manage.py test <app_name>`

Also ensure that the changes don't chip away important context just to get tests to work.

### DIFF
--- a/opentreemap/api/auth.py
+++ b/opentreemap/api/auth.py
@@ -59,7 +59,10 @@ def get_signature_for_request(request, secret_key):
         ).digest()
     )
 
-    return sig
+    if sig is None:
+        return sig
+
+    return sig.decode()
 
 
 def create_401unauthorized(body="Unauthorized"):
--- a/opentreemap/api/auth.py
+++ b/opentreemap/api/auth.py
@@ -9,7 +9,7 @@ import urllib.request
 import urllib.parse
 import urllib.error
 
-from django.http import HttpResponse
+from django.http import HttpResponse, RawPostDataException
 from django.contrib.auth import authenticate
 
 
@@ -37,17 +37,27 @@ def get_signature_for_request(request, secret_key):
 
     sign_string = '\n'.join([httpverb, hostheader, request_uri, paramstr])
 
-    # Sometimes reeading from body fails, so try reading as a file-like
+    # Sometimes reading from body fails, so try reading as a file-like object
     try:
-        body_encoded = base64.b64encode(request.body)
-    except:
-        body_encoded = base64.b64encode(request.read())
+        body_decoded = base64.b64encode(request.body).decode()
+    except RawPostDataException:
+        body_decoded = base64.b64encode(request.read()).decode()
 
-    if body_encoded:
-        sign_string += body_encoded
+    if body_decoded:
+        sign_string += body_decoded
+
+    try:
+        binary_secret_key = secret_key.encode()
+    except (AttributeError, UnicodeEncodeError):
+        binary_secret_key = secret_key
 
     sig = base64.b64encode(
-        hmac.new(secret_key, sign_string, hashlib.sha256).digest())
+        hmac.new(
+            binary_secret_key,
+            sign_string.encode(),
+            hashlib.sha256
+        ).digest()
+    )
 
     return sig
 
@@ -67,11 +77,15 @@ def firstmatch(regx, strg):
         return m.group(1)
 
 
-def decodebasicauth(strg):
-    if strg is None:
+def split_basicauth(strg):
+    """
+    Returns username, password from decoded,
+    stringified, basic auth credentials
+    """
+    if strg is None or len(strg) == 0:
         return None
     else:
-        m = re.match(r'([^:]*)\:(.*)', base64.decodestring(strg))
+        m = re.match(r'([^:]*)\:(.*)', strg)
         if m is not None:
             return (m.group(1), m.group(2))
         else:
@@ -79,7 +93,22 @@ def decodebasicauth(strg):
 
 
 def parse_basicauth(authstr):
-    auth = decodebasicauth(firstmatch('Basic (.*)', authstr))
+    string_wrapped_binary_credentials = firstmatch("Basic (.*)", authstr)
+    if string_wrapped_binary_credentials is None:
+        return None
+
+    # tease bytes-like object out of string, i.e. "b'credentials'"
+    reg_exp = r"'(.*?)\'"
+    parsed_credentials = re.search(r"'(.*?)\'", string_wrapped_binary_credentials)
+    str_credentials = parsed_credentials.groups()[0]
+
+    # decode the binary encoded credentials
+    decoded_str_credentials = base64.decodebytes(
+        bytes(str_credentials, 'utf-8')
+    ).decode()
+
+    auth = split_basicauth(decoded_str_credentials)
+
     if auth is None:
         return None
     else:
--- a/opentreemap/api/plots.py
+++ b/opentreemap/api/plots.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.shortcuts import get_object_or_404
 from django.contrib.gis.geos import Point
 from django.contrib.gis.measure import D
+from django.http import RawPostDataException
 
 from django_tinsel.exceptions import HttpBadRequestException
 
@@ -50,8 +51,7 @@ def plots_closest_to_point(request, instance, lat, lng):
         distance = float(request.GET.get(
             'distance', settings.MAP_CLICK_RADIUS))
     except ValueError:
-        raise HttpBadRequestException(
-            'The distance parameter must be a number')
+        raise HttpBadRequestException('The distance parameter must be a number')
 
     plots = Plot.objects.distance(point)\
                         .filter(instance=instance)\
@@ -72,7 +72,16 @@ def update_or_create_plot(request, instance, plot_id=None):
     # The API communicates via nested dictionaries but
     # our internal functions prefer dotted pairs (which
     # is what inline edit form users)
-    request_dict = json.loads(request.body)
+    # Sometimes reading from body fails, so try reading as a file-like object
+    try:
+        body_decoded = request.body.decode()
+    except RawPostDataException:
+        body_decoded = request.read().decode()
+
+    if body_decoded:
+        request_dict = json.loads(body_decoded)
+    else:
+        request_dict = {}
 
     data = {}
 
--- a/opentreemap/api/tests.py
+++ b/opentreemap/api/tests.py
@@ -1884,7 +1884,7 @@ class SigningTest(OTMTestCase):
         acred = APIAccessCredential.create()
         url = ('http://testserver.com/test/blah?'
                'timestamp=%%s&'
-               'k1=4&k2=a&access_key=%s' % acred.access_key)
+               'k1=4&k2=a&access_key=%s' % acred.access_key.decode())
 
         curtime = datetime.datetime.now()
         invalid = curtime - datetime.timedelta(minutes=100)
@@ -1950,14 +1950,14 @@ class SigningTest(OTMTestCase):
 
         url = ('http://testserver.com/test/blah?'
                'timestamp=%%s&'
-               'k1=4&k2=a&access_key=%s' % acred.access_key)
+               'k1=4&k2=a&access_key=%s' % acred.access_key.decode())
 
         req = self.sign_and_send(url % ('%sFAIL' % timestamp),
                                  acred.secret_key)
 
         self.assertEqual(req.status_code, 400)
 
-        req = self.sign_and_send(url % timestamp, acred.secret_key)
+        req = self.sign_and_send(url % timestamp, acred.secret_key.decode())
 
         self.assertRequestWasSuccess(req)
 
@@ -1973,7 +1973,7 @@ class SigningTest(OTMTestCase):
 
         self.assertEqual(req.status_code, 400)
 
-        req = self.sign_and_send('%s&access_key=%s' % (url, acred.access_key),
+        req = self.sign_and_send('%s&access_key=%s' % (url, acred.access_key.decode()),
                                  acred.secret_key)
 
         self.assertRequestWasSuccess(req)
@@ -1988,9 +1988,8 @@ class SigningTest(OTMTestCase):
         req = self.sign_and_send('http://testserver.com/test/blah?'
                                  'timestamp=%s&'
                                  'k1=4&k2=a&access_key=%s' %
-                                 (timestamp, acred.access_key),
-                                 acred.secret_key)
-
+                                 (timestamp, acred.access_key.decode()),
+                                 acred.secret_key.decode())
         self.assertEqual(req.user.pk, peon.pk)
 
 
--- a/opentreemap/api/tests.py
+++ b/opentreemap/api/tests.py
@@ -1569,7 +1569,7 @@ class TreePhotoTest(LocalMediaTestCase):
                                                       self.instance.url_name,
                                                       plot_id)
 
-        with open(path) as img:
+        with open(path, 'rb') as img:
             req = self.factory.post(
                 url, {'name': 'afile', 'file': img})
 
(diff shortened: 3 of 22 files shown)


opentreemap/api/auth.py
```diff
@@ -68,18 +78,23 @@ def firstmatch(regx, strg):
...
+        m = re.match(r'([^:]*)\:(.*)', base64.encodebytes(strg).decode())
         if m is not None:
             return (m.group(1), m.group(2))
         else:
             return None
 
 
 def parse_basicauth(authstr):
-    auth = decodebasicauth(firstmatch('Basic (.*)', authstr))
+    firstmatch("Basic b'(.*)'", authstr)
```
jwalgran: This line without an assignment looks like it may be safely removed.

opentreemap/api/auth.py
```diff
@@ -68,18 +78,23 @@ def firstmatch(regx, strg):
 
 
 def decodebasicauth(strg):
-    if strg is None:
+    if strg is None or not len(strg):
```
jwalgran: I personally try to avoid type coercion when possible, but I had trouble searching for which method is more Pythoic. Consider `if strg is None or len(strg) == 0` if you also prefer it.

opentreemap/api/plots.py
```diff
@@ -72,7 +71,16 @@ def update_or_create_plot(request, instance, plot_id=None):
     # The API communicates via nested dictionaries but
     # our internal functions prefer dotted pairs (which
     # is what inline edit form users)
-    request_dict = json.loads(request.body)
+    # Sometimes reading from body fails, so try reading as a file-like object
+    try:
+        body_decoded = request.body.decode()
+    except:
```
jwalgran: It is generally preferable to catch an explicit exception type in these cases to avoid "swallowing" an exception that is unrelated to the body reading (`MemoryError` for example)

requirements.txt
```diff
@@ -19,7 +19,7 @@ django-redis==4.8.0
 django-registration-redux==1.7
 django-storages==1.6.5
 django-threadedcomments==1.1
-django-tinsel==1.0.1
+#django-tinsel==1.0.2
```
jwalgran: It looks like this may be commented out by mistake.

opentreemap/treemap/instance.py
```diff
@@ -153,7 +153,7 @@ def create_from_geojson(cls, geojson):
         try:
             geojson_dict = json.loads(geojson)
         except ValueError as e:
-            raise ValidationError(e.message)
+            raise ValidationError(str(e))
```
jwalgran, line 156: Consider rewording the commit message to substitute "attribute" for "argument."

>The message attribute (and all attributes other than `args`) were removed from the Exception base class between Py2 and Py3.

This more closely matches the text in https://www.python.org/dev/peps/pep-0352/

>This PEP proposes introducing a new exception named BaseException that is a new-style class and has a single attribute, args.

opentreemap/api/tests.py
```diff
@@ -369,14 +370,13 @@ def test_locations_plots_endpoint(self):
         self.assertEqual(response.status_code, 200)
 
     def test_locations_plots_endpoint_max_plots_param_must_be_a_number(self):
-        response = get_signed(
+        self.assertRaises(HttpBadRequestException, get_signed(
```
jwalgran: We removed the `response = ...` assignment above, so these `assertEqual` statements are now checking a `response` variable that is not set. The same test uses a similar construction, but does not switch to an `assertRaises`. It looks like this may be a mistake.

This test is failing with an exception in my dev environment
```
> ./scripts/manage.sh test api.tests.Locations.test_locations_plots_endpoint_max_plots_param_must_be_a_number
+ ARGS='test api.tests.Locations.test_locations_plots_endpoint_max_plots_param_must_be_a_number'
+ [[ 2 -eq 1 ]]
+ vagrant ssh app -c 'cd /opt/app/core/opentreemap && envdir /etc/otm.d/env ./manage.py test api.tests.Locations.test_locations_plots_endpoint_max_plots_param_must_be_a_number'
Creating test database for alias 'default'...
Installed 16 object(s) from 1 fixture(s)

Installed 5 object(s) from 1 fixture(s)
Installed 18 object(s) from 1 fixture(s)
System check identified no issues (0 silenced).
E
======================================================================
ERROR: test_locations_plots_endpoint_max_plots_param_must_be_a_number (api.tests.Locations)
----------------------------------------------------------------------
  File "/opt/app/core/opentreemap/api/tests.py", line 376, in test_locations_plots_endpoint_max_plots_param_must_be_a_number
    API_PFX, self.instance.url_name)))
  File "/usr/lib/python3.7/unittest/case.py", line 756, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/lib/python3.7/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
TypeError: 'HttpResponseBadRequest' object is not callable

----------------------------------------------------------------------
Ran 1 test in 0.309s
```

I fixed it with this diff
```diff
diff --git a/opentreemap/api/tests.py b/opentreemap/api/tests.py
index ccffd879..f8bb4c94 100644
--- a/opentreemap/api/tests.py
+++ b/opentreemap/api/tests.py
@@ -370,13 +370,14 @@ class Locations(OTMTestCase):
         self.assertEqual(response.status_code, 200)
 
     def test_locations_plots_endpoint_max_plots_param_must_be_a_number(self):
-        self.assertRaises(HttpBadRequestException, get_signed(
+        response = get_signed(
             self.client,
             "%s/instance/%s/locations/0,0/plots?max_plots=foo" % (
-                API_PFX, self.instance.url_name)))
+                API_PFX, self.instance.url_name))
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.content,
-                         b'The distance parameter must be a number')
+                         b'The max_plots parameter must be '
+                         b'a number between 1 and 500')
 
     def test_locations_plots_max_plots_param_cannot_be_greater_than_500(self):
         response = get_signed(

```

opentreemap/api/tests.py
```diff
@@ -2003,26 +2003,26 @@ def test_401(self):
...
-        auth = base64.b64encode("jim:password")
+        auth = base64.b64encode(b"jim:password")
         withauth = {"HTTP_AUTHORIZATION": "Basic %s" % auth}
-
         ret = get_signed(self.client, "%s/user" % API_PFX, **withauth)
         self.assertEqual(ret.status_code, 200)
 
     def test_malformed_auth(self):
-        withauth = {"HTTP_AUTHORIZATION": "FUUBAR"}
+        auth = base64.b64encode(b"FUUBAR")
```
fungjj92: I am just finding out in later test fixing this should be `auth = base64.b64encode(b"FUUBAR").decode()` and likewise of the changes in this test class. While `base64.b64encode()` now requires bytes, the request headers want strings

fungjj92: eh standby, false positive.

opentreemap/api/tests.py
```diff
@@ -2003,26 +2003,26 @@ def test_401(self):
...
+        withauth = {"HTTP_AUTHORIZATION": "%s" % auth }
```
fungjj92: Oops i accidentally clobbered this test. Possibly a copy paste error.

opentreemap/api/tests.py
fungjj92: OK i still think this is true even though these tests still fail.

jwalgran (commented):
The majority of these changes are clear and straightforward. I made a few comments and suggestions.

opentreemap/api/auth.py
fungjj92: How'd that get there! 😅

fungjj92:
Standby. I am in the middle of fixing auth test issues in this branch, I realized there were some errors I was making in encoding/decoding.

fungjj92:
Alright this is ready to be looked at.
```
Ran 1115 tests in 297.171s

FAILED (failures=25, errors=125, skipped=23)
```

I had been clobbering the basic auth tests with unwanted layers of encoding & decoding. I'm not sure how these failures didn't arise earlier, but alas. I was tipped off by the lack of output at the regex parsers; no surprise but regex is _great_ at preserving and communicating its desired data format, which was helpful in figuring out how the auth data should look.

opentreemap/api/tests.py
```diff
@@ -2003,26 +2004,25 @@ def test_401(self):
...
         self.assertEqual(ret.status_code, 200)
 
     def test_malformed_auth(self):
-        withauth = {"HTTP_AUTHORIZATION": "FUUBAR"}
+        withauth = {"HTTP_AUTHORIZATION": "FUUBAR" }
 
         ret = get_signed(self.client, "%s/user" % API_PFX, **withauth)
         self.assertEqual(ret.status_code, 401)
 
-        auth = base64.b64encode("foobar")
```
jwalgran, line 2018: Consider updating the commit message to explain why we are removing this base 64 encoding. It is my understanding that basic auth requires base 64.

>In basic HTTP authentication, a request contains a header field in the form of `Authorization: Basic <credentials>`, where credentials is the base64 encoding of id and password joined by a single colon `:`. 

_from https://en.wikipedia.org/wiki/Basic_access_authentication_

opentreemap/api/auth.py
```diff
@@ -68,18 +78,22 @@ def firstmatch(regx, strg):
 
 
 def decodebasicauth(strg):
-    if strg is None:
+    if strg is None or len(strg) == 0:
         return None
     else:
-        m = re.match(r'([^:]*)\:(.*)', base64.decodestring(strg))
```
jwalgran, line 74: Consider updating the commit message to explain why we are removing this base 64 encoding. It is my understanding that basic auth requires base 64.

>In basic HTTP authentication, a request contains a header field in the form of `Authorization: Basic <credentials>`, where credentials is the base64 encoding of id and password joined by a single colon `:`. 

_from https://en.wikipedia.org/wiki/Basic_access_authentication_

opentreemap/exporter/tests.py
```diff
@@ -213,16 +217,20 @@ class UserExportsTest(UserExportsTestCase):
...
 
+        # strip the BOM and entry line out
+        reader = csv.reader(str_rows[2:])
+
+        # grab and strip the header
         header = next(reader)
 
-        data = [dict(list(zip(header, [x.decode('utf8') for x in row])))
-                for row in reader]
+        reader_rows = list(reader)
```
jwalgran: Consider adding a comment explaining why it is necessary to convert the reader to a list. It looks like the only use of `reader_rows` in in the list comprehension below, which I would expect to be able to iterate over `reader` as-is.

jwalgran (commented):
I got 264 errors instead of 125

```
Ran 1115 tests in 322.333s

FAILED (failures=25, errors=264, skipped=23)
Destroying test database for alias 'default'...
```

I verified that I had the correct branches and SHAs checked out.

```
~/Projects/otm-cloud (python3)
> kj git cb
===== otm-cloud =====
python3
===== otm-core =====
jf/fix-tests-2to3
===== otm-addons =====
jf/fix-tests-2to3
===== otm-tiler =====
develop
===== otm-ecoservice =====
develop

~/Projects/otm-cloud (python3)
> kj git status
===== otm-cloud =====
On branch python3
nothing to commit, working tree clean
===== otm-core =====
On branch jf/fix-tests-2to3
Your branch is up to date with 'origin/jf/fix-tests-2to3'.

nothing to commit, working tree clean
===== otm-addons =====
On branch jf/fix-tests-2to3
Your branch is up to date with 'origin/jf/fix-tests-2to3'.

nothing to commit, working tree clean
===== otm-tiler =====
On branch develop
Your branch is up to date with 'origin/develop'.

nothing to commit, working tree clean
===== otm-ecoservice =====
On branch develop
Your branch is up to date with 'origin/develop'.

nothing to commit, working tree clean

~/Projects/otm-cloud (python3)
> kj git rev-parse HEAD
===== otm-cloud =====
25eda864651968a4f67298b5c1acada7d437d3a8
===== otm-core =====
82c79a2991aa2bd407706880ae734d859ff773c9
===== otm-addons =====
4b94756207566b68736e9e6fe3fe089cb743d11a
===== otm-tiler =====
65328ada3d6303b59b60d31d16dbc2909e3c2816
===== otm-ecoservice =====
8aeb53b2d858a3227e443c7f912d13bca77bacb0
```

I will reprovision my VM from scratch to make sure something strange isn't going on.

I also commented on a concern I have about removing base64 encoding from the auth process.

jwalgran:
I did provision a fresh `app` VM and got the same results

```
FAILED (failures=25, errors=264, skipped=23)
```

opentreemap/api/tests.py
fungjj92: Thanks so much for this detail. I took another dive in with this better baseline understanding, and get what is going on 100% now. In py3, there are ripple effect from having to have the initial credentials start as bytes. This complication appears on these lines: https://github.com/OpenTreeMap/otm-core/pull/3288/files#diff-c7aac248bb6ed915db70251295a116a3R96-R103. I pushed up changes to this effect.

opentreemap/api/auth.py
```diff
@@ -68,18 +78,30 @@ def firstmatch(regx, strg):
 
 
 def decodebasicauth(strg):
```
jwalgran: Now that have moved the actual decoding into `parse_basicauth` consider renaming this to `split_basicauth` or similar.

opentreemap/api/tests.py
```diff
@@ -2003,26 +2004,25 @@ def test_401(self):
...
-        auth = base64.b64encode("jim:password")
+        auth = base64.b64encode(b"jim:password")
         withauth = {"HTTP_AUTHORIZATION": "Basic %s" % auth}
-
         ret = get_signed(self.client, "%s/user" % API_PFX, **withauth)
         self.assertEqual(ret.status_code, 200)
 
     def test_malformed_auth(self):
-        withauth = {"HTTP_AUTHORIZATION": "FUUBAR"}
+        withauth = {"HTTP_AUTHORIZATION": "FUUBAR" }
```
jwalgran: Consider removing this trailing space

opentreemap/api/tests.py
```diff
@@ -2083,7 +2083,7 @@ def test_json_requires_admin(self):
 class PasswordResetTest(OTMTestCase):
     def setUp(self):
         self.instance = setupTreemapEnv()
-        self.jim = User.objects.get(username="jim")
+        self.jim = User.objects.get(username="jim", password="password")
```
jwalgran: Consider explaining in this PR conversation why we now need to also include `password` as a filter parameter.

jwalgran (approved):
Thank you for your diligent work on this PR. I compared the summary output of `./scripts/manage.sh test`.

On the `python3` branch
> FAILED (failures=19, errors=392, skipped=23)

On this branch
> FAILED (failures=25, errors=264, skipped=23)

👍 

I also made a few nonessential suggestions.

opentreemap/api/tests.py
fungjj92: Actually, I think this is a mistake. Good catch!

fungjj92:
Thank you for the well-considered rounds of reviews. Onward! Merging.

PR #3292, opened 2020-02-04
Fix importer app tests

fungjj92:
## Overview

This PR fixes failing unit tests in the `importer` django app. Same old story as other PRs, all unit test failures continue to be because of the clarification between Binary vs Unicode encoding.

The more interesting case seen here is that CSV generation is more strict. Before, a temporary file wrapper of a byte stream was considered a valid csv file. Now, it is not. A valid csv file must be an iterator object of unicode strings. So there's a bit of conversion we do to repackage the byte stream into a proper csv. I left an in-line comment that I hope is clear. **Let me know if it could be made clearer**: in-line comments tend to be particularly helpful in infrequently touched codebases like this.

### Testing
`./scripts/manage.sh test importer` should run cleanly and only `treemap` tests should be broken 🌮 

```
Ran 1121 tests in 309.863s

FAILED (failures=18, errors=11, skipped=23)
```

### DIFF
--- a/opentreemap/importer/util.py
+++ b/opentreemap/importer/util.py
@@ -3,7 +3,7 @@
 
 import codecs
 import csv
-
+from io import StringIO
 
 def _clean_string(s):
     s = s.strip()
@@ -45,5 +45,11 @@ def utf8_file_to_csv_dictreader(f):
     # CSV standard "" to escape a double quote. Excel and LibreOffice
     # use this escape by default when saving as CSV.
     dialect.doublequote = True
-    return csv.DictReader(_as_utf8(f),
+
+    # at this point, the binary filestream must be repackaged to the csv spec,
+    # that is an iterator of unicode strings. time limited a more elegant solution
+    # than allocating memory for up to two copies of the data but we anticipate having
+    # sufficient RAM
+    csv_body = StringIO(f.read().decode())
+    return csv.DictReader(csv_body,
                           dialect=dialect)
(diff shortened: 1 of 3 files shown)


opentreemap/importer/util.py
```diff
@@ -45,5 +45,9 @@ def utf8_file_to_csv_dictreader(f):
     # CSV standard "" to escape a double quote. Excel and LibreOffice
     # use this escape by default when saving as CSV.
     dialect.doublequote = True
-    return csv.DictReader(_as_utf8(f),
+
+    # the binary filestream (f) must be repackaged to the csv spec,
+    # that is an iterator of unicode strings
+    csv_body = StringIO(f.read().decode())
```
jwalgran, line 53: I was concerned that we were now reading the entire CSV into memory when we were previously wrapping it with a reader that could iterate over it line by line. However, when I traced the code path backward from `utf8_file_to_csv_dictreader` I found that `process_csv` is already reading the entire file.

https://github.com/OpenTreeMap/otm-core/blob/053e5c17322719a80ee93c190c05df285890bc58/opentreemap/importer/views.py#L712

This may mean that we are going to have 2 copies of the entire file in memory while it is being sliced into individual database rows. That is probably ok.

jwalgran (approved):
Nice job. Tests are passing.

opentreemap/importer/util.py
fungjj92: I considered the memory implications and it slipped my mind to vocalize that in the PR write-up, thanks for bringing it up.

Ideally, the CSV data would be read in as a stream in one encoding (like unicode `str` with StringIO) and have that be OK all the way through but that is not the case. In earlier CSV processing steps, a bytestream is necessary but then a string stream is needed here. Surely there is a more robust way to handle this than holding up to two copies of the data in memory, but I think that would involve much more time spent in the code base. I have no idea how big user-uploaded CSVs are IRL, but I am glad for the sanity check from your comment that this may not be too big a burden for this code pattern.

I will update the commit message with detail of this consideration of memory.

Issue #3298, opened 2020-07-23
Add API documentation


jwalgran:
Add markdown documentation explaining:

- How to obtain an API key (contact OTM)
- The available API endpoints
- How to do the required HMAC signing, with Python examples.

PR #3299, opened 2020-07-27
Add API documentation and a script to make HMAC signed requests via cURL

jwalgran:
We add API documentation covering available, active endpoints and poviding JSON request and response examples for third-party developers who would like to integrate OTM data into other systems.  

We also add a command line tool (`scripts/hmaccurl.py`) to make HMAC signed requests via cURL with support for a small subset of cURL options, suitable for testing API endpoints from the command line.

Connects #3298 

### Testing

To conserve development time this does not need a full interactive test. I used `hmaccurl.py` to generate the JSON responses throughout the document.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1,0 +1,1339 @@
+# OTM API Documentation
+
+**Table of Contents**
+
+- [Concepts](#concepts)
+- [Endpoints](#endpoints)
+    - [Status](#status)
+    - [Version](#version)
+    - [User](#user)
+    - [Instance](#instance)
+    - [Plot and Tree](#plot-and-tree)
+- [HMAC](#hmac)
+
+## Concepts
+
+OpenTreeMap hosts data for many distinct groups. The API refers to these as instances. Instances are completely separate from each other. The same user can access data from different instances, but no data is passed between instances.
+
+Plot and tree are separate resources. The plot represents the physical location of a street tree and has a latitude and longitude. A tree has a species and diameter and must be associated with a plot. A plot may have no tree, or a single tree. A plot may be associated with multiple tree records over time (when trees die and are replanted, for instance), but only one at a time.
+
+The OTM API was developed to provide the data required by the native OTM mobile applications. The current version is not designed as a general purpose API and does not have all the conveniences of a general purpose API (for example, errors responses are not serialized as JSON).
+
+All requests to the OTM API must be signed with HMAC. See the HMAC section of this document for details.
+
+## Endpoints
+
+### Status
+
+`GET /`
+
+Get the status of the API.
+
+Returns an array JSON objects.
+
+```js
+[
+    {
+        "api_version": "v4",
+        "message": "",
+        "status": "online"
+    }
+]
+```
+
+### Version
+
+`GET /version`
+
+Get the current version of the API.
+
+Returns a JSON object.
+
+```js
+{
+    "api_version": "v4",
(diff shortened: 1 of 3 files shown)


jwalgran:
The build failure was in Update APT Cache

```
TASK [Update APT Cache] ********************************************************
Monday 27 July 2020  18:10:56 -0400 (0:00:03.628)       0:00:03.772 *********** 
[WARNING]: Updating cache and auto-installing missing dependency: python3-apt
fatal: [services]: FAILED! => {"changed": false, "msg": "Could not import python modules: apt, apt_pkg. Please install python3-apt package."}
```

Perhaps related to the end of Python 2.7 support. We need to fix the build in general, but this documentation and non-runtime tool update PR is not exercised by CI.

doc/api.md
```diff
@@ -0,0 +1,1326 @@
...
+        },
+        "geoRevHash": "c81e728d9d4c2f636f067f89cc14862c"
+    }
+]
+```
+---
+
+`GET /locations/{latitude},{longitude}/instances`
+
+Get the instances which the current user has contributed to and instances within a specified distance of the specified point.
```
mmcfarland, line 199: This mentions a "specified distance" - but that doesn't appear to be a path or query parameter. Can this distance actually be specified or is it hard coded?

doc/api.md
```diff
@@ -0,0 +1,1326 @@
...
+
+Returns a JSON object representing the photo.
+
+### HMAC
+
+All OTM API requests must be signed using
+[HMAC](https://enwikipediaorg/wiki/HMAC). The server-side HMAC implementation is based on the [`hmac` package included with Python](https://docs.python.org/3/library/hmac.html)
+
+The [`hmaccurl.py`](../scripts/hmaccurl.py) command line tool is an example of
+how API requests can are signed with HMAC. It is a Python script that wraps the
```
mmcfarland:
```suggestion
how API requests can be signed with HMAC. It is a Python script that wraps the
```

doc/api.md
```diff
@@ -0,0 +1,1326 @@
...
+`curl` command and adds the required HMAC signature and timestamp query string
+arguments. This command line tool can be used as-is to make requests to the API
+or can be embedded in a larger Python application.
+
+As additional reference, these are sections of the OTM code related to HMAC signing.
```
mmcfarland, line 1195: It's not immediately clear to me which ACCESS_KEY credentials I'm meant to use with the HMAC tool included. Is this guide aimed at users of OTM Cloud to interact with instances and would therefore need credentials provided from the app? Perhaps a small section could be included in how to access required credentials for the signing process.

mmcfarland (commented):
This looks like a good level of detail; though I had just a few comments suggestions for clarity.

doc/api.md
jwalgran: Good catch. There is a `distance` query string param. Added in fixup c79d39c7.

doc/api.md
jwalgran: Good question. There is not web UI for generating access tokens. It must be done through the Django shell. I added a couple paragraphs in fixup 335465dd that explain more about the generation and usage of the key pair.

jwalgran:
Thanks for the thoughtful review. I address the comments with fixups.

mmcfarland (approved):
Fixups look good, this will be a helpful doc :+1:

